PR #24, opened 2020-12-05
Fix error when branch is only local

hkupty:
If the branch was not yet pushed, tokens[2] is nil and vim.split breaks


TimUntersberger:
Thank you!

PR #25, opened 2021-01-10
Allow configuration of buffer kind

RianFuro:
tl;dr: I want splits :)

With this change, a user can set the way in which the status buffer will
open by using the lua invocation directly like so:

`require'neogit.status'.create('split')`

Defaults to 'tab' to preserve old behavior. No argument sanitization, but it's also undocumented, so eh. Not sure if we should add a specific command like `SNeogit` or just document the lua api.

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -637,7 +637,8 @@ local function unstage()
   refresh_status()
 end
 
-local function create()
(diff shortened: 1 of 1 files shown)


TimUntersberger:
Whoops. I forgot to add this. Thank you!

PR #26, opened 2021-01-10
Display improvements to status buffer overview

RianFuro:
Adds the commit message of the current commit of head and upstream to the status buffer and only displays upstream when there's actually an upstream configured.

Had to switch highlighting of the head and upstream lines to the highlighting api, because I don't know how to do the current highlighting with :syntax. The highlights may need to be cleaned up on refresh, but I wasn't able to actually refresh an open status buffer with a changed branch and I didn't want to implement blind.

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -152,11 +152,10 @@ local function display_status()
   end
 
   locations = {}
-
   local line_idx = 3
   local output = {
-    "Head: " .. status.branch,
-    "Push: " .. status.remote,
+    "Head: " .. status.branch .. " " .. status.HEAD.message,
+    "Push: " .. status.remote .. " " .. (status.upstream.message or ''),
     ""
   }
 
@@ -310,11 +309,17 @@ local function refresh_status()
     return
   end
 
-  for _,x in pairs(status) do
-    if type(x) == "table" then
-      for _,i in pairs(x) do
-        i.diff_open = false
-      end
+  for _,x in ipairs({
+    'untracked_files',
+    'unstaged_changes',
+    'unmerged_changes',
+    'staged_changes',
+    'unpulled',
+    'unmerged',
+    'stashes'
+  }) do
+    for _,i in ipairs(status[x]) do
+      i.diff_open = false
     end
   end
 
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -152,12 +152,15 @@ local function display_status()
   end
 
   locations = {}
-  local line_idx = 3
+  local line_idx = 2
   local output = {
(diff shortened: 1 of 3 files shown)


lua/neogit/status.lua
```diff
@@ -254,6 +256,10 @@ local function display_status()
   })
 
   status_buffer:set_lines(0, -1, false, output)
+  vim.api.nvim_buf_add_highlight(0, -1, 'NeogitBranch', 0, 5, 5 + status.head.branch:len() + 1)
+  if status.upstream ~= nil then
+    vim.api.nvim_buf_add_highlight(0, -1, 'NeogitRemote', 1, 5, 5 + status.upstream.branch:len() + 1)
+  end
```
TimUntersberger, line 262: Why is this needed? Shouldn't the existing syntax file cover these cases?

syntax/NeogitStatus.vim
```diff
@@ -20,8 +20,6 @@ for section in b:sections
   execute 'hi def link Neogit' . id . ' Function'
 endfor
 
-syn region NeogitHeadRegion start=/^Head: \zs.*/ end=/$/ contains=NeogitBranch
-syn region NeogitPushRegion start=/^Push: \zs.*/ end=/$/ contains=NeogitRemote
```
TimUntersberger, line 24: Related to question above.

lua/neogit/status.lua
RianFuro: When I added the commit message header to the line, the highlighting extended to the full line, I only wanted to have the branch name highlighted as before.
I skimmed through vim documentation, but I couldn't figure out how to do that with Syntax regions, so I used the highlighting api instead.
There's probably a way by defining a Syntax region with a start like `^Head: \w` that resets the highlighting for the commit message, but then using the api sounds less involved to me

syntax/NeogitStatus.vim
```diff
@@ -3,8 +3,9 @@ if exists("b:current_syntax")
 endif
 
 syn match NeogitObjectId /^[a-z0-9]\{7} /
-syn match NeogitBranch /.*/ contained
-syn match NeogitRemote /[a-zA-Z/]/ contained
+syn match NeogitCommitMessage /.*/ contained
+syn match NeogitBranch /\S\+/ contained nextgroup=NeogitCommitMessage
+syn match NeogitRemote /\S\+/ contained nextgroup=NeogitCommitMessage
```
RianFuro, line 8: @TimUntersberger Ok, decided to have another go and made it work with just syn groups. Thanks for pulling me back to that :D

TimUntersberger (approved):

TimUntersberger:
LGTM, thank you!

PR #27, opened 2021-01-17
WIP: Contextual section highlighting

RianFuro:
There's a lot rough edges on this one, but I wanted your opinion on if you even want to take the UI in that direction, so here's a pull request :smile: 

The changes come in 2 parts:
1. I refactored how file diffs are rendered to the buffer, so I can better keep hunk location data in the locations table. I suppose this worsens performance on the refresh somewhat, but it's easier to keep the locations table up-to-date like that. If it *is* a problem, I'm happy to optimize that.
2. I used the aforementioned location data to get the current smallest section under the cursor, then used `signs` to highlight that region. This gives the same context highlighting that magit implements. As a next step, using the same data to contextually move the cursor to the next section/hunk should be trivial.

The rough edges:
- I realize there already is logic to get the current hunk under the cursor, however I couldn't quite get that to work for my use case without digging too deep and possibly breaking things. Again, happy to take that instead of my locations table extension if needed.
- The signs all need a numeric unique ID and currently that's just an incrementing counter. That just screams of a potential overflow, so I suppose that needs to be optimized. Also I don't know how much duplicate IDs destroy the placed signs in our use case; might be fine to ignore the unique constraint (vim doesn't complain, but claims undocumented behavior in the docs)
- I included my own highlights for hunk headers and selection highlighting. I could possibly get away with an existing highlighting group for the standard highlighting, but for hunk headers it already gets tricky, since it needs a *highlighted* variant. I'm not sure what's the correct approach on this one.
`magit` just seems to role it's own faces with default values in emacs for that and the color scheme libraries take care of them. I'm not sure how widely spread this is in the vim ecosystem, but a quick skim through my `solarized.vim` tells me that something similar is happening there.
- Diff markers may need something similar to the HunkHeader and -Highlighted highlight groups.
- Didn't test everything yet, so selective staging might be broken somewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,3 +41,15 @@ L - open log popup
 P - open push popup
 
 p - open pull popup
+
+## Contextual Highlighting
+
+The colors for contextual highlighting are defined with these highlight groups:
+```viml
+hi def NeogitDiffAddHighlight guibg=#404040
+hi def NeogitDiffDeleteHighlight guibg=#404040
+hi def NeogitDiffContextHighlight ctermbg=4 guibg=#333333
+hi def NeogitHunkHeader guifg=#cccccc guibg=#404040
+hi def NeogitHunkHeaderHighlight guifg=#cccccc guibg=#4d4d4d
+```
+You can override them to fit your colorscheme by creating a `syntax/NeogitStatus.vim` in your vim configuration.
--- a/ftplugin/NeogitStatus.vim
+++ b/ftplugin/NeogitStatus.vim
@@ -15,3 +15,7 @@ setlocal foldminlines=0
 setlocal foldtext=NeogitFoldFunction()
 
 au BufWipeout <buffer> lua __NeogitStatusOnClose()
+
+augroup NeogitStatusHighlightUpdater
+autocmd CursorMoved NeogitStatus :lua require'neogit.status'.update_highlight()
+augroup END
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -93,6 +93,20 @@ function Buffer:set_foldlevel(level)
   vim.cmd("setlocal foldlevel=" .. level)
 end
 
+function Buffer:open_fold(line, reset_pos)
+  local pos
+  if reset_pos == true then
+    pos = vim.fn.getpos()
+  end
+
+  vim.fn.setpos('.', {self.handle, line, 0, 0})
+  vim.cmd('normal zo')
+
+  if reset_pos == true then
+    vim.fn.setpos('.', pos)
+  end
+end
+
 function Buffer:set_filetype(ft)
   vim.cmd("setlocal filetype=" .. ft)
 end
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -1,7 +1,6 @@
 package.loaded['neogit.buffer'] = nil
 
 local mappings_manager = require("neogit.lib.mappings_manager")
-local next_sign_id = 1
 
 local Buffer = {
   handle = nil,
@@ -108,10 +107,14 @@ function Buffer:open_fold(line, reset_pos)
   end
 end
 
-function Buffer:place_sign(line, name, group)
-  local sign_id = next_sign_id
-  next_sign_id = next_sign_id + 1
+function Buffer:place_sign(line, name, group, id)
+  -- Sign IDs should be unique within a group, however there's no downside as
+  -- long as we don't want to uniquely identify the placed sign later. Thus,
+  -- we leave the choice to the caller
+  local sign_id = id or 1
 
+  -- There's an equivalent function sign_place() which can automatically use
+  -- a free ID, but is considerable slower, so we use the command for now
   local cmd = 'sign place '..sign_id..' line='..line..' name='..name
   if group ~= nil then
     cmd = cmd..' group='..group
@@ -119,6 +122,7 @@ function Buffer:place_sign(line, name, group)
   cmd = cmd..' buffer='..self.handle
 
   vim.cmd(cmd)
+  return sign_id
 end
 
 function Buffer:clear_sign_group(group)
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -1,6 +1,7 @@
 package.loaded['neogit.buffer'] = nil
 
 local mappings_manager = require("neogit.lib.mappings_manager")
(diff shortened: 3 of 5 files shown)


RianFuro:
As an alternative to rolling custom highlight groups, we could probably get away with just using `CursorLine` for section highlighting, `StatusLine` for hunk headers and not doing any extra highlighting for diff markers and hunk headers when they are in the highlighted section.

TimUntersberger:
Could you add some screenshots so I can quickly appreciate the changes you made?

RianFuro:
@TimUntersberger These are with  current custom highlights (just grey shades, everything else is my solarized theme; very much subject to change, if we even go that way):
image
image
image
image
image
image

These would be using just existing highlight groups:
image
image
image
They obviously look better in my colorscheme since they use the same colors, but have the drawback of having the same background as the diff markers when highlighted, and folded hunks are not highlighted, which looks confusing (cursor is on the folded hunk here):
image

TimUntersberger:
> I refactored how file diffs are rendered to the buffer, so I can better keep hunk location data in the locations table. I suppose this worsens performance on the refresh somewhat, but it's easier to keep the locations table up-to-date like that. If it is a problem, I'm happy to optimize that

I think we shouldn't eager load this. As I understood it you are now preloading all diffs and then adding highlighting, which has caused some serious slowness when I initially created this plugin.

> used signs to highlight that region

I have never used signs before, can you give me a quick tldr why you chose to use them and which alternatives you have tried/thought about?

RianFuro:
> I think we shouldn't eager load this. As I understood it you are now preloading all diffs and then adding highlighting, which has caused some serious slowness when I initially created this plugin.

Well, loading is still deferred to be async, but they are rendered as soon as they are loaded, not when they are expanded. Basically what happens when a user pressed `4` now always happens. I can totally see how that would be a slow render on a huge changeset though. I'll make some time to profile that and revert changes as needed.
Also note that the current render-when-diff-is-opened mechanic doesn't respect usual fold controls like `zO` and `zR` and potential user-remaps that control the foldlevel. So, I would prefer my approach since it supports those natively, granted I can get it to perform of course.

> I have never used signs before, can you give me a quick tldr why you chose to use them and which alternatives you have tried/thought about?

Signs are basically vim's way of "quickly highlighting a line" and is supposed to be used by IDE-tools, debugger, etc. to place signs and highlight lines of interest - which is basically what we are when highlighting the status buffer, so that's a fit.
My first attempt was just using neovim's highlighting api to apply line-wise highlighting, which in principle works just as well. The only difference is that signs allow you to set a `linehl` which sets a highlighting group for a whole line, extending to the end. Usually, and that includes neovim's highlight api, you can only apply highlighting to actual characters in the buffer. Using signs I can extend the highlighting to the whole line, which for highlighting the section just looks way better.

TimUntersberger:
> Well, loading is still deferred to be async, but they are rendered as soon as they are loaded, not when they are expanded. Basically what happens when a user pressed 4 now always happens

Ah then I misunderstood what happens. This should be fine I think.

> Signs are basically vim's way of ...

Ah ok thank you! 

I really hope we manage to merge this soon as this will make the UI look way more cooler thanks to you! :)

TimUntersberger:
Could you provide a small gif that shows this PR in action? From startup time to moving through sections, hunks, etc.

RianFuro:
@TimUntersberger I tried but it looks laggy compared to what's going on on my screen, so keep that in mind.
Peek 2021-01-18 15-15

You can however just checkout my branch and try it yourself, it's directly on top of your master branch anyway.

TimUntersberger:
This is really cool! I'd gladly merge this once this PR is ready.

RianFuro:
@TimUntersberger Did some *very* rudimentary performance testing:
Setup:
- Empty file in version control
- Added lots of lines to the file
- Opened nvim and neogit

With below 2k lines, there's no noticeable lag while moving the cursor. Above that it's noticeable, but not problematic up to about 5k lines. With 10k lines the UI becomes basically unusable, because vim just takes too long updating signs. Some of that time should be from regex matching every line 3 times (hunk header, diff add & diff delete), so there's definitely room for improvement.

Even with 10k lines there were no problems rendering the loaded diff, so I suppose we have no problem on that front.

TimUntersberger:
> Even with 10k lines there were no problems rendering the loaded diff, so I suppose we have no problem on that front.

That's good to hear 👍 

> With below 2k lines, there's no noticeable lag while moving the cursor. Above that it's noticeable, but not problematic up to about 5k lines. With 10k lines the UI becomes basically unusable, because vim just takes too long updating signs. Some of that time should be from regex matching every line 3 times (hunk header, diff add & diff delete), so there's definitely room for improvement.

I think this is not possible to fix for us. I can notice serious lag when having complex syntax highlighting in 10k lines files in vim. Maybe we could have "virtual highlighting"? Deleting highlighting that is out of some range? Would this be possible or is too much work for such a rare situation?

RianFuro:
> virtual highlighting

That's definitely an avenue we can take. I would optimize my regex use before that though. In any case, I think that's an issue for when somebody actually wants to manage a 10k lines diff through neogit.

TimUntersberger:
> In any case, I think that's an issue for when somebody actually wants to manage a 10k lines diff through neogit.

👍 I think you shouldn't have a diff of 10k lines anyways 😄. If someone hits this problem we can fix it, but for now this should be already enough.

RianFuro:
Alright, I think I'm about done. One last question: 

I left the grayscale highlight groups in there since they are more customizable. For example, to play nice with my solarized theme, these overrides do nicely:
```viml
hi! NeogitHunkHeader guibg=#839496 guifg=#073642
hi! NeogitHunkHeaderHighlight guibg=#93a1a1 guifg=#073642
hi! NeogitDiffContextHighlight guibg=#073642
hi! NeogitDiffAddHighlight guibg=#073642
hi! NeogitDiffDeleteHighlight guibg=#073642
```
Since no colorscheme supports these natively (yet? :D), I suppose we should document that somewhere? Also, I only defined guicolors since I haven't used a 16 color terminal in ages and I'm not sure who still would. If you're fine with that, I'm ready to go forward :)

TimUntersberger:
I played with the context highlighting a bit and I think that the grayscale coloring looks really beautfiul. 

Could you add some documentation about the highlighting groups in the README? 

After this change I'd gladly merge this PR if you think it is ready!

RianFuro:
Already added the README update with the latest commit :)

TimUntersberger:
> 
> 
> Already added the README update with the latest commit :)

Whoops! sry!

PR #33, opened 2021-01-22
test: PoC Unit Test setup

### LINKED ISSUE #32: Including luaunit?
Should we include [luaunit](https://github.com/bluebird75/luaunit) to be able to run some unit tests? Judging from our discussions from #30 we could really benefit from some :)

RianFuro:
Closes #32 

Setup unit tests using `plenary.nvim`'s test harness. Plenary is
expected to be installed as a plugin on the neovim instance you're
running your tests on.

I opted out of adding plenary to the repository since git plugins are usually directly pulled from repositories and including plenary into that seems wrong. So I just expected it to be installed already.

I included a small sample test suite to show what's possible, but no actually meaningful tests yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,2 @@
+test:
+	nvim --headless --noplugin -c "lua require(\"plenary.test_harness\").test_directory_command('tests/ {minimal_init = \"tests/minimal-init.nvim\"}')"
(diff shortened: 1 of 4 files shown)


TimUntersberger:
Could you resolve the conflicts so we can merge this?

RianFuro:
Done!

PR #35, opened 2021-02-02
Add documentation to the project with vim's :help

KarlWithK:
Every plugin should have help tags. I don't think it's a good idea, in the long run, to have everything in the README. I started the documentation which is heavily inspired by  Tim Pope's fugitive, hopefully, that's ok.


TimUntersberger:
Shouldn't it be possible to generate this from a markdown file?

PR #40, opened 2021-02-07
Tests and Refactors

### LINKED ISSUE #39: Staging a selection, then staging the rest of the file creates 2 entries in the Staged Changes list
Just a visual issue, the repository itself is of course correct.

image

RianFuro:
I used our new testing setup to put patch generation and staging functionality under test. Then used my newfound confidence and refactored staging to use the patch generation where possible.

Things to note:
- I had to remove a few optimizations due to problems I found while working on this.
  - I had to always compute the git base directory as changing directory inside neovim didn't update the cache - and between `:cd`, `:lcd` and `:tcd` the cache becomes difficult to maintain.
  - I had to completely re-fetch git information after staging as the UI wasn't updating correctly when staging selections. Incidentally, this should also fix #39.
- For the end-to-end tests for the staging functionality, I had to setup and prepare the repository for every test, since plenary runs specs in parallel. Also the .git directory needs to be reset for every test, for obvious reasons. I used bash commands to move files around, so the tests can't be run from windows cmd and powershell currently.


TimUntersberger:
> Then used my newfound confidence

😆 

> I had to completely re-fetch git information after staging as the UI wasn't updating correctly when staging selections. Incidentally, this should also fix

This kind of concerns me. Staging should feel really slow now right?

RianFuro:
Nah not at all. Status is fast and fetching diffs happens off the main thread anyway. Redrawing the status buffer could be a problem potentially, but judging from my performance tests with highlighting I don't think this would be a problem for your average repository.

That being said, I completely realize this is a step back and opens at least a small can of worms. I also only tested it with the neogit repository and my own little toy repo i created for the test suite, so bigger repositories *might* feel a little slow down.
I also feel though that we shouldn't compute changes ourselves and always depend on git for status changes. I'd rather improve performance by selectively updating status information instead of trying to update our own application state to match git's state; simply because it's more robust.

RianFuro:
Did some quick tests on my machine with some repositories from work, cannot notice any slowdown (with a changelist of ~30 files, apparently I did not clean up :D)

PR #42, opened 2021-02-14
Using the libuv event loop for spawning processes

RianFuro:
Alright, this will be a bigger one.

I wanted to play around with neovim's integration with libuv and since you have #12 open I thought doing it in here might get some actual use as well. Feedback is very welcome.

First of all, since I remember callback hell from nodejs all too well, I needed some way to flatten the callback hierarchy. Thankfully there's an [amazing tutorial](https://github.com/ms-jpq/neovim-async-tutorial) on using coroutines to design an async-await-like set of functions for just that, so I took that, adapted the api a little and ironed out some rough edges.

Next, since `textlock` is a problem with using libuv, all functions below the surface api must not call to vimscript (either with `vim.fn.*` or `vim.cmd`). So, all references to `vim.fn.matchlist` have been rewritten to use plain `string.match` logic instead. There's of course the option to wait for the lock with `vim.schedule[_wrap]`, but I wanted to prevent needless stalling as much as possible.

Finally, all functions related to stage/unstage and reload are now running on the new async library. The old jobs module exists so existing code continues to function, but will be removed once I'm finished with the transition.

Performance-wise, I do not notice any difference during normal use so far, but the testsuite feels a tad faster. I'm not sure how `jobstart` and friends synchronize back to the main loop, but I didn't expect much of a difference anyway. Libuv I'm fairly familiar with at least.

Open points:
- [x] New job control doesn't write to the command history yet.
- [x] Porting the remaining jobs over
- [ ] More tests as to not accidentally break things


lua/neogit/lib/git/diff.lua
```diff
@@ -31,17 +29,17 @@ local function parse_diff(output)
   for i=1,len do
     local line = hunks[i]
     if not vim.startswith(line, "+++") then
-      local matches = vim.fn.matchlist(line, "^@@ -\\([0-9]*\\),\\?\\([0-9]*\\)\\? +\\([0-9]*\\),\\?\\([0-9]*\\)\\? @@")
+      local index_from, index_len, disk_from, disk_len = line:match('@@ %-(%d+),?(%d*) %+(%d+),?(%d*) @@')
```
TimUntersberger, line 32: Damn `string.match` looks way cleaner.

TimUntersberger (commented):
LGTM for now! Good work!

lua/neogit/lib/git/diff.lua
RianFuro: You don't have to escape as much, but it's also less powerful than vim's (or any) regex engine. Thankfully our matching needs aren't that complicated to begin with :D

RianFuro:
@TimUntersberger Alright, I do think I'm done. 

No new tests after all, there's a few more questions I need to tackle for that:
- I can't get past the confirmation for discarding changes programmatically, so no idea how to put that under test. I could add an option to disable the confirmation, but that feels cheap.
- I can't test push/pull without substantially enhancing the repository fixture. Need to setup local remote fixtures, possibly separate ones for push and pull, so testing those manually was just much easier.
- Testing log was just too much of a hassle compared to just opening the log buffer and checking that everything is alright.

All changes have been tested manually though, so I'm pretty confident everything is alright.

TimUntersberger (approved):

TimUntersberger:
Do you still want to add something? I saw that you were still commiting to this PR so just making sure I am not merging prematurely.

RianFuro:
I was just fixing stuff I noticed from general usage, so I'm fine as it is right now. If i notice anything else I'll just open a new one^^

RianFuro:
I lied. found more things :sweat_smile: Btw, are you fine with the batch as it is or should i squash it down?

TimUntersberger:
> Btw, are you fine with the batch as it is or should i squash it down?

It is fine like this.

> I lied. found more things 

Can I merge now? 😆

RianFuro:
> Can I merge now?

yes :joy:

PR #45, opened 2021-02-27
Could you add "vsplit" for create argument?

PHSix:
In most time, I this "vsplit" is more comfortable than "split".
Like this:
图片

### DIFF
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -137,8 +137,9 @@ function Buffer.exists(name)
   return vim.fn.bufnr(name) ~= -1
 end
 
-function Buffer.create(config)
+function Buffer.create(config, options)
   local config = config or {}
+  local width = options.width or 50
   local kind = config.kind or "split"
   local buffer = nil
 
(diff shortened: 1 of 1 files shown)


lua/neogit/lib/buffer.lua
```diff
@@ -137,8 +137,9 @@ function Buffer.exists(name)
   return vim.fn.bufnr(name) ~= -1
 end
 
-function Buffer.create(config)
+function Buffer.create(config, options)
   local config = config or {}
+  local width = options.width or 50
```
TimUntersberger, line 142: I don't think the default width should be a static value. Providing no width should behave like executing `vnew` without a width argument.

TimUntersberger (changes requested):
Could you also add "vsplit" to the usage section?

lua/neogit/lib/buffer.lua
PHSix: yes, I think there is no need to add a configure about windows width. But add "vsplit" to usage section is necessary.

PHSix: Could you add a usage section about "vsplit".

PR #46, opened 2021-03-01
Keep commit message on commit failure and more

### LINKED ISSUE #23: Save commit message
When in a repo that uses git hooks, commit might not go through. Saving the commit message and reopening it if the last command didn't go through would be good. If it's stored in a file, not overriding the content of the file if it exists (only replacing the commented lines) would solve the issue.

RianFuro:
Closes #23 

This is from my master branch though, so there's quite a bit more in the MR:
- there's a new way to build and call git commands via a builder pattern. I found the existing way a bit cumbersome and metatables are fun, so here we are :D
- added some of the missing commands for commit and log popups that were easy to add.
- replicated the pull/push commands to both p and u. Since we currently do not have the distinction implemented (and do we even want to?), having one of them do nothing is just confusing.

Regarding #23, I've changed the file we use to store the commit message in. The primary reason is that git manages the `COMMIT_EDITMSG` file itself and that conflicted with the file handling I implemented. We now read the content of the commit message file when opening the dialog and put the existing text into the commit message buffer. That way, if the commit fails, the user gets the last commit message back into the buffer on the next attempt. On a successful commit the file is deleted.

### DIFF
--- a/lua/neogit/popups/commit.lua
+++ b/lua/neogit/popups/commit.lua
@@ -156,7 +156,19 @@ local function create()
         {
           key = "w",
           description = "Reword",
-          callback = function() end
+          callback = function()
+            a.dispatch(function ()
+              local msg = a.wait(cli.log.max_count(1).pretty('%B').call())
+              msg = vim.split(msg, '\n')
+
+              a.wait(prompt_commit_message(msg))
+              local _, code = a.wait(cli.commit.commit_message_file(COMMIT_FILE).amend.only.call())
+              if code == 0 then
+                a.wait(uv.fs_unlink(COMMIT_FILE))
+                __NeogitStatusRefresh(true)
+              end
+            end)
+          end
         },
         {
           key = "a",
--- a/lua/neogit/popups/commit.lua
+++ b/lua/neogit/popups/commit.lua
@@ -94,6 +94,12 @@ local function create()
         enabled = false
       },
       {
+        key = "S",
+        description = "Do not sign this commit",
+        cli = "no-gpg-sign",
+        enabled = false
+      },
+      {
         key = "R",
         description = "Claim authorship and reset author date",
         cli = "reset-author",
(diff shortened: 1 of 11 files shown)


lua/neogit/popups/commit.lua
```diff
@@ -156,7 +156,19 @@ local function create()
...
           key = "w",
           description = "Reword",
-          callback = function() end
+          callback = function()
+            a.dispatch(function ()
+              local msg = a.wait(cli.log.max_count(1).pretty('%B').call())
+              msg = vim.split(msg, '\n')
+
+              a.wait(prompt_commit_message(msg))
+              local _, code = a.wait(cli.commit.commit_message_file(COMMIT_FILE).amend.only.call())
```
TimUntersberger, line 171: Is `only` required here?

RianFuro: yes, otherwise the operation would fail without staged changes.
From the manpage:
```
If this option is specified together with --amend, then no paths need to be specified, which can be used to amend the last commit without committing changes that have already been staged. 
```

TimUntersberger:
LGTM. Tell me when you are happy with the PR.

RianFuro:
@TimUntersberger I'm good to go. Next thing for me will probably be either rebasing or stash management, since I need that quite often, so the next bit will probably take a while anyway.

On that note, would you be open to the idea of giving me collaborator access to the project? I'll still run bigger changes by you of course but I feel like I can manage smaller stuff that comes up occasionally by now - and I'm pretty set to contribute for the foreseeable future :D

TimUntersberger:
> since I need that quite often

That is also how I decide what to implement 😆 

> would you be open to the idea of giving me collaborator access to the project

Sure 👍 

> and I'm pretty set to contribute for the foreseeable future

Thank you for past and future contributions :)

PR #50, opened 2021-03-10
Stash popup

RianFuro:
Stash control. This PR implements stashing everything or just the index, as well as popping, applying or dropping a highlighted stash.

Things to note:
- I chose `Z` for the popup, since `z` is used for folds in vim. Any other key would make just as much sense I guess, I'm just not very creative :sweat_smile: 
- The logic for staging the index only is lifted straight from magit. It *should* support also stashing the worktree, however I had no success with that, so I fell back to just using `git stash` for that. Executing the git commands manually though gives the correct result, so I'm not even sure what's the problem. Any hint would be much appreciated, since this issue is also what prevents me from implementing the `--untracked` flag.
- Trying to apply (or drop/pop) a stash while not highlighting one with the cursor currently just silently does nothing. We could in this case, like magit does, prompt for a stash to apply the operation to, but I didn't want to dabble with user input just yet. I need to look into that for branch control anyway though, so I'll get back to that later.
- No stash message can be set for the same reason.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A **work-in-progress** Magit clone for Neovim that is geared toward the Vim phil
 
 | Plugin Manager                                       | Command                                                                        |
 |------------------------------------------------------|--------------------------------------------------------------------------------|
-| [Packer](https://github.com/wbthomason/packer.nvim)  | `use 'TimUntersberger/neogit'`                                               |
+| [Packer](https://github.com/wbthomason/packer.nvim)  | `use 'TimUntersberger/neogit'`                                                 |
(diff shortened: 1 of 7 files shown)


TimUntersberger (approved):

RianFuro:
> If you think fixing this won't take too much time

Thing is, I stared at it for too long with no avail, I'm a bit frustrated by now. So I'd rather get back to it later with a clear mind. I left the comments in for that reason as well.

> Could you also update the README?

Whoops, completely forgot about that, will do!

RianFuro:
@TimUntersberger just to be on the same page, it's fine if I do the merge myself?

PR #52, opened 2021-03-14
feat: very basic branch control

RianFuro:
Add a branch popup to checkout local or remote branches and create new branches.
Add module to manage prompting the user for input.

Things to note:
- The input module is set up in a way so the api can be plugged by monkey patching it. I use this for mocking the module for testing purposes, but that basically screams for optional replacements to the default prompt using fzf or telescope.
- There's now a top-level module where neogit operations need to register into. This is mainly done for testing, so tests can more easily wait for the async execution to finish. I built this as a stop-gap; busted does support finishing the test via a callback, but that doesn't seem to have made it into the plenary port... yet? I feel like with the async pr coming, that might soon change, making the operations module obsolete.


TimUntersberger:
Damn you are going ham on these features 😮 

Thank you for your contributions 🙏

lua/neogit/lib/input.lua
```diff
@@ -0,0 +1,63 @@
+local M = {}
+
+if not _G.neogit then _G.neogit = {} end
```
TimUntersberger, line 3: `neogit` is in my opinion a bad name for a global library variable. Maybe prefixing the variable with 2 `_` will make it less likely to be used by users.

lua/neogit/lib/input.lua
```diff
@@ -0,0 +1,63 @@
...
+local function generate_random_string(len)
+  local result = ''
+  for _ = 1, len do
+    result = result .. chars[math.random(1, #chars)]
+  end
+  return result
+end
+
+local function make_completion_function(options)
+  local id = generate_random_string(5)
```
TimUntersberger, line 32: Why can't we just use an incrementing id instead of generating a random string?

lua/neogit/operations.lua
```diff
@@ -0,0 +1,30 @@
+local a = require('neogit.async')
+-- This is a table to look up pending neogit operations.
```
TimUntersberger, line 2: What is your definition of an "neogit operation"?

lua/neogit/lib/git/branch.lua
```diff
@@ -0,0 +1,77 @@
...
+    .list
+    .all
+    .call())
+
+  return parse_branches(branches)
+end)
+
+local function prompt_for_branch(options)
+  local chosen = input.get_user_input_with_completion('branch > ', options)
+  if not contains(options, chosen) then
```
TimUntersberger, line 46: When being inside the prompt and hitting `ESC` the input should be cancelled in my opinion. Currently I also get the error message when trying to cancel the prompt.

lua/neogit/lib/git/branch.lua
```diff
@@ -0,0 +1,77 @@
...
+
+M.checkout = a.sync(function ()
+  local branches = a.wait(get_all_branches())
+
+  a.wait_for_textlock()
+  local chosen = prompt_for_branch(branches)
+  a.wait(cli.checkout.branch(chosen).call())
+end)
+
+M.checkout_new = a.sync(function ()
```
TimUntersberger, line 69: Why do you think an extra binding for specifically creating a new branch is good if you could also create a branch by using the other two bindings?

TimUntersberger:
> I use this for mocking the module for testing purposes, but that basically screams for optional replacements to the default prompt using fzf or telescope

How and where would you imagine having telescope support? I can't really imagine usable places right now.

clason:
> > I use this for mocking the module for testing purposes, but that basically screams for optional replacements to the default prompt using fzf or telescope
> 
> How and where would you imagine having telescope support? I can't really imagine usable places right now.

Anywhere you'd be presented with a list of more than, say, three to five entries you'd like to act on. See the [builtin git pickers](https://github.com/nvim-telescope/telescope.nvim/wiki/Showcase#git-pickers) for inspiration, as well as how https://github.com/pwntester/octo.nvim uses telescope for its UI.

lua/neogit/lib/input.lua
RianFuro: good point well made :D no idea why i went straight to random string there, especially with how cumbersome it is in lua

lua/neogit/operations.lua
RianFuro: that's a very loose notion there, i basically just mean "top-level operations", those that are directly invoked by the end user. Think "commit", "switch branch", "stash", etc. As mentioned it's really just a stop-gap implementation, since I need a way to wait for those operations from the test execution.

lua/neogit/lib/git/branch.lua
RianFuro: Good catch, haven't even tried cancelling :X

lua/neogit/lib/git/branch.lua
RianFuro: Mnemonics for now, mostly. 
I'm also not entirely sure on if we can and want to allow branch creation from the other dialogs as well. For "checkout anything" (`bb`), since it's supposed to checkout anything, we'd need to check against branches, tags and commit hashes if we want to create a branch on other inputs; at that point, it might be easier to just throw it into `git branch ...` and let the error bubble up.
For the distinction between checkout local and create new branch, it's pretty much just the choice of dialog: For checkout local we provide a dialog with completion options, for create new just prompt for any text. Not much of a distinction, but the former could potentially be plugged by telescope, while it doesn't make that much sense for the latter.

lua/neogit/operations.lua
TimUntersberger: Could you please add this to the comment? I will probably forget what this was about in 1 month or so 😆

TimUntersberger (approved):

PR #65, opened 2021-03-22
Add cmd arguments

p00f:
This allows 0 or 1 arguments to `:Neogit`, so `:Neogit` retains default behaviour but `:Neogit floating` now uses a floating window

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -700,7 +700,7 @@ local function create(...)
   end
 
   kind = kind or 'tab'
-  if kind ~= 'tab' or kind ~= 'floating' or kind ~= 'split' then
+  if kind ~= 'tab' and kind ~= 'floating' and kind ~= 'split' then
     vim.api.nvim_err_writeln("Invalid kind")
     return
   end
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -688,23 +688,7 @@ local discard = a.sync(function()
   __NeogitStatusRefresh(true)
 end)
 
-local function create(...)
-  local args = {...}
-  local kind = nil
-
-  for _, val in pairs(args) do
-    local kindpos = string.find(val, "kind=")
-    if kindpos then
-      kind = string.sub(val, 6)
-    end
-  end
-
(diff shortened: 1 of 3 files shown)


TimUntersberger:
Could you change the syntax from `Neogit floating` to `Neogit kind=floating`? Simple parsing is enough, but we need the syntax change to make future subcommands easier to distinguish.

p00f:
Sure!

p00f:
How would it work though, I think you'd have to make `neogit.status.create()` take a table

TimUntersberger:
For now just pass in the value of kind directly

p00f:
Can I do the parsing in the lua function? Vimscript regex is very hard

p00f:
https://user-images.githubusercontent.com/36493671/112026931-b54e6d00-8b5c-11eb-84ca-b914deaebcc7.mp4

p00f:
(I did the parsing in the lua function, I couldn't figure out how to match things in vimscript)

TimUntersberger:
> Can I do the parsing in the lua function? Vim regex is very hard

Parsing in lua is 100% prefered. 

Obviously the name I chose for the function is just a placeholder and you can change it to whatever you think fits best.

lua/neogit/status.lua
```diff
@@ -688,8 +688,23 @@ local discard = a.sync(function()
...
+local function create(...)
+  local args = {...}
+  local kind = nil
+
+  for _, val in pairs(args) do
+    local kindpos = string.find(val, "kind=")
+    if kindpos then
+      kind = string.sub(val, 6)
+    end
+  end
```
TimUntersberger, line 700: Sorry, but could you extract the parsing out of the create function and change the signature back to how it is right now? You can then change mapping to:

```lua
require'neogit.status'.create(require'neogit.util'.parse_command_args(<f-args>))
```
Obviously the name I chose for the function is just a placeholder and you can change it to whatever you think fits best.

TimUntersberger (changes requested):

TimUntersberger (approved):
Thank you!

Issue #67, opened 2021-03-22
Add visual feedback for collapse content


TimUntersberger:
As a user I would like to see which content can be toggled without having the manually try `<tab>` and see which content is open/closed.

TimUntersberger:
We should probably copy what magit does with signs (? Idk what to use for this). I think they use arrows by default, but I'd like for this to be configurable.

RianFuro:
Try doing `set foldcolumn=1` inside the neogit status buffer ;)

TimUntersberger:
Can we change what gets used to display the markers? I couldn't find anthing while looking through the docs.

akinsho:
You can set fill chars for the fold column, unfortunately it doesn't support a character at the end. It's under `:h fillchars`.
```
	  foldopen:c	'-'		    mark the beginning of a fold
	  foldclose:c	'+'		    show a closed fold
	  foldsep:c	'│' or '|'      open fold middle marker

```

Odie:
You might want to try `sign_place()`. You can basically put whatever character you want in the gutter area. =)

TimUntersberger:
I added support for this in the latest commit. Should this be enabled by default? I myself feel like the markers are actually unnecessary after implementing them (might be because I have used this plugin a lot).

akinsho:
@TimUntersberger another heads up haven't gotten round to creating an issue yet as I'm a little tight for time today, but I was going to mention that the signs actually go out of sync whilst in use say you incrementally stage several hunks that are open. When you do commit and the hunks are removed some of the signs linger. You then end up with signs not associated with anything.

PR #72, opened 2021-03-23
Keeping folds open

### LINKED ISSUE #48: Errors when git status buffer is open in a repo on a branch without an upstream
Thanks for your work on this plugin 🚀 I'm really excited to see where things go 💯 .

I just wanted to note a bug I encountered whilst using it (I appreciate it's WIP but thought you might want to have this tracked). When calling `NeogitStatus` whilst on a branch without an upstream I get repeated command error popups. Looking at the log via `$` I see
```
 128 git log --oneline @{upstream}.. --
  | fatal: no upstream configured for branch 'feature/multiple-carer-notes'
 128 git log --oneline ..@{upstream} --
  | fatal: no upstream configured for branch 'feature/multiple-carer-notes'
  0 git log --max-count=1 --pretty=%B --
  | fix: refetch the token for every request
  | 
  | This prevents it going out of sync when logging out and in to the app as
  | different users
 128 git log --max-count=1 --pretty=%B @{upstream} --
  | fatal: no upstream configured for branch 'feature/multiple-carer-notes'
  0 git status --porcelain=1 -b --
  | ## feature/multiple-carer-notes
  |  M lib/services/api.dart
```
When I run the erroring commands in a terminal it prints

```
git log --oneline ..@{upstream}.. --
fatal: no upstream configured for branch 'feature/multiple-carer-notes'
```
I wonder if there should be a check for if there is a valid upstream before upstream related commands are tried? not sure the best way round this based on how this plugin is structured but in this case I don't really think this is a real error in this case since it's expected as I just created the branch and haven't pushed it yet

EDIT: also not sure if it's intended but `--` is being appended to all commands it looks like without any further arguments not sure if that's intentional or not

### LINKED ISSUE #85: Signs diplayed incorrectly
## Steps to reproduce

1. Make two changes in two different files:
image

1. Unfold second file
image

1. Navigate via `k` (important!) to the first file and unfold it too.

## Expected behavior
All signs looks unfolded
image

## Actual behavior
One of the signs looks unfolded
image

## Additional notes

If you move with the mouse in step 3, the behavior of the signs will be as expected.

RianFuro:
With this pr we completely change how folds are managed. We no longer rely on vim's folding mechanism and instead manage folded content ourselves in memory. The buffer is completely rewritten on change and renders only unfolded content.
This plays nicely with #67 since we now manage both the foldmarkers and the folds ourselves.

Additionally, the data loading mechanism has been updated. There is now a central `repository` abstraction, which encapsulates data loading. I initially thought I needed to be more selective with refetching data to improve performance, so I updated the data structure to accomodate for that, but that hasn't been a problem throughout, so it's just structural improvement for now. It now allows for more granular cache updates if performance ever becomes a problem though.

Additionally, this PR fixes a few more minor issues:
- [x] Closes #48 
- [x] Closes #85 
- [x] Fixes an outstanding problem in #67 
- [x] Fixes the settings bleed problem from #95


akinsho:
@RianFuro regarding extmarks I should probably have mentioned in the issue but had forgotten at the time. `nvim_buf_set_lines` is due to be refactored I think since it doesn't work well with the type of byte ranges extmarks use (if I understand correctly) so actually using `:h nvim_buf_set_text()` is recommended instead, but think that's largely for inline changes. Anyway just as an FYI (since you aren't currently using them).

RianFuro:
I open this up for review because I think it would be an improvement over the current state, however I'm not really happy with it, largely because vim is really weird with it's fold state tracking under the given circumstances.

I'll give the extmarks one more try and see if maybe it would be better to completely manage the fold state internally and not have actual folds at all. This would of course mean that usual fold commands (`z*`) will no longer work, but at this point I think it could be worth the tradeoff.

@TimUntersberger & @akinsho I would appreciate if you could try out the pr and tell me what you think, maybe I'm just overthinking this.

akinsho:
@RianFuro been giving it a go and so far it's a much better experience 👍🏿 . I think there might be a few quirks but will wait till I can concretely report back on what's going on.

Regarding the folds or not folds I *personally* think their actually being is probably more of a moot point for me. I think toggling hunks and them staying in positions is all I really need it to do +/- ability to toggle all of a block potentially.

I think it being tied to vim folds can also mean in the long term things are limited by the behaviour and quirks of folds 🤷🏿  like the lack of highlighting for the folded line whilst that's being refactored etc.

I think gitabra uses a custom implementation for the "folds" which is pretty interesting looking.

akinsho:
@RianFuro just noticed a bug during usage. Whilst trying to discard a section (unstage) I hit
```
E5108: Error executing lua ...re/nvim/site/pack/packer/opt/neogit/lua/neogit/async.lua:76: ...e/nvim/site/pack/packer/opt/neogit/lua/neogit/status.lua:599: attempt to concatenate field 'name' (a nil value)
```

RianFuro:
> just noticed a bug during usage

whoops, thanks for the heads-up :)

> I personally think their actually being folds is probably more of a moot point for me.

That's exactly why I'm asking; I like the existing fold commands and use them quite heavily, so I'm kind of hung up on the issue. But if others are indifferent on that, there's really no reason for me to stick to that.

> I think gitabra uses a custom implementation for the "folds" which is pretty interesting looking.

Yeah, they manage the fold state locally and apply it when rendering. I'll see if I can whip up something similar; that also works completely without extmarks, so it should work without issue for once :grin:

Odie:
> are you using folds in your plugin or are you manually managing the collapsed content and why did you decide one of the options?

@TimUntersberger 
With gitabra, I also tried using vim’s builtin fold facilities. I found I was spending too much time on coercing folds to work exactly the way I wanted, which was just a simple outline view.

I ended up manually managing the buffer contents by keeping everything in a (document?) tree. Vim is only used as the presentation layer. On refresh, the buffer contents are blown away and rebuilt from the tree. The tree itself might also be rebuilt/modified to match git’s output.

This approach turned out to be quite liberating.  Node collapse states are can no longer hidden behind vim. You can also attach whatever additional info into the nodes. The makes the type of the item under the cursor very clear as well as the types and info of the parents/ancestors.

This approach also turned out to be surprisingly fast. I’m not sure if this is due to some neovim or luajit magic, but I really don’t remember having to perform any optimizations. Buffer rebuilds take only somewhere around 1ms during some informal tests.

RianFuro:
> With gitabra, I also tried using vim’s builtin fold facilities. I found I was spending too much time on coercing folds to work exactly the way I wanted, which was just a simple outline view.
> 
> I ended up manually managing the buffer contents by keeping everything in a (document?) tree
> ...
> This approach turned out to be quite liberating.

After going through all the alternatives myself, I feel you :sweat_smile:

Odie:
Btw, do feel free to take some code from gitabra if find it helpful. It’s MIT licensed for a reason. :)

akinsho:
@RianFuro another thing I noticed not sure if it regressed at some other point but I've only seen this since I switched to your branch but I'm getting #48, no upstream branch is causing the git log command to fail which is throwing up error notifications

RianFuro:
Ok, next (and hopefully final) attempt.
- Folds are now done "manually" much like gitabra does.
- The render thus looks closer to it's original state, wiping the buffer and redrawing it on each refresh.
- Fold state is kept with the location data; toggling with tab simply updates the state and redraws the buffer as well.

I also included a copy of [md5](https://github.com/kikito/md5.lua) in the repo; I needed some way to identify a hunk over multiple updates and hashing the content seemed the best approach - thanks @Odie for the idea :grin: 

I have a few lose ends to clean up now, where normal folds are still assumed (`1-4` foldlevel manipulation and the couple test cases I think), but otherwise this looks the most stable approach so far. Should be able to wrap this up over the weekend

RianFuro:
Incidentally, since this now reworks how folds are handled, it should also fix #85 & @akinsho's comment in #67.

akinsho:
@RianFuro this looks good/works quite well. One thing I've noticed though is that it seems the cursor is kept in it's absolute position which say you had a large diff unfolded and then close it so the buffer content is more than halved your cursor ends up where there isn't relevant content anymore. I'm not sure entirely what the right solution is since it's technically right I think but somewhat unintuitive.

I think my assumption if I were in the unstaged section and I staged a large hunk is that my cursor would be where the old hunk used to have been if that makes any sense i.e. I'd expect to still be in unstaged rather than line say 70 or so where the cursor used to be.

Anyway not sure if this just me though

Odie:
I guess it’d make sense if the cursor to move to the next unstaged hunk?

akinsho:
Yeah I think that makes sense. I think ultimately as a user you probably expect the cursor to stay in the logical zone/area you are in. If you stage something you should stay with the rest of the unstaged things as opposed to move with the staged thing or end up at the line number where that thing would have been if it was still there. 

I think this is the same as how fugitive works. since if you stage something in fugitive you just stay in that section.

Also just to mention. I'm still seeing #48, despite the latest changes 
image

RianFuro:
@akinsho Thanks for the heads-up! That's a repository with an origin, but the current branch has no upstream, right?

akinsho:
@RianFuro yep that error happens when you create a new branch in the repo but don't immediately set an upstream. I created another branch called `test` the output you requested was
```
# branch.oid 788d80f364b4f269703e6abaa9c15665aba7ef0e
# branch.head test
```

lua/neogit/lib/collection.lua
```diff
@@ -0,0 +1,42 @@
+local M = {}
+
+function M.keyBy(tbl, key)
```
TimUntersberger, line 3: I'm not really a fan of the name as it was a bit confusing at the beginning without looking at the code. Could you change this to a more specific name (and also try to keep the names snake case like the rest of the code base)?

lua/neogit/lib/functional.lua
```diff
@@ -0,0 +1,15 @@
+local util = require 'neogit.lib.util'
+local collect = require 'neogit.lib.collection'
```
TimUntersberger, line 2: Initially I thought `collect` was a function, but this seems more like a constructor right? Could we PascalCase the variable if it is a constructor function?

Is the name supposed to be a short version of Collection or the verb?
```suggestion
local Collect = require 'neogit.lib.collection'
```

lua/neogit/status.lua
```diff
@@ -64,390 +69,191 @@ local function get_section_item_for_line(linenr)
...
+  return setmetatable({ }, {
+    __index = {
+      append = function (tbl, data)
+        if type(data) == 'string' then table.insert(tbl, data)
+        elseif type(data) == 'table' then
+          for _, r in ipairs(data) do table.insert(tbl, r) end
+        else error('invalid data type: ' .. type(data), 2) end
+        return tbl
+      end,
+      into_buffer = function (tbl, from, to)
```
TimUntersberger, line 89: When I read something with `into....` I always think about a conversion and not a side effect. Could it be possible to change this name?

lua/neogit/status.lua
```diff
@@ -64,390 +69,191 @@ local function get_section_item_for_line(linenr)
...
-
-
-    if on_hunk then
-      local ignored_sections = { "untracked_files", "stashes", "unpulled", "unmerged" }
-
-      for _, val in pairs(ignored_sections) do
-        if val == section.name then
-          return
-        end
+local function new_output()
```
TimUntersberger, line 79: Same thing as with `collect`. Could we also do some metatable magic here like with collect? Also the name `output` seems very specific even though the implementation could be reused easily. Isn't this more something like a string buffer of builder?

lua/neogit/status.lua
TimUntersberger: After thinking about this some more I don't really think that `into_buffer` should be a part of this class (I guess we can this lua metatable magic classes?).

lua/neogit/status.lua
```diff
@@ -64,390 +69,191 @@ local function get_section_item_for_line(linenr)
...
-        c.diff_content = git.diff.parse(vim.split(results[i], '\n'))
-        break
-      end
-    end
-  end
-end)
+  refresh_status()
+end
 
+function __NeogitStatusReset()
```
TimUntersberger, line 252: Also do you think we should remove these global `__...` functions and refactor the code that uses these to just use something like this: `require('neogit').status.reset()` ?

TimUntersberger (changes requested):

TimUntersberger:
A lot of nitpicks in this one 😄

lua/neogit/lib/collection.lua
RianFuro: Sure on the snake case (guess I got confused for a second there). How would you call it though? I only know the method by that name ([here](https://laravel.com/docs/8.x/collections#method-keyby) is where I got it from), so I'm kinda biased here.

lua/neogit/lib/functional.lua
RianFuro: The idea is that it's the verb. I'm kinda mixing paradigms here:  The structure is class-like, but the invocation is more like a builder pattern. We can totally change the initial invocation to `Collection:new()` if you feel that's more readable.

lua/neogit/status.lua
RianFuro: Yeah that thing is a mess :sweat_smile: Currently it's just a helper decorator that wraps the lines i felt were too verbose; it can't really exist outside of the context it's used right now as well, because it closes around the status buffer.
As for the `into_buffer` method, how does a `replace_content_with`-method on the buffer object sound to you instead? :)

lua/neogit/status.lua
RianFuro: Totally agree. We do need true globals sometimes (`v:lua` only takes globals), but I much prefer the `__NEOGIT` object I started for that anyway, since it scales better in that regard.

RianFuro:
> A lot of nitpicks in this one

Not nearly as much as I thought, given the scope of my changes :sweat_smile: Thanks for the feedback, really appreciate it!

lua/neogit/lib/collection.lua
TimUntersberger: Hmm, maybe this is more of a convention thing? I guess if you just add some comments explaining what this does it would be good 👍

lua/neogit/lib/functional.lua
TimUntersberger: > We can totally change the initial invocation to Collection:new() if you feel that's more readable.

I would really appreciate that! 👍 

> Collection:new()

Shouldn't it be `Collection.new()` ?

RianFuro: Yeah it should. I just got stuck on that as well while pulling out that `output` thingy into a separate thing. I'm really not used to that operator :sweat_smile:

lua/neogit/status.lua
```diff
@@ -64,391 +70,245 @@ local function get_section_item_for_line(linenr)
...
-  for _,l in pairs(locations) do
-    if l.first <= line and line <= l.last then
-      vim.api.nvim_win_set_cursor(0, { l.first, 0 })
-      break
-    end
+local function restore_cursor_location(section_loc, file_loc, hunk_loc)
+  if #locations == 0 then return vim.fn.setpos('.', {0, 1, 0, 0}) end
+  if not section_loc then section_loc = {1, ''} end
+
+  local section = Collection.new(locations):find(F.C(F.dot('name'), F.eq(section_loc[2])))
```
RianFuro, line 251: @TimUntersberger raising this one myself, since I know this can raise an eyebrow or 2 :sweat_smile:
Am I overdoing it here? I'm used to thinking this way and thus tend to solve problems through function composition; so I thought I'd ask with a trivial problem (the composition can easily be replaced by `function (i) return i.name == section_loc[2] end`) if code like this is alright with you.

TimUntersberger: I am a fan of functional programming, but most of the time it overcomplicates simple things just like in this case imo.

RianFuro: I do agree on the sentiment, but I feel *overcomplicated* is the wrong way to put it.
For example, given fsharp where I can easily express the same logic in both paradigms:
```fsharp
dot "name" >> eq section[2]
fun x -> x.name == section[2]
```
both look equally complex to me.

Lua is not nearly as expressive though, so I need to add a ton of **noise** to represent function composition. So, while writing the function out normally is actually *longer* than using my composition helper, it looks cleaner, since more of the elements used convey useful meaning.

I'll put in normal functions there, but I think I'll keep the composition function around; since you're up for it in principle, it will be useful some other time I hope :grin:

RianFuro:
@TimUntersberger I'm done I think. On your OK we can merge :)

TimUntersberger (approved):

Shatur95:
@RianFuro, can `setlocal fillchars=fold:\ ` be removed after your changed from the filetype options?

RianFuro:
@Shatur95 You're right, thanks for the heads-up :)

PR #75, opened 2021-03-24
Remove extra padding from buffer actions

### LINKED ISSUE #74: Trailing whitespace in Neogit Buffers
This is a very minor point 😅 .

Neogit buffers all seem to contain some trailing whitespace in the output. This is evident when using `listchars=trail:-` or some other value.

image

Ordinarily I'd add this to my `after/ftplugin` but this happens in most of neogit's buffers not all of which have filetypes, and even if they did would be kind of tedious to target all the various buffers to hide the trailing spaces. I don't know if it's just a minor rendering artefact or needed somehow or complicated to remove.

I'd be willing to take a stab at removing them if someone could point me to where they might be being added. Seems likely to be in one spot given it happens in almost every popup

akinsho:
This PR fixes #74, by checking whether or not the action being rendered has a neighbour or not and if not padding it or not. I've tested this in most of the buffers and don't think it broke anything. Also wasn't sure of the rules re formatting but I've indented some long lines with the extra conditionals

### DIFF
--- a/lua/neogit/lib/popup/lib.lua
+++ b/lua/neogit/lib/popup/lib.lua
@@ -79,16 +79,24 @@ local function draw_popup(popup)
 
   for i=1,actions_grid_height do
     local result = " "
-    for _, col in pairs(columns) do
+    for index, col in pairs(columns) do
       local item = col.items[i]
+
+      local next_col = columns[index + 1]
+      local has_neighbour = next_col and col.items and item and next_col.items[i]
+
       if item == nil then
-        local key = util.str_right_pad("", col.k_width + 1, " ")
-        local description = util.str_right_pad("", col.d_width + 6, " ")
+        local key = next_col and util.str_right_pad("", col.k_width + 1, " ") or ""
+        local description =  next_col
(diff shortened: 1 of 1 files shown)


lua/neogit/lib/popup/lib.lua
```diff
@@ -79,16 +79,26 @@ local function draw_popup(popup)
...
   for i=1,actions_grid_height do
     local result = " "
-    for _, col in pairs(columns) do
+    for index, col in pairs(columns) do
       local item = col.items[i]
+
+      local next_col = columns[index + 1]
+      local has_neighbour = (next_col and col.items)
+        and (item and next_col.items[i])
+        or false
```
RianFuro, line 88: This looks like you wanted to build a ternary, but it ends up just being a boolean calculation.
Should be equivalent to just
```lua
next_col and col.items and item and next_col.items[i]
```
right?

akinsho: Yep, that's right forgot to condense it down from an earlier iteration

RianFuro:
lgtm! Thanks for your contribution :)

PR #86, opened 2021-03-27
chore: move notification hl to syntax-file

### LINKED ISSUE #79: Notification highlights no longer applied
Since e321022e02092a63d5f352bd8db89c61ab86ea2d notification highlights no longer seem to be applied. I see they were moved into the `colors/` directory which I'd have assumed would work like `ftplugin` or `syntax` but does not I don't think. I couldn't find much information on it in the docs but from a brief look online at a few plugins it seems that *colorschemes* tend to place their code in that folder but other plugins that apply some highlighting use ftplugin or syntax. Placing the commands in either subdir for neogit in my config re-applies the highlights.

Alternatively this could be an issue related to the colors dir and using `:h packages` but I think this is less likely since my colorscheme which uses that dir works fine.

As another small test I created a `colors` dir in my `&rtp` and it doesn't load the vimscript file in the colors directory.

RianFuro:
Closes #79 

Doing this one as a PR because I'm not entirely sure this is the right way to do this either. The highlights could just as well go into a `ftplugin`-file or just into our `neogit.vim`. I'm confused :sweat_smile:

### DIFF
--- a/colors/neogit.vim
+++ b/colors/neogit.vim
@@ -1,3 +1,0 @@
-hi NeogitNotificationInfo guifg=#80ff95
-hi NeogitNotificationWarning guifg=#fff454
-hi NeogitNotificationError guifg=#c44323
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -24,9 +24,6 @@ local neogit = {
     end
   end,
   setup = function(opts)
-    vim.cmd("hi NeogitNotificationInfo guifg=#80ff95")
-    vim.cmd("hi NeogitNotificationWarning guifg=#fff454")
-    vim.cmd("hi NeogitNotificationError guifg=#c44323")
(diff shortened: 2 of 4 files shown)


akinsho:
@RianFuro this is from the nvim docs
```
2. Syntax files						*:syn-files*

The syntax and highlighting commands for one language are normally stored in
a syntax file.
```

So I think even though it's a bit ambiguous this is the right place.

TimUntersberger:
Moving this into its own syntax file is properly the cleanest/easiest solution.

TimUntersberger (approved):

RianFuro:
> highlighting commands for one language

This is the part that makes me a bit unsure, since we're not dealing with a *language* really. But I guess it doesn't matter either way and since we're all on the same page here, that's how it will be :grin:

Issue #10, opened 2020-09-16
Keep folds open when staging


hkupty:
When debugging #7 , I realized Neogit constantly falls back to mode `2` after I stage the hunks. I guess the ideal would be to not revert, since one can probably add another hunk from that same file, for example. An alternative would be to make the behavior configurable.

akinsho:
Just wanted to bump this issue since it's a little old now and this is a daily frustration for me as a new neogit user. If someone wants to point out what would be needed to fix this I can have a look when I get the time/I'm tipped over the edge by my frustration. Alternatively I can wait for a fix if it's better done by someone with know how about the code base

TimUntersberger:
This probably is because we redraw the whole buffer after modifying the status.\

@RianFuro What do you think?

RianFuro:
Yeah that's part of the problem. Another challenge we need to solve is that even if we only selectively redrawing the buffer, we still need to potentially update/replace the diffs inside the folds without breaking the fold itself. Well, or restoring the fold status after a redraw i guess.
I'd like to avoid manually updating the buffer to match the operation result, so we'd need some logic to calculate a delta from the new status (after a operation or reload) against the current one and then redrawing parts of the buffer based on that delta. Think shadow dom handling, basically.

I definitely feel the frustration just as much though :sweat_smile: Going to have a look at that soon-ish

akinsho:
This might be a red herring and I apologise if it is I haven't read through all the code myself, but I think https://github.com/wbthomason/packer.nvim uses nvim extmarks to keep track of items and update each specifically in it's UI as plugins are being updated in parallel, the code is in its `display.lua` module. That might be a useful insight for what you're describing

hkupty:
I think using extmark is the way to go, but from what I remember from the code, it might need a refactoring on how things are presented.

I guess the idea would be to keep track of the state between redraws and take in consideration the previously set fold level on the first pass directly.

hkupty:
Feels like this has been sorted out. I'm fixing it. Thanks!

PR #94, opened 2021-03-31
Fix issue when cwd is not git root directory

### LINKED ISSUE #90: Committing does not work for git worktrees
I must admit that I'm not 100% about this. I just started to use this plugin. On repositories which have no commit hooks, I can nicely commit as expected. But when there are such hooks I observe the following:
When I type `cc` in the status window to start a commit, I get the error `Error executing vim.schedule lua callback: ...site/pack/packer/opt/neogit/lua/neogit/popups/commit.lua:20: Vim(write):E212: Can't open file for writing: not
a directory`.  If I then hit `return` I still get the commit message window. But as a result of the former error on save the error `E212: Can't open file for writing: not a directory` occurs.
I must admit I'm not too sure why the Lua function fails. But could it be that it does not expect the hooks to execute? On fugitive I was seeing the shell output of the pre-commit hooks and once done I could type my commit message. If a hook failed I got navigated back to the status window and the message was printed.
So atm this is a blocker for me to finally make the cut and use Neogit exclusively over Fugitive.

### LINKED ISSUE #91: Commit fails when cwd is not the repo root
# Steps to reproduce

```bash
mkdir testrepo && cd testrepo
git init
mkdir foo && cd foo
touch bar && git add bar
nvim +"lua require'neogit'.open({ 'commit' })" +"norm c"
```

The following error message shows up:

```
Error executing vim.schedule lua callback: ...te/pack/packer/start/neogit/lua/neogit/popups/commit.lua:20: Vim(write):E212: Can't open file for writing: no such file or directory
```

The commit window still opens and I can edit it, but when doing `:write` it fails with:

```
".git/NEOGIT_COMMIT_EDITMSG"
E212: Can't open file for writing: no such file or directory
```

This looks like `neogit` is expecting the `.git` directory to be directly acessible from the cwd, when that is necessarily not the case:

https://github.com/TimUntersberger/neogit/blob/2f504fcf734ddb6b281a93e01782ecb0dae0774e/lua/neogit/popups/commit.lua#L9

# Possible solutions

* Internally change `cwd` when calling `neogit.open({ "commit" })` (maybe using `lcd`?)
* Make neogit directly open `<project_root>/.git/COMMIT_EDITMSG` (this looks like the better long term option to me since it does not require changing any cwd)

Fugitive does not have this issue.

# Use case: why would one ever want to do this?

In big repositories (like monorepos), it's often not pratical always work from the project root due to the huge amount of unrelated files which just clutter and slow down fuzzy-finding tools.

Meanwhile, is there any config workarround I can use?

## Edit

Possibly related: https://github.com/TimUntersberger/neogit/issues/90

hkupty:
Address issue of assuming {,t,l}cd is root folder.

Fix #90
Fix #91

---

The function introduced in neogit/popups/commit.lua can be moved elsewhere if needed, but seemed to make sense to keep it local for the time being.

### DIFF
--- a/lua/neogit/popups/commit.lua
+++ b/lua/neogit/popups/commit.lua
@@ -6,11 +6,26 @@ local a = require('neogit.async')
 local uv = require('neogit.async.uv')
 local split = require('neogit.lib.util').split
 
-local COMMIT_FILE = '.git/NEOGIT_COMMIT_EDITMSG'
+local find_root = function()
+  local base_pwd = vim.fn.getcwd()
+  local target = string.byte("/")
+  for idx = #base_pwd, 1, -1 do
+    if base_pwd:byte(idx) == target then
+      local gitpath = base_pwd:sub(1, idx) .. ".git"
+      if vim.fn.isdirectory(gitpath) > 0 then
+        return gitpath
+      end
+    end
+  end
+  return ".git"
+end
+
+local commit_file = function() return find_root() .. '/NEOGIT_COMMIT_EDITMSG' end
 
 local get_commit_message = a.wrap(function (content, cb)
+
   Buffer.create {
-    name = COMMIT_FILE,
+    name = commit_file(),
     filetype = "gitcommit",
     buftype = "",
     modifiable = true,
@@ -82,6 +97,7 @@ local prompt_commit_message = a.sync(function (msg)
 end)
 
 local function create()
+  local COMMIT_FILE = commit_file()
   popup.create(
     "NeogitCommitPopup",
     {


lua/neogit/popups/commit.lua
```diff
@@ -6,11 +6,26 @@ local a = require('neogit.async')
...
+  for idx = #base_pwd, 1, -1 do
+    if base_pwd:byte(idx) == target then
+      local gitpath = base_pwd:sub(1, idx) .. ".git"
+      if vim.fn.isdirectory(gitpath) > 0 then
+        return gitpath
+      end
+    end
+  end
+  return ".git"
+end
```
RianFuro, line 21: Just wondering, since separate worktrees for git do not need to be in a subdirectory of the repository, does this actually solve #90?
> So for example the main local repo is at /home/user/projects/foo and a worktree of it is in /home/user/projects/foo-prs, then the .git filesystem entry in the latter repo is a file that includes /home/user/projects/foo/.git/worktree/foo-prs.

Haven't tried it, but this doesn't look like it would be solved by above function.

Also, we already have a function to get the repository root here:
https://github.com/TimUntersberger/neogit/blob/2f504fcf734ddb6b281a93e01782ecb0dae0774e/lua/neogit/lib/git/cli.lua#L185-L187
I wonder if this one has the same problems

ruifm:
Just wanted to say this solves https://github.com/TimUntersberger/neogit/issues/91 for me

ruifm:
Commiting is not the only thing that needs the same fix (see https://github.com/TimUntersberger/neogit/issues/91#issuecomment-816886424)
I think a more general approach is needed.

hkupty:
I was out on vacations and just came back to this. I'm feeling the proper way to handle this would be to set up configuration object which contains the state of neogit for that session and it'd hold, for example, the root cwd. #10 would also benefit from this.

Given this PR doesn't properly fix the issue, I'm inclined to close it instead of trying to fix it. wdyt?

RianFuro:
@hkupty I still think the PR has merit. The thing is that all these similar problems are actually slightly different:
- The file paths for opening a file are given from the *worktree*-root
- The .git folder where we need some files from is in the *repository*-root
- The file path for opened buffers are given from the *cwd* of the window

Each of these needs a slightly different solution, enough that we probably cannot have a single solution for all of our problems. So, imo, just fixing commits for now is very valid. If we can additionally lay some ground work for our other problems then all the better.

hkupty:
I'll fix it to use the `git_root` function from `lib/git/cli.lua` then.

RianFuro:
> I'll fix it to use the git_root function from lib/git/cli.lua then.

Which I'm afraid **does** have the same problem and actually resolves to the worktree-root (see https://github.com/TimUntersberger/neogit/issues/90#issuecomment-813972969). To solve #90 we'd also need to check if we're in a separated worktree and resolve back to the repository root if so.

kamalmarhubi:
Re
>     * The .git folder where we need some files from is in the _repository_-root

copying my comment https://github.com/TimUntersberger/neogit/issues/91#issuecomment-841896769 here since it's relevant:

> Drive-by, but one should not assume that `.git` exists at the top-level (since you can set `GIT_DIR` to any path), or that it is a directory (if using [worktree](https://git-scm.com/docs/git-worktree)s). The best thing to do is to run `git rev-parse --git-dir` to have git tell you where it is.

TimUntersberger:
thanks to @kamalmarhubi both issues have been closed so I think we can close this PR?

PR #96, opened 2021-04-02
Add mappings to open a file in a split, vsplit, or tab

bigjazzsound:
It would be great to be able to open a file not just in the current buffer but also in a tab, split, or vsplit.


TimUntersberger:
Thank you!

PR #100, opened 2021-04-06
Split diff

### LINKED ISSUE #57: (Git) Split Diff viewer
90% of my fugitive use is `:Gdiffsplit`, it's the most useful git command (for me) by far. 

I would love to see a method for viewing (and stepping through) diffs. Happy to contribute to this if you think it is in scope.

Unrelated but @RianFuro @TimUntersberger do you all actively lurk on our gitter/IRC? I'm a huge fan of this plugin and would like to make sure we're doing whatever we can in core to support your work.

TimUntersberger:
Pressing `D` when over a change opens a new tab with 2 windows. 
You can also configure the diff view to be 2 floating windows.
Both windows are inside a vimdiff (`:h vimdiff`).

You can only modify the left side.

TODO:

- [ ] Writing (:w) one of the buffers applies the changes
- [ ] Retain cursor position when modifying an opened file
- [x] Map `<C-W>h` and `<C-W>l` buffer scoped for easy switching between windows
- [ ] Add default mappings for common diff command usages (like `:diffput`)
- [ ] Add syntax highlighting to diff
- [x] Change to plenary async once #101 is merged
- [ ] set the buffer name
- [x] Support custom mappings

Closes #57


akinsho:
@TimUntersberger not sure if this is relevant to what you are doing at all or not, but neovim merged in official support for borders a little while ago (few weeks maybe) so you *maybe* don't need to manually create your own borders. Might not be the case depending on what you're doing but just thought I'd mention in case it's helpful 🤷🏿‍♂️

TimUntersberger:
@akinsho oh cool! Under which help page can I find information on this?

akinsho:
@TimUntersberger I think the docs have been lagging behind, the PR is https://github.com/neovim/neovim/pull/13998. The TLDR is that the config object now takes a border key. I can link an example in a little while

RianFuro:
As a general note, to avoid having to rework everything later, I think you should already depend on plenary's async instead of our own. From my experience the 2 are largely compatible on the async/await level - alternatively, I'll get my branch up to shape later today, so you could rebase off of that one.

lua/neogit/diff.lua
```diff
@@ -0,0 +1,166 @@
...
+    vim.api.nvim_buf_call(state.rhs.buf, function()
+      vim.cmd [[au! * <buffer>]]
+    end)
+    vim.api.nvim_win_close(state.rhs.win, false)
+    state.rhs = nil
+  end
+  state.open = false
+end
+
+M.open = a.sync(function(lhs_info, rhs_info)
```
RianFuro, line 42: Is this method async for future use? There are no asynchronous calls in there except for textlock schedules, so you could also just not make this function async and wait for the textlock before calling it. Might also fix the oddity at the end of the function.

RianFuro:
> Map <C-W>h and <C-W>l buffer scoped for easy switching between windows

My initial impulse would have been to open the split diff in a new tab. Gives window movements for free and is generally also what tabs are designed for (holding specific window layouts).

Floating windows are definitely fancier though :smile:

akinsho:
@TimUntersberger here's the lagging docs PR https://github.com/neovim/neovim/pull/14302/files

the border docs section says
```
                                • `border`: style of (optional) window border. This can
                                either be a string or an array. the string
                                values are:
                                • "none" No border. This is the default
                                • "single" a single line box
                                • "double" a double line box
                                • "shadow" a drop shadow effect by blending
                                  with the background. If it is an array it
                                  should be an array of eight items or any
                                  divisor of eight. The array will specifify
                                  the eight chars building up the border in a
                                  clockwise fashion starting with the top-left
                                  corner. As, an example, the double box style
                                  could be specified as: [ "╔", "═" ,"╗", "║",
                                  "╝", "═", "╚", "║" ] if the number of chars
                                  are less than eight, they will be repeated.
                                  Thus an ASCII border could be specified as:
                                  [ "/", "-", "\\", "|" ] or all chars the
                                  same as: [ "x" ] An empty string can be used
                                  to turn off a specific border, for instance:
                                  [ "", "", "", ">", "", "", "", "<" ] will
                                  only make vertical borders but not
                                  horizontal ones. By default `FloatBorder`
                                  highlight is used which links to `VertSplit`
                                  when not defined. It could also be specified
                                  by character: [ {"+", "MyCorner"}, {"x",
                                  "MyBorder"} ]
```

[here's](https://github.com/akinsho/flutter-tools.nvim/blob/6655d1a2b739db096d7bc950a41ee1bd5ce8acdb/lua/flutter-tools/ui.lua#L188) an example of where I use it in one of my plugins

lua/neogit/diff.lua
TimUntersberger: Yeah I only made the function async to be able to use wait_for_textlock. I guess you are right that the responsiblity should lie on the caller side.

TimUntersberger:
@RianFuro how can we make an async function sync without having to wrap the outer function?

example
```lua
function f()
  local res = make_sync(async_f())
end

f()
```

RianFuro:
> how can we make an async function sync without having to wrap the outer function?

generally, with plenary's async implementation, `void` can be used to wrap an async into a "normal" function. In our implementation we have `dispatch` which admittedly is pretty awkard to use comparatively, since it takes a plain function and makes it async (compared to `void` which takes an already async'd function).

RianFuro:
https://github.com/TimUntersberger/neogit/blob/d716eed7e0e163aae4b556b10b70fba3578788d1/lua/neogit/status.lua#L654-L666

Here is a good example I think. (check the same part before my pr for our own async)

TimUntersberger:
I implemented file cycling for just unstaged changes for now just to see how it feels. 

* `]f` next file
* `[f` previous file.

Not to sure how/whether to handle cycling between sections.

RianFuro:
> Not to sure how/whether to handle cycling between sections.

since we only have empty lines between sections, `{` and `}` have done the trick for me so far in that regard :)

TimUntersberger:
@RianFuro I am talking about the kind of cycling like in the below video.


https://user-images.githubusercontent.com/32014449/115081920-5682d600-9f05-11eb-8526-dbb6aed65de3.mp4

RianFuro:
Aaah, I'm completely on the wrong topic :sweat_smile: Time for some sleep I guess

lua/neogit/status.lua
```diff
@@ -638,18 +638,18 @@ local cmd_func_map = function ()
...
+    end)),
+    ["Depth2"] = void(async(function()
+      await(set_folds({ false, true, false }))
+    end)),
+    ["Depth3"] = void(async(function()
+      await(set_folds({ false, false, true }))
+    end)),
+    ["Depth4"] = void(async(function()
+      await(set_folds({ false, false, false }))
+    end)),
```
RianFuro, line 652: This shouldn't change anything afaik. Am I missing something?

lua/neogit/lib/git/cli.lua
```diff
@@ -443,7 +452,8 @@ local cli = setmetatable({
     local calls = {...}
     return new_parallel_builder(calls)
   end,
-  git_root = git_root
+  git_root = git_root,
+  git_root_sync = void(git_root)
```
RianFuro, line 456: This doesn't do anything. `void` doesn't *block* until the future completes, it just makes sure it's running to completion sometime in the future. Because of that, it's also not actually returning the result of `git_root` like that.

RianFuro (changes requested):

lua/neogit/lib/collection.lua
```diff
@@ -58,8 +58,15 @@ function M.reduce(tbl, func, ...)
...
+  end
+  return nil
+end
+
+function M.at(tbl, idx)
+  for i, item in ipairs(tbl) do
+    if i == idx then return item end
   end
   return nil
 end
```
RianFuro, line 72: Is this any different from just doing `tbl[idx]`? :sweat_smile:

lua/neogit/lib/git/cli.lua
TimUntersberger: Ah whoops.

lua/neogit/status.lua
TimUntersberger: If I don't wrap these functions in async I get the c-bound error.

lua/neogit/lib/collection.lua
TimUntersberger: true

lua/neogit/status.lua
RianFuro: Aaah and it needs to be async because i made the refresh async and forgot to make the `set_folds` method async too. I was missing something quite big, apparently :sweat_smile:

lua/neogit/lib/collection.lua
```diff
@@ -58,8 +58,8 @@ function M.reduce(tbl, func, ...)
...
 
 function M.find(tbl, func)
```
oberblastmeister, line 60: I don't think we need this module, we can just use `plenary/iterators.lua`. It is more general because it works on iterators and not tables which is probably faster too.

TimUntersberger: This probably belongs in a seperate PR, because this was introduced before this PR.

akinsho:
@TimUntersberger I just tried this out and it's very cool 💪🏾 .

I noticed a few things I thought I'd mention
* The buffers don't have syntax highlighting, seems to be because the newly created buffers don't have filetypes set, adding `setfiletype lua` etc. fixes the issue.
* The diff can only be opened once. Once you close them and hit `D` again nothing happens.
* I think maybe the windows should be `vim.bo.bufhide = "wipe"` and `vim.bo.buflisted = false` so they don't hang around when you are done diffing things.

TimUntersberger:
> The diff can only be opened once. Once you close them and hit D again nothing happens.

This doesn't happen for me. Are you encountering an error?

akinsho:
@TimUntersberger I'm not seeing the issue with it not reopening anymore.

I just tested out a bit more and it's related to if a diff is already open. So if I close the diff with `q` and try another that works fine but if I trigger neogit again whilst the diff is open I get taken back to my original tab (which is fine) and if I then try to open another diff for another hunk it won't work i.e. you can open multiple different diffs

TimUntersberger:
Closing this in favor of #124

PR #101, opened 2021-04-07
refactor: move code over to plenary's async

RianFuro:
TODO:
- [x] fix problem with cursor restore when not in the neogit status buffer


lua/neogit/status.lua
```diff
@@ -316,25 +317,41 @@ function refresh(force)
...
+        future(function ()
+          await(repo:load_diffs())
+          print('reload diff', os.clock() - time)
+          await(scheduler())
+          refresh_status()
+        end)
+      })
       vim.cmd [[do <nomodeline> User NeogitStatusRefreshed]]
     end
+    print('reload', os.clock() - time)
```
RianFuro, line 351: Hey @oberblastmeister, quick question: I noticed an ever so slight delay in the initial buffer draw and tried to improve the time to the initial redraw by staggering the loading, but that didn't seem to have any effect. I assume that neovim doesn't redraw the buffer until the lua coroutine finished unwinding?
Do you have any insight into that or know how to force neovim to update the buffer? If not I think I can deal with that some other way, just thought I'd ask.

TimUntersberger: @RianFuro I also noticed a slight delay when opening the status, but sometimes it is instant for me (Not sure why).

RianFuro: The answer to one of my questions is `:redraw` :smile:

oberblastmeister: `:redraw` seems to remove the delay

oberblastmeister: However I don't think we should redraw every time. This can cause some visual glitches especially when spamming neogit keybindings. 
neogit

RianFuro: Yeah, we definitely need to be smarter about how much data we have to reload

RianFuro:
I'm not quite sure what to do about describing the dependency in the README. I just left a general note of the needed dependency, as adding it explicitly to all plugin descriptions really blows up the section.

@TimUntersberger any opinion on this?

lua/neogit/popups/commit.lua
```diff
@@ -3,13 +3,29 @@ local status = require 'neogit.status'
...
-local uv = require('neogit.async.uv')
+local a = require 'plenary.async_lib'
+local async, await, scheduler, void, wrap = a.async, a.await, a.scheduler, a.void, a.wrap
+local uv = a.uv
 local split = require('neogit.lib.util').split
 
 local COMMIT_FILE = '.git/NEOGIT_COMMIT_EDITMSG'
 
-local get_commit_message = a.wrap(function (content, cb)
+local read_file = async(function(path)
```
TimUntersberger, line 13: I think we should move this function into util or a separate file

TimUntersberger (changes requested):

lua/neogit/popups/commit.lua
oberblastmeister: or maybe into plenary

RianFuro: oh I would love that in plenary, but I think our use case is a bit too specific, since I don't want the function to error if the file doesn't exist.

Separate file does sound good though, I'll also wrap that with `git_root` while I'm at it so we can be sure our file is read from the git root when calling (which is what we want 99% of the time :sweat_smile:

oberblastmeister: I don't think that this use case is too specific. We shouldn't be ignoring errors anyway, they could be more than the errors that we are just trying to ignore. We should be dedicating checking if the file doesn't exist to the caller. A better way to check if a file doesn't exist is to just use `uv_fs_stat` before calling the function. We could also just use the plenary path library to check but tj is refactoring it.

RianFuro: > A better way to check if a file doesn't exist is to just use uv_fs_stat before calling the function.

At least in nodejs, this is considered bad practice, since the file can appear/disappear in between the calls (likely not a problem in our case, but still)

See:
https://nodejs.org/api/fs.html#fs_fs_stat_path_options_callback

> Using fs.stat() to check for the existence of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Instead, user code should open/read/write the file directly and handle the error raised if the file is not available.

oberblastmeister: That makes sense

plugin/neogit.vim
```diff
@@ -1,19 +1,19 @@
 lua require("neogit")
 lua require("neogit.status")
 
-function! s:refresh()
+function! s:refresh(file)
   if match(bufname(), "^\\(Neogit.*\\|.git/COMMIT_EDITMSG\\)$") == 0
     return
   endif
-  lua vim.defer_fn(function() require 'neogit.status'.refresh() end, 0)
+  call luaeval('(function() require "neogit.status".dispatch_refresh({ status = true, diffs = {_A}}) end)()', a:file)
```
TimUntersberger, line 8: Is there a reason why you changed to `luaeval`?

RianFuro: Yeah I needed to pass in the filename from viml. Is there a way to do that with `:lua`?

RianFuro: Hmm.... it doesn't work properly anyway, we have the usual problem when `cwd` is not in the repository root, becase the filename doesn't line up with git's representation.

RianFuro:
Alright I think I'm done, implementation-wise at least. I do think we need to do a bit more about advertising the new dependency though. Maybe `pcall` the require early on (somewhere in `neogit.lua` maybe?) and print a warning message on error? I mean it will still hail errors right after I guess, but then at least the information what's actually wrong is in there.

lua/neogit/lib/uv.lua
```diff
@@ -0,0 +1,22 @@
...
+
+  local err, data = await(uv.fs_read(fd, stat.size, 0))
+  if err then return err end
+
+  local err = await(uv.fs_close(fd))
+  if err then return err end
+
+  return nil, data
+end)
+
```
TimUntersberger, line 21: Are you no longer using the read_lines function? It had line splitting that supports windows line endings.

RianFuro: https://github.com/TimUntersberger/neogit/blob/dc6ba8299f1826016b76aa673a111c8630ad08d0/lua/neogit/async/uv.lua#L24-L34
This one? 

That's only in #100 for now :D I can copy it over though.

TimUntersberger: Ah sorry mb.

TimUntersberger (approved):

PR #103, opened 2021-04-07
Remove fold-related code from NeogitStatus.vim

Shatur95:
No longer necessary after #72. See [the answer](https://github.com/TimUntersberger/neogit/pull/72#issuecomment-814642242) from @RianFuro.


ftplugin/NeogitGitCommandHistory.vim
```diff
@@ -1,15 +0,0 @@
...
-
-function! NeogitFoldFunction()
-  return getline(v:foldstart)
-endfunction
-
-setlocal foldmethod=manual
-setlocal foldlevel=1
-setlocal fillchars=fold:\ 
-setlocal foldminlines=0
-setlocal foldtext=NeogitFoldFunction()
```
RianFuro, line 15: The command history still uses folds though, the changes made only apply to the status buffer. So this one needs to stay in for now. Sorry for not stating that more clearly before.

RianFuro (changes requested):

ftplugin/NeogitGitCommandHistory.vim
Shatur95: Got it, thanks!

RianFuro (approved):
lgtm! Thank you for your contribution :)

PR #123, opened 2021-05-13
Add luacheck CI

theHamsta:
Thanks for this nice plugin! I was browsing a bit through the source code to see how it's working and I saw a few warnings.

I thought a style CI would be nice so I largely copied it from nvim-treesitter. The tests are failing at the moment (also on my local machine). This is why I activated `continue-on-error` for this step for now.

The style options are always very opinionated so you should review them on whether you agree on the policies (and changes) or not.

### DIFF
--- a/lua/neogit/lib/git/branch.lua
+++ b/lua/neogit/lib/git/branch.lua
@@ -1,5 +1,5 @@
 local a = require 'plenary.async_lib'
-local async, await, scheduler, void = a.async, a.await, a.scheduler, a.void
+local async, await, scheduler = a.async, a.await, a.scheduler
 local cli = require('neogit.lib.git.cli')
 local util = require('neogit.lib.util')
 local input = require('neogit.lib.input')
--- a/lua/neogit/lib/git/status.lua
+++ b/lua/neogit/lib/git/status.lua
@@ -3,23 +3,9 @@ local git = {
   stash = require("neogit.lib.git.stash")
 }
 local a = require 'plenary.async_lib'
-local async, await, await_all, future = a.async, a.await, a.await_all, a.future
+local async, await, await_all = a.async, a.await, a.await_all
 local util = require("neogit.lib.util")
 local Collection = require('neogit.lib.collection')
-
-local function marker_to_type(m)
-  if m == "M" then
-    return "Modified"
-  elseif m == "A" then
-    return "New file"
-  elseif m == "D" then
-    return "Deleted"
-  elseif m == "U" then
-    return "Conflict"
-  else
-    return "Unknown"
-  end
-end
 
 local update_status = async(function (state)
   local result = await(
@@ -61,7 +47,8 @@ local update_status = async(function (state)
         elseif kind == '!' then
           -- we ignore ignored files for now
         elseif kind == '1' then
-          local mode_staged, mode_unstaged, _, _, _, _, _, _, name = rest:match('(.)(.) (....) (%d+) (%d+) (%d+) (%w+) (%w+) (.+)')
+          local mode_staged, mode_unstaged, _, _, _, _, _, _, name =
+            rest:match('(.)(.) (....) (%d+) (%d+) (%d+) (%w+) (%w+) (.+)')
           if mode_staged ~= '.' then
             table.insert(staged_files, {
               mode = mode_staged,
@@ -79,7 +66,8 @@ local update_status = async(function (state)
             })
           end
         elseif kind == '2' then
-          local mode_staged, mode_unstaged, _, _, _, _, _, _, _, name = rest:match('(.)(.) (....) (%d+) (%d+) (%d+) (%w+) (%w+) (%a%d+) (.+)')
+          local mode_staged, mode_unstaged, _, _, _, _, _, _, _, name =
+            rest:match('(.)(.) (....) (%d+) (%d+) (%d+) (%w+) (%w+) (%a%d+) (.+)')
           local entry = {
             name = name
           }
--- a/lua/neogit/lib/md5.lua
+++ b/lua/neogit/lib/md5.lua
@@ -36,7 +36,7 @@ local bit_or, bit_and, bit_not, bit_xor, bit_rshift, bit_lshift
 
 local ok, bit = pcall(require, 'bit')
 if ok then
-  bit_or, bit_and, bit_not, bit_xor, bit_rshift, bit_lshift = bit.bor, bit.band, bit.bnot, bit.bxor, bit.rshift, bit.lshift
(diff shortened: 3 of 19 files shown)


theHamsta:
Actions run on my fork: https://github.com/theHamsta/neogit/actions

TimUntersberger:
Thank you!

This was long overdue so I really appreciate you opening this PR. I will look at this later today or tomorrow.

lua/neogit/lib/util.lua
```diff
@@ -124,7 +124,7 @@ return {
   filter = filter,
   str_right_pad = str_right_pad,
   str_count = str_count,
-  create_fold = create_fold,
+  --create_fold = create_fold, -- create_fold is not defined
```
theHamsta, line 127: I couldn't find the definition of that function

TimUntersberger:
Have you looked at [selene](https://github.com/Kampfkarren/selene) before? 

Selene looks more convenient to use than luacheck at first glance and it should be faster, because it is written in rust.

theHamsta:
I didn't like selene. Especially in nvim-treesitter it didn't understand and luacheck had far more checks. I was really suprised why some would care for performance for something that runs in a few milliseconds. I mean luacheck is really fast you probably have to check a few 100 files to realize a difference

TimUntersberger:
I'll trust your experience 👍

theHamsta:
I'll add selene later. More additional linting can't harm.

TimUntersberger:
@theHamsta I'd like to avoid having multiple linters. If anything is missing we should improve the linter we are using or switch to something else IMO

lua/neogit/lib/git/branch.lua
```diff
@@ -1,4 +1,5 @@
 local a = require 'plenary.async_lib'
+--luacheck: ignore void
```
TimUntersberger, line 2: why is this needed?

lua/neogit/lib/git/status.lua
```diff
@@ -3,24 +3,10 @@ local git = {
   stash = require("neogit.lib.git.stash")
 }
 local a = require 'plenary.async_lib'
-local async, await, await_all, future = a.async, a.await, a.await_all, a.future
+local async, await, await_all = a.async, a.await, a.await_all
 local util = require("neogit.lib.util")
 local Collection = require('neogit.lib.collection')
 
-local function marker_to_type(m)
```
TimUntersberger, line 10: Why is this getting removed? Can we configure luacheck to tell us about unused functions, but not remove it?

lua/neogit/lib/md5.lua
```diff
@@ -36,7 +36,9 @@ local bit_or, bit_and, bit_not, bit_xor, bit_rshift, bit_lshift
 
 local ok, bit = pcall(require, 'bit')
 if ok then
-  bit_or, bit_and, bit_not, bit_xor, bit_rshift, bit_lshift = bit.bor, bit.band, bit.bnot, bit.bxor, bit.rshift, bit.lshift
+  -- luacheck: ignore bit_not
```
TimUntersberger, line 39: why is this needed?

lua/neogit/lib/util.lua
TimUntersberger: Ah yeah you can remove this.

lua/neogit/popups/commit.lua
```diff
@@ -163,7 +163,7 @@ local function create()
             -- we need \r? to support windows
             data = split(data, '\r?\n')
             await(prompt_commit_message(data, skip_gen))
-            local _, code = await(cli.commit.commit_message_file(COMMIT_FILE).args(unpack(popup.get_arguments())).call())
+            local _, code = await(cli.commit.commit_message_file(COMMIT_FILE).args(unpack(popup.get_arguments())).call()) --luacheck: ignore
```
TimUntersberger, line 166: why are you ignoring this line?

lua/neogit/status.lua
```diff
@@ -5,7 +5,9 @@ local cli = require('neogit.lib.git.cli')
 local notif = require("neogit.lib.notification")
 local config = require("neogit.config")
 local a = require 'plenary.async_lib'
-local async, await, await_all, future, void, scheduler, run = a.async, a.await, a.await_all, a.future, a.void, a.scheduler, a.run
+-- luacheck: ignore run future
```
TimUntersberger, line 8: same thing here

tests/status_buf_spec.lua
```diff
@@ -1,3 +1,5 @@
+-- Ignore line with only whitespace
+-- luacheck: ignore 611
```
TimUntersberger, line 2: I really don't like that we have to do this. I think we should disable this by default.

TimUntersberger (commented):
Looks good overall, just have some questions regarding lint-customizations.

lua/neogit/lib/git/branch.lua
theHamsta: Because "void" is unused. It can also just be removed

lua/neogit/lib/git/status.lua
theHamsta: You can ignore the warning. But better would be to just remove everything unused.

lua/neogit/popups/commit.lua
theHamsta: Topo long. Line breaks look a bit look a bit weird but they are probably the better alternative

tests/status_buf_spec.lua
theHamsta: I can also just remove the few remaining  lines with whites spaces. The warnings are all in the git commit messages

lua/neogit/status.lua
theHamsta: Removed everything unused now. I originally thought that you may had a reason to always import this set of functions in the same manner in all files but they can just be imported when needed.

lua/neogit/lib/git/branch.lua
TimUntersberger: Unused imports can be safely removed imo.

lua/neogit/lib/git/status.lua
TimUntersberger: 👍 to removing the function, I was just a bit concerned that some code might get removed that is accidently unused.

theHamsta: -- luacheck: ignore marker_to_type
would silence the warning

theHamsta:
One thought to selene: I think you can give it a try and migrate to it later if you like it better. But I wouldn't include it into this PR.

TimUntersberger:
Thank you!

PR #131, opened 2021-05-25
commit inspection

TimUntersberger:
This will add support for the following:

* commit view (`enter` on a commit)
* commit summary (`tab` on a commit)

**TODO**

- [x] performance
- [x] ui abstraction
- [x] folds
- [x] commit summary (`tab` on a commit)
- [x] ~refactor status buffer to use the new ui lib and also add commit inspection to commit history there~ (Separate PR)
- [x] refactor command history to use ui lib
- [x] refactor popup lib to use ui lib
- [x] ~maybe rename `row` component to something else?~ (we can think about this later)
- [x] documentation
- [x] commit log parser seems to be not working
- [x] convert popups to builder pattern
  - [x] log
  - [x] pull
  - [x] push
  - [x] help
  - [x] diff
  - [x] stash
  - [x] commit
  - [x] branch


TimUntersberger:
@RianFuro in this PR I want to experiment with extracting a UI library for our buffer needs.

Currently we do everything manually, but somehow having an abstraction over manual toggling, highlighting, signs, etc. would be really nice to have. Do you have any ideas how we could do this? Do we want it to be retained or immediate?

TimUntersberger:
I implemented the thing I said above.

In the below snippet you can see the relevant code so you can get a feel for the library.

```lua
local Ui = require 'neogit.lib.ui'
local text = Ui.text
local col = Ui.col
local row = Ui.row

local map = util.map
local range = util.range

function create_file_component(file)
  return row({
    text(file.path, { highlight = "NeogitFilePath" }),
    text(" | "),
    text(file.changes, { highlight = "Number" }),
    text(" "),
    text(file.insertions, { highlight = "NeogitDiffAdd" }),
    text(file.deletions, { highlight = "NeogitDiffDelete" }),
  })
end

function create_diff_component(diff)
  return col({
    text(""),
    text(diff.kind, " ", diff.file),
    col(map(diff.hunks, function(hunk) 
      return create_hunk_component(diff, hunk)
    end))
  })
end

function create_hunk_component(diff, hunk)
  return col({
    text(diff.lines[hunk.diff_from], { sign = "NeogitHunkHeader" }),
    col(map(range(hunk.diff_from + 1, hunk.diff_to), function(i)
      local l = diff.lines[i]
      local sign
      if diff_add_matcher:match_str(l) then
        sign = 'NeogitDiffAdd'
      elseif diff_delete_matcher:match_str(l) then
        sign = 'NeogitDiffDelete'
      end

      return text(l, { sign = sign })
    end))
  })
end


function M:open()
  if self.is_open then
    return
  end

  self.is_open = true
  self.buffer = Buffer.create {
    name = "NeogitCommitView",
    filetype = "NeogitCommitView",
    kind = "vsplit",
    mappings = {
      n = {
        ["q"] = function()
          self:close()
        end
      }
    },
    initialize = function(buffer)
      local info = self.commit_info
      local overview = self.commit_overview

      self.ui = Ui.new(buffer)

      self.ui:render(
        text("Commit ", info:abbrev(), { sign = "NeogitCommitViewHeader" }),
        text("<remote>/<branch> ", info.oid),
        text("Author:     ", info.author_name, " <", info.author_email, ">"),
        text("AuthorDate: ", info.author_date),
        text("Commit:     ", info.committer_name, " <", info.committer_email, ">"),
        text("CommitDate: ", info.committer_date),
        text(""),
        col(map(info.description, text), { sign = "NeogitCommitViewDescription" }),
        text(""),
        text(overview.summary),
        col(map(overview.files, create_file_component)),
        col(map(info.diffs, create_diff_component))
      )
    end
  }
end
```

TimUntersberger:
We can also easily visualize the node tree for debugging.

image

RianFuro:
@TimUntersberger Hey that looks super neat! What's the difference between a `row` and a `col` though? In the code they seem to do almost the same thing?

A few ideas from my side:
- in the renderer, toss anything that's not text or a table. That way, you can *inline if* parts of the ui, like so:
  ```
  col(
    text("some line"),
    condition and text("only added when condition matches")
  )
  ```
- a margin property for `row` (or `col`?) to more succintly control spacing between sections (like in the status buffer)
- peronal preference, but I would make `text` take a single argument before the options instead of vararg. Since lua allows you to skip paranthesis is certain scenarios, that would make the markup look more pleasing, as long as you don't need the options table:
  ```
  col {
    text "some static text"
    text { "some ", "concatenated text" },
    text({ "full syntax with options" }, { })
  }
  ```
- I'd experiment with having the components produce their content using the iterator pattern. If you can consume the component's content line-wise, a simple column-split layout shouldn't be too complex (which we could then use for transient helper windows like pull/push/commit etc.) Maybe I'll find some time next weekend to check that out.

Honestly, I like the way this looks, reminds me a lot of js-component frameworks :grin:

TimUntersberger:
> What's the difference between a row and a col though?

`row` will squash all of the children into one line.
`col` is more like a grouping component, in that it doesn't change the layout itself but you an apply a highlight group to all of the children.

> in the renderer, toss anything that's not text or a table. That way, you can inline if parts of the ui, like so: 

I don't really understand what you mean.

> a margin property for row (or col?) to more succintly control spacing between sections (like in the status buffer)

👍 

> peronal preference, but I would make text take a single argument before the options instead of vararg. Since lua allows you to skip paranthesis is certain scenarios, that would make the markup look more pleasing, as long as you don't need the options table: 

I am not sure about this one 🤔

> Honestly, I like the way this looks, reminds me a lot of js-component frameworks 😁

That is where I got my inspirations 👍

TimUntersberger:
In the newest version you can also see how we can easily implement folding and section hovering.

Do you have some more feedback regarding the ui library?

TimUntersberger:
@RianFuro 

The below `Commit` function is a component for the LogViewBuffer. This really show how nice it is to work with the ui library. One thing I dislike are how I currently have to set custom options sometimes.

```lua
local function Commit(commit)
  return col {
    row { 
      text(("* "):rep(commit.level + 1), { highlight = "Character" }), 
      text(commit.oid:sub(1, 7), { highlight = "Number" }), 
      text " ", 
      text(commit.description[1]) 
    },
    col({
      row {
        text "Author:     ",
        text(commit.author_name),
        text " <",
        text(commit.author_date),
        text ">"
      },
      row {
        text "AuthorDate: ",
        text(commit.author_date)
      },
      row {
        text "Commit:     ",
        text(commit.committer_name),
        text " <",
        text(commit.committer_date),
        text ">"
      },
      row {
        text "CommitDate: ",
        text(commit.committer_date)
      },
      text " ",
      col(map(commit.description, text), { padding_left = 4 })
    }, { hidden = true, padding_left = (commit.level + 1) * 2 })
  }
end
```

Might experiment with adding metatable functions to a component which would ideally allow the following syntax:

```lua
col.highlight("Number") {
  text.id("hello").highlight("Character") "hello",
  text "world"
}
```

vs

```lua
col({
  text("hello", { id = "hello", highlight = "Character"}),
  text "world"
}, { highlight = "Number" })
```

I have no idea whether it is possible, but this looks nice to me.

RianFuro:
> I have no idea whether it is possible, but this looks nice to me.

That should be possible with a little `__call` magic and a builder pattern. The tricky bit would just be have a kind of "mirrorred" interface on the root of the builder, which creates a new object instead of adding to one:
```lua
text.id('hello').highlight('character')
--  ^           ^-- needs to add to that builder
--  |-- needs to create a new builder
```

Other than that, you can check the git cli builder for reference, it basically does the same thing. 
But yeah, that would look *really* nice :)

RianFuro:
> Do you have some more feedback regarding the ui library?

I think overall it looks great! Only point of criticism I have is how collapsing is handled. I think it's really brittle to basically grab into the UI-tree and flip the flag, especially when you access them with array offsets like here:
```lua
          local stack = self.buffer.ui:get_component_stack_under_cursor()
          local c = stack[#stack]
          c.children[2].options.hidden = true
```
A declarative approach like react/vue does it would be much more robust to changes in the UI layout, but I realize that sounds like a lot of work. Maybe giving a component a "reference name" and pulling it out into a hash table would be a *good enough* compromise?
Something along the lines of:
```
-- in your component
row {
  text 'hi',
  text.ref('my_ref') 'i can be referenced!'
}

-- then somewhere else
local comp = buffer.ui.refs.my_ref
comp.options.hidden = true
```

TimUntersberger:
I think I should clarify that components will only be rendered once and then the layout gets saved. So right now the ui lib is more akin to normal html/js, where the render function represents the markup and everything else is more like DOM access.

> I think overall it looks great! Only point of criticism I have is how collapsing is handled. I think it's really brittle to basically grab into the UI-tree and flip the flag, especially when you access them with array offsets like here

Yeah, I was a bit lazy here. 

We could give the Commit a custom tag and then use this to find the correct one, but for now this seems to be more simpler as the layout shouldn't really change that often, but I 100% agree that this is very unstable and could easily break if we forget about something.

> A declarative approach like react/vue does it would be much more robust to changes in the UI layout, but I realize that sounds like a lot of work

Exporing a declarative approach sounds interesting! 

I do have a few performance concerns with this though, but I guess right now I am completely rerendering the buffer on every update anyway and it works really well if we make sure to change each line only once in an update cycle, so maybe I am just overthinking this.

I will try to implement this and see how it feels!

TimUntersberger:
Just a small heads up: I won't be refactoring the status buffer in this PR, because this takes a lot more time than initially planned and I don't want to slow down this PR.

TimUntersberger:
@RianFuro what do you think about the performance of this right now? Big commits can take around 1-2 seconds too load.

TimUntersberger:
@RianFuro I hope you don't mind me merging this without your review.

RianFuro:
@TimUntersberger Sure :) I've looked at your progress a few times anyway, but I don't really have time at the moment to do a proper review; sorry about not being around much lately (summer is usually pretty busy for me).

TimUntersberger:
> sorry about not being around much lately

no problem! 😆

PR #132, opened 2021-05-25
Add option to disable commit confirmation prompt

ascandella:
Open for discussion: I'm not sure whether this is something that would be useful to other people (and hence belongs in this plugin) or not. I personally find it a bit burdensome (and takes me out of my flow) since I'm used to typing my message, hitting "ZZ" and then moving on.

This is also my first time writing Lua so if there's a better way to format this conditional, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ local neogit = require("neogit")
 neogit.setup {
   disable_signs = false,
   disable_context_highlighting = false,
+  disable_commit_confirmation = false,
   -- customize displayed signs
   signs = {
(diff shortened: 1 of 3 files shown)


TimUntersberger:
That's a good idea! 

Could you add this setting to the `README.md`?

ascandella:
Sure thing. I wasn't sure whether this warranted adding to the readme, but since it seems people are interested, I've added it

TimUntersberger:
Sorry! 

I thought I merged this already.

PR #137, opened 2021-06-03
fix(diffview): Handle files without stats correctly.

sindrets:
Neogit's file struct items don't always have stats attached to them (for instance for deleted files), in which case an error is thrown when trying to open diffview.


TimUntersberger:
Thank you for the bandaid fix!

I guess we should normally still have stats even if the file is deleted right?

sindrets:
I wasn't sure if you needed it in neogit. In the normal diffview I show stats for deleted files as the number of deleted lines.

PR #172, opened 2021-07-17
Interactive push support

TimUntersberger:
This PR will attempt to add interactive push support.

The basics are already down and most of the current problems should be fixable.

Concerns right now:

- [x] performance
- [x] vim won't update the screen until you interact with vim after a push
- [x] ssh passphrase support when pulling

@RianFuro we have to use my custom made `push_interactive` function instead of our process library, because on windows interactive commands are not supported by default. The `jobstart` native vim function supports pty so I have to use it here.

Also regarding the update problem - I think the problem has to do with how I am using async, so if you could look over this I'd really appreciate it.


RianFuro:
> on windows interactive commands are not supported by default
Can you elaborate on that? I don't use windows except for gaming, but I'd like to know how it differs.

> Also regarding the update problem - I think the problem has to do with how I am using async, so if you could look over this I'd really appreciate it.
I'll have a look soon~ish. Besides just a usage error regarding async, my bet is on `jobstart` just interacting poorly with the event loop. You can try not wrapping and awaiting your custom `push_interactive` function and instead doing it the classic style - just passing the callback.

TimUntersberger:
> Can you elaborate on that? I don't use windows except for gaming, but I'd like to know how it differs.

I think magit describes this problem really good [here](https://github.com/magit/magit/wiki/Pushing-with-Magit-from-Windows).
Neovim can simulate pty using jobstart by using the terminal feature.

> You can try not wrapping and awaiting your custom push_interactive function and instead doing it the classic style - just passing the callback.

Will try that!

TimUntersberger:
@RianFuro I think the problem is actually that the `on_exit` callback waits for the event loop, but the event loop waits for the callback (which gets called in the `on_exit` function), but honestly I have no idea and this is more an educated guess than anything.

tjdevries:
I can take a look a bit more later, but I would recommend that you actually use plenary's newer async module (async_lib will be deprecated at some point soon, we are working on documenting and more testing of new async).

New async module will just be `require('plenary.async')` and it removes the requirement to call `await` on functions (as it uses lua's coroutines to functions that yield and resume execution without explicit keywords).

Ping me if you have any specific questions about it though :)

TimUntersberger:
@tjdevries so you'd recommend switching to the new async lib before trying to fix this?

RianFuro:
> vim won't update the screen until you interact with vim after a push

My first problem is that I can't reproduce the problem :sweat_smile: anything specific about the setup? I just created a bare repo and locally cloned that to get an origin i can push to. UI updates properly without interaction after a `Pp`

TimUntersberger:
> My first problem is that I can't reproduce the problem

You are using linux right? It must be a windows problem again 😆

tjdevries:
> @tjdevries so you'd recommend switching to the new async lib before trying to fix this?

You don't necessarily have to do it before this, but you might find it easier to work with and reason about. I'd recommend switching at some point.

> Edit:
> 
> One thing that is a bit annoying with having to use `pty` and `jobstart`, is that the `stdin` and `stdout` just get squashed together into the `stdout`. This makes saving the stdout really hard and cumbersome. Is there another solution or a way to easily see only the stdout?

Hmm, can you give an example of what you mean? I'm not really sure what you're trying to accomplish, I just dropped by because of the ping :)

TimUntersberger:
> Hmm, can you give an example of what you mean?

stdout
```
Username for 'https://github.com'::\Windows\system32\cmd.EXE
T
imUntersberger
Password for 'https://TimUntersberger@github.com': 

r
 
remote: Invalid username or password.
fatal: Authentication failed for 'https://github.com/TimUntersberger/yolked.git/
'
```

The `T\nimUntersberger` part is stdin for example.

Ideally I wouldn't have to manually filter out the stdin.

TimUntersberger:
@RianFuro are you using an ssh key or http based authentication? If the former then please try with http based authentication.

TimUntersberger:
It looks like vim only won't update if you have to input your credentials. Using an ssh key without passphrase fixes the problem.

RianFuro:
> are you using an ssh key or http based authentication? If the former then please try with http based authentication.
> It looks like vim only won't update if you have to input your credentials. Using an ssh key without passphrase fixes the problem.

I just cloned a local repository for testing (`git clone ../upstream`), so I had none of that :sweat_smile: . You can always force a redraw with ` vim.cmd('redraw')`, make sure you're not in a fast-loop though (`await(scheduler())` with the current async).

TimUntersberger:
I can only reproduce this when the following things happen:

* You have to enter your password/username
* The command succeeds

If you enter your password/username and the command fails it responds instantly, but if the command succeeds it takes around 4 seconds until the job exits and the callback is called.

tjdevries:
It's possible the job is waiting for the handle for stdin to be closed or something like that -- it's quite hard to say for me without spending a lot of time digging in haha

PR #173, opened 2021-07-23
feat: add support for pushing to ssh:// remotes

ascandella:
I'm not totally clear on why we need to parse out the username/rest (rather than just using the raw remote URL for the push operation), but this seems to work and follows the existing pattern.

I've tested this on my personal git server, which has a remote of the form `ssh://git@<host>:<port>/<repo>`

### DIFF
--- a/lua/neogit/popups/push.lua
+++ b/lua/neogit/popups/push.lua
@@ -54,7 +54,9 @@ local get_remote_url = async(function(remote)
         end
       end
     elseif vim.startswith(raw_url, "ssh") then
-      error("TODO: ssh protocol")
+      local url = {}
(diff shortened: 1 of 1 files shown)


TimUntersberger:
> I'm not totally clear on why we need to parse out the username/rest (rather than just using the raw remote URL for the push operation), but this seems to work and follows the existing pattern.

This was intended as a dynamic solution for password prompts, but I found a way to support interactive prompts in #172.

My PR should make your PR obsolete, but I can merge this one once it LGTM if you need support right now.

ascandella:
Ah ok cool, let's just close this then :)

PR #194, opened 2021-08-24
Use vim.notify

### LINKED ISSUE #193: Use vim.notify to display messages
At the moment, a library built into the plugin is used for notifications.

I suggesting to use `vim.notify` instead. This will allow users to display messages as they want. For example, using [a dedicated plugin](https://github.com/rcarriga/nvim-notify). Also, this change will simplify the code a lot by delegating all notification stuff to user.

Other plugins use the same approach. [gitsigns](https://github.com/lewis6991/gitsigns.nvim/commit/4f0bf0d931da1d2b55f64a87a9fe11e39681194d), for example.

Let me know if you like the suggestion, I can send a PR for it.

Shatur:
I replaced all notifications with `vim.notify`. All messages will be displayed in echo area by default, but users can simply override this function to have nice notifications as they want or use a dedicated plugin, such as https://github.com/rcarriga/nvim-notify

In theory you can also use `vim.log.levels.TRACE` for debugging.

Closes #193

### DIFF
--- a/lua/neogit/config.lua
+++ b/lua/neogit/config.lua
@@ -4,6 +4,7 @@ M.values = {
   disable_context_highlighting = false,
   disable_signs = false,
   disable_commit_confirmation = false,
+  disable_builtin_notifications = false,
   auto_refresh = true,
   commit_popup = {
       kind = "split",
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -313,11 +313,7 @@ local function handle_new_cmd(job, popup, hidden_text)
 
   if popup and job.code ~= 0 then
     vim.schedule(function ()
-      notif.create({
-        "Git Error (" .. job.code .. ")!",
-        "",
-        "Press $ to see the git command history."
-      }, { type = "error" })
+      notif.create("Git Error (" .. job.code .. "), press $ to see the git command history", vim.log.levels.ERROR)
     end)
   end
 end
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ neogit.setup {
   disable_context_highlighting = false,
   disable_commit_confirmation = false,
   auto_refresh = true,
+  disable_builtin_notifications = false,
   commit_popup = {
       kind = "split",
   },
--- a/lua/neogit/buffers/commit_view/init.lua
+++ b/lua/neogit/buffers/commit_view/init.lua
@@ -33,7 +33,7 @@ function M.new(commit_id, notify)
     buffer = nil
   }
 
-  if notify then
+  if notification then
     notification:delete()
   end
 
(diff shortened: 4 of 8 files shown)


akinsho:
I really don't think that full on removing all of neogit's notifications is fair on all users who aren't all using the new floating window notification plugin, especially given that the plugin in question is billed as a proof of concept and there is really only one available short of a user rolling their own solution. I think that neogit should continue to accommodate users that don't have/use other notification plugins at least till that because the minority/uncommon.

Shatur:
We could make old notification code compatible with `vim.notify` to allow users to use previous notification system in the same way. Or wrap `vim.notify` into old notification code with the option to use it instead.

I just thought that maintain notifications system for git plugin is kinda out of scope, so I asked the author. But we could keep compatibility, sure.

> the plugin in question is billed as a proof of concept

But it much more nicer that the current notifications.

lua/neogit/config.lua
```diff
@@ -10,6 +10,10 @@ M.values = {
     item = { ">", "v" },
     section = { ">", "v" }
   },
+  notify_opts = {
```
TimUntersberger, line 13: I am not really a fan of having this in the configuration.

lua/neogit/lib/git/cli.lua
```diff
@@ -313,11 +312,7 @@ local function handle_new_cmd(job, popup, hidden_text)
 
   if popup and job.code ~= 0 then
     vim.schedule(function ()
-      notif.create({
-        "Git Error (" .. job.code .. ")!",
-        "",
-        "Press $ to see the git command history."
-      }, { type = "error" })
+      vim.notify("Git Error (" .. job.code .. "), press $ to see the git command history", vim.log.levels.ERROR, require("neogit.config").values.notify_opts)
```
TimUntersberger, line 315: Maybe you can just keep all of the `notif.*` calls and rewrite those instead? Would make using vim.notify a bit easier.

TimUntersberger (commented):
> Or wrap vim.notify into old notification code with the option to use it instead.

That is a really good idea 👍 

> I just thought that maintain notifications system for git plugin is kinda out of scope, so I asked the author

I agree.

lua/neogit/config.lua
Shatur: I thought users would like to override icon, for example. But I can hardcode it, yes.

lua/neogit/lib/git/cli.lua
Shatur: Okay, will try.

Shatur: @TimUntersberger ping :)

TimUntersberger: Sorry! The object is mainly used to destroy the notification early on. You might have to remove code related to that.

Shatur: Okay!

Shatur:
@TimUntersberger Done! I left the old notifications, but changed their API a bit to make it look more like `vim.notify` and added an option to disable built-in notifications (disabled by default, so user will have old notifications).

TimUntersberger:
LGTM. Thank you!

Issue #196, opened 2021-08-26
Refresh of status does not work when using split_above


jens1205:
With using kind="split_above", the status is only updated partially when doing a commit, i.e. after writing and closing the commit message, the last commit message is not updated in the status view, but the "staged changes" tree is removed. 

When using :Neogit (with default kind=tab), also the commit message is updated.

Maybe this is related to another problem: When using splits, the cursor does not jump back to the status window, but instead to some other window with an currently edited file.
A manual refresh after jumping to the status window manually does work. The debug log shows some problem acquiring the refresh lock - this does not occur when using kind=tab.

[neogit] [DEBUG 21:36:07] ...nvim/site/pack/packer/start/neogit/lua/neogit/status.lua:365: [STATUS BUFFER]: Refresh lock is now free                                                     
".git/NEOGIT_COMMIT_EDITMSG" [Neu] 5L, 189C geschrieben                                                                                                                                 
[neogit] [DEBUG 21:36:19] ...site/pack/packer/start/neogit/lua/neogit/lib/git/cli.lua:543: [CLI]: Executing 'ls-files --others --cached --modified --deleted --full-name
 .git/NEOGIT_COMMIT_EDITMSG'                                                                                                                                            
[neogit] [DEBUG 21:36:19] ...site/pack/packer/start/neogit/lua/neogit/lib/git/cli.lua:307: Execution of 'git ls-files --others --cached --modified --deleted --full-name
 .git/NEOGIT_COMMIT_EDITMSG'                                                                                                                                            
[neogit] [DEBUG 21:36:19] ...site/pack/packer/start/neogit/lua/neogit/lib/git/cli.lua:311:   took 0 ms                                                                                      
[neogit] [DEBUG 21:36:19] ...site/pack/packer/start/neogit/lua/neogit/lib/git/cli.lua:543: [CLI]: Executing 'ls-files --others --cached --modified --deleted --full-name
 cmd/one/main.go'                                                                                                                                                       
[neogit] [DEBUG 21:36:19] ...site/pack/packer/start/neogit/lua/neogit/lib/git/cli.lua:307: Execution of 'git ls-files --others --cached --modified --deleted --full-name
 cmd/one/main.go'                                                                                                                                                       
[neogit] [DEBUG 21:36:19] ...site/pack/packer/start/neogit/lua/neogit/lib/git/cli.lua:311:   took 0 ms                                                                  
[...]

PR #197, opened 2021-08-27
Config value commit_popup.kind to allow overwrite of default behaviour

jens1205:
commit_popup.kind defaults to "split" (which is also the default before this PR). Having this config value allows to overwrite it e.g. with "split_above", which is handy in case you open neogit also with split_above (see #196)


TimUntersberger:
Thank you!

PR #199, opened 2021-08-31
chore: Update diffview integration.

sindrets:
Making some [big changes again](https://github.com/sindrets/diffview.nvim/pull/64) over in diffview.

Really sorry about making two breaking changes in less than a month. But after this there won't be any breaking changes in the foreseeable future. Due to the growing scope of the plugin I saw an immediate need to create some more descriptive namespaces.

I'll wait at least until tomorrow to merge the PR on my end.


TimUntersberger:
Thank you!

PR #217, opened 2021-09-18
Add support for staging/unstaging on section title

JCapucho:
Magit allows the user perform actions on sections titles which in turn affects all of the section contents. This PR adds support for doing staging/unstaging.

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -573,25 +573,40 @@ end
 local stage = function()
   M.current_operation = "stage"
   local section, item = get_current_section_item()
+  local mode = vim.api.nvim_get_mode()
 
   if section == nil
     or (section.name ~= "unstaged" and section.name ~= "untracked" and section.name ~= "unmerged")
-    or item == nil then
+    or (mode.mode == "V" and item == nil) then
     return
   end
-
-  local mode = vim.api.nvim_get_mode()
 
   if mode.mode == "V" then
     stage_selection()
   else
     local on_hunk = current_line_is_hunk()
-    if on_hunk and section.name ~= "untracked" then
-      local hunk = get_current_hunk_of_item(item)
-      local patch = generate_patch_from_selection(item, hunk)
-      cli.apply.cached.with_patch(patch).call()
+    if item == nil then
+      if section.name == "unstaged" then
+        git.status.stage_modified()
+      elseif section.name == "untracked" then
+        local add = git.cli.add;
+        for i,_ in ipairs(section.files) do
+          local item = section.files[i];
+          add.files(item.name)
+        end
+        add.call()
+      end
+      refresh(true)
+      M.current_operation = nil
+      return
     else
-      git.status.stage(item.name)
+      if on_hunk and section.name ~= "untracked" then
+          local hunk = get_current_hunk_of_item(item)
+          local patch = generate_patch_from_selection(item, hunk)
+          cli.apply.cached.with_patch(patch).call()
+        else
+          git.status.stage(item.name)
+        end
     end
   end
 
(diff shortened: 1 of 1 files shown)


lua/neogit/status.lua
```diff
@@ -573,26 +573,41 @@ end
...
+    if item == nil then
+	  if section.name == "unstaged" then
+	    git.status.stage_modified()
+--	  elseif section.name == "untracked" then
+--		local files = {};
+--	    for i,_ in ipairs(section.files) do
+--		  local item = section.files[i];
+--		  table.insert(git.cli.add.files, item.name);
+--	    end
+--		git.cli.add.files(files).call()
```
JCapucho, line 597: I had trouble getting this part to work, if anyone could help I would be grateful otherwise I can remove it from this PR.

TimUntersberger:
This PR should add support for staging/unstaging sections AND visual selections right?

JCapucho:
I'm not quite sure I understand what you mean, visual selections from what I can see are already implemented.

lua/neogit/status.lua
```diff
@@ -573,26 +573,41 @@ end
 local stage = function()
   M.current_operation = "stage"
   local section, item = get_current_section_item()
+  local mode = vim.api.nvim_get_mode()
 
   if section == nil
     or (section.name ~= "unstaged" and section.name ~= "untracked" and section.name ~= "unmerged")
-    or item == nil then
+	or (mode.mode == "V" and item == nil) then
```
TimUntersberger, line 580: Why did you add this condition?

lua/neogit/status.lua
TimUntersberger: What exactly is the problem?

lua/neogit/status.lua
```diff
@@ -573,26 +573,41 @@ end
...
     local on_hunk = current_line_is_hunk()
-    if on_hunk and section.name ~= "untracked" then
-      local hunk = get_current_hunk_of_item(item)
-      local patch = generate_patch_from_selection(item, hunk)
-      cli.apply.cached.with_patch(patch).call()
-    else
-      git.status.stage(item.name)
-    end
+    if item == nil then
+	  if section.name == "unstaged" then
```
TimUntersberger, line 589: Can you please try to fix the formatting?

TimUntersberger (changes requested):

lua/neogit/status.lua
JCapucho: git doesn't provide a way stage all untracked files so it must be done with a `git add` specifying all untracked files. Problem is that I can't find a way to pass the cli a list of files without it making them all the same argument.

lua/neogit/status.lua
JCapucho: Because `stage_selection`, calls `generate_patch_from_selection` which expects `item` to not be `null`.

lua/neogit/status.lua
```diff
@@ -573,25 +573,40 @@ end
...
   else
     local on_hunk = current_line_is_hunk()
-    if on_hunk and section.name ~= "untracked" then
-      local hunk = get_current_hunk_of_item(item)
-      local patch = generate_patch_from_selection(item, hunk)
-      cli.apply.cached.with_patch(patch).call()
+    if item == nil then
+      if section.name == "unstaged" then
+        git.status.stage_modified()
+--    elseif section.name == "untracked" then
```
TimUntersberger, line 591: Here you have to through each item in the untracked section and call `git.status.stage(item.name)` on it.

JCapucho: Ok so there's no way to bulk stage? I'm asking because what magit does is call only once git stage with all the untracked files.

TimUntersberger: I guess you could build your own git command using our [CLI Api](https://github.com/TimUntersberger/neogit/blob/master/lua/neogit/lib/git/cli.lua).

Example: ```cli.add.files(...).call()``` ( or something like that )

JCapucho: I managed to figure it out, sorry for wasting your time but I'm a lua noob especially with all the metatable magic that the CLI uses.

TimUntersberger:
LGTM. Thank you for your work!

PR #218, opened 2021-09-24
Add Insert on commit configuration

Jimmyscene:
Hey! 

First off, love the plugin, appreciate all the work you've done.

One thing I missed from Magit was, once the commit popup appeared, you could begin typing immediately instead of switching into insert mode and then typing.

I added a configuration option to allow for that behavior, while keeping the default. 

The naming is kind of awkward since I was trying to stick with the other vars you had but let me know if you'd like me to switch to something like "enable_insert_on_commit" instead of the negative.

Thanks again!


TimUntersberger:
LGTM. Thank you!

PR #225, opened 2021-10-07
feat(status): add hint

wongjiahau:
## Preface
As a beginner I feel that it would be convenient to have the common keymap displayed all the time, otherwise I have to keep looking at the help file until I can memorize them.

## Screenshots

| Before | After |
 |--|--|
|image|image|


bronzehedwick:
I think this is a great idea, but could there also be a preference to disable it?

PR #229, opened 2021-10-20
feat(status/discard): improve prompt copywriting

wongjiahau:
## Preface
Currently, the prompt for discarding changes is too generic and thus not very helpful for user to confirm that they've chose the correct file to discard.

## Screenshot
|Before|After|
|--|--|
|image|image|


TimUntersberger:
LGTM! Thank you!

PR #236, opened 2021-10-26
feat: Dynamically generate highlight groups based on color scheme.

sindrets:
Currently the highlights in the status buffer can look very out of
place when you're using a color scheme that doesn't have highlights
tailored for Neogit. Especially so for light color schemes. This PR
implements a relatively capable `Color` class that provides a simple
interface to convert colors between some common color models, as well
as some convenient utility methods for shading and blending colors.
Using this we can generate some much more fitting highlights based on
the colors from the active color scheme.

Colors will only be generated if they're not already defined in the
active color scheme. This way color schemes can still provide a more
tailored experience.

### Examples

> Note: some of these color schemes ***do*** define highlights for Neogit,
but I've forcibly applied the generated colors for the sake of
demonstration.

<details>
<summary>Click to expand</summary>

2021-10-26-221635_maim
<sup>rose-pine moon</sup>

2021-10-26-221431_maim
<sup>rose-pine dawn</sup>

2021-10-26-221145_maim
<sup>everforest dark</sup>

2021-10-26-221242_maim
<sup>everforest light</sup>

2021-10-26-220756_maim
<sup>onedarkpro.nvim</sup>

</details>

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -1,6 +1,7 @@
 local config = require("neogit.config")
 local lib = require("neogit.lib")
 local signs = require("neogit.lib.signs")
+local hl = require("neogit.lib.hl")
 local status = require("neogit.status")
 
 local neogit = {
@@ -47,6 +48,7 @@ local neogit = {
     if not config.values.disable_signs then
       signs.setup()
     end
+    hl.setup()
   end
 }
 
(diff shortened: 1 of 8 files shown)


mizlan:
@sindrets Using the rose-pine light theme right now, it doesn't seem like I'm getting the highlights you show in your screenshot... Has something changed since then?

mizlan:
Ah, they were added, but they don't look as good as the dynamically-generated one, it seems 😅

PR #239, opened 2021-10-29
feat(push): get tracking upstream before push

Avimitin:
The behavior for the push to upstream action is not what I expected. When I press u it actually runs the command "git push upstream branch". But what I expect is it can push to user-defined upstream. Which is set by the "git push -u" command.

So here I use the command "git branch -vv" to get the tracking remote information. The output of the tracking upstream will have a "\*" prefix. The upstream information is wrapped in the "[]" signs. For example, in an output like this: "* master 3d9279e [github/master] docs: update title". The "\*" mark this line as the tracking upstream. And the upstream remote is GitHub, the branch is master.

I add a new flag "-vv" in lua/neogit/lib/git/cli.configuration.branch. And call the cli.branch.tracking.call() function to get the branch text. Besides I implement the get_upstream() function in lua/neogit/lib/git/branch, it matches the string and returns a table with the remote name and tracking upstream name. It will return nil when no match is found. If the action detected nil return, it will call the logger to print an error and return.

### DIFF
--- a/lua/neogit/popups/push.lua
+++ b/lua/neogit/popups/push.lua
@@ -38,7 +38,10 @@ function M.create()
     :action("u", "Push to upstream", function(popup)
       local upstream = git.branch.get_upstream()
       a.util.scheduler()
-      if upstream == nil then logger.error("No upstream set") return end
+      if upstream == nil then
+        logger.error("No upstream set")
+        return
+      end
 
       push_to(popup,
         upstream.remote.." "..upstream.branch,


TimUntersberger:
> Do I need to modify the README for my PR? Maybe I need to introduce the new behavior.

AFAIK you just fixed the current behaviour, right? The way we did it before was incorrect.

lua/neogit/lib/git/branch.lua
```diff
@@ -32,6 +32,19 @@ local function get_all_branches()
   return parse_branches(branches)
 end
 
+local function get_upstream()
+  local upstream_text = cli.branch.tracking.call()
+
+  for _, branch in ipairs(upstream_text) do
+    local ups_remote, ups_branch = branch:match([[^%*.+%[(.+)/([^:]+):?.*%].+]])
```
sindrets, line 39: This pattern seems to make a lot of wrong assumptions about the formatting of both remotes and branches. Both remotes and branches may contain slashes (`/`). Take for instance this example where the remote is called `foo/bar/baz`, and the branch is called `lorem/ipsum/dolor`. Using the current approach, this will be parsed as:

```lua
{
  remote = "foo/bar/baz/lorem/ipsum",
  branch = "dolor"
}
```

Avimitin: Oh, I forgot those edge case.

PR #243, opened 2021-10-29
fix: Parsing of upstream remote + branch.

sindrets:
As I mentioned in #239, the current approach makes some false assumptions about valid formatting for remote- and branch names. A regex won't be enough to separate these when they're presented together because there's no unique separator between them. I think the most straight forward approach is to just read the branch's remote from the git config.


TimUntersberger:
LGTM! Also thank you for cleaning up the `lib/git/branch.lua` file! 😄

PR #246, opened 2021-11-07
fix: Account for partial lines in job output.

sindrets:
The last line in a job channel callback may be a partial line. Until now this has just been ignored. Seemingly, this works fine most of the time, but I had a commit view consistently fail to parse `git-show` output because several characters were dropped from the beginning of a line.

From `:h channel-callback`:

```vimhelp
      {data}	    Raw data (|readfile()|-style list of strings) read from
		    the channel. EOF is a single-item list: `['']`. First and
		    last items may be partial lines! |channel-lines|
```

### DIFF
--- a/lua/neogit/lib/job.lua
+++ b/lua/neogit/lib/job.lua
@@ -50,6 +50,9 @@ function Job:start()
     task = { 'cmd', '/C', task }
   end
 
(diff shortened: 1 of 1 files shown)


TimUntersberger:
LGTM! Thank you!

PR #250, opened 2021-11-21
Fix git commit amend appending unnecessary template

casprwang:
The issue is when we do `ca`, the pop up git message appends git message template. This contradicts the behavior from `git commit --amend`, which just use the previous git message without appending anything.


TimUntersberger:
LGTM! Thank you!

PR #252, opened 2021-11-30
get_current_hunk_of_item returns nil if the item is not tracked by git

emmanueltouzery:
previously get_current_hunk_of_item would fail if the file was not tracked
by git. The reason was that get_current_hunk_of_item calls
get_hunk_of_item_for_line, which loops over ipairs(item.hunks), which
will fail if item.hunks is nil.

An important consequence of this is that going to a specific file in the
status would fail if the file was not tracked by git (for instance, it's
a new file we're adding to the repo now). The code to go to a file is
calling get_current_hunk_of_item to find out which line to go to in the
file.

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -422,6 +422,9 @@ local function get_hunk_of_item_for_line(item, line)
 end
 
 local function get_current_hunk_of_item(item)
+  if item.hunks == nil then
+    return nil
+  end
   return get_hunk_of_item_for_line(item, vim.fn.line("."))
 end
 


emmanueltouzery:
thinking about it again, an alternative fix could be:

```diff
local function get_hunk_of_item_for_line(item, line)
  local hunk
  local lines = {}
+ if item.hunks == nil then
+   return hunk, lines
+ end
  for _, h in ipairs(item.hunks) do
```

But I haven't tried it. If you prefer that, I can test it and change it to that...

TimUntersberger:
LGTM! Thank you!

Issue #244, opened 2021-11-02
Help popup doesn't update according to user config


TimUntersberger:
The menu that shows when pressing `?` still shows `p` as Pull though, which is a little confusing :)

![Screenshot 2021-11-02 at 20 03 03](https://user-images.githubusercontent.com/379823/139928707-10da1b0c-7146-4d5a-b44b-aec217b19554.png)

_Originally posted by @Linuus in https://github.com/TimUntersberger/neogit/issues/221#issuecomment-958081437_

PR #262, opened 2022-01-07
fix: update help popup according to user config

cheeze2000:
Addresses #244 

Changes of this PR:
- using magit keybindings can be done directly through the `setup` function with `use_magit_keybindings = true`
  - `false` by default
- unbind `p` in magit keybindings, used to be "push" but `P` does the same thing and this is more magit-like
- help popup now shows correct keybindings by inspecting `config.values.use_magit_keybindings`
- documentation changes

I welcome any changes/comments to this PR.

P.S. for backward compatibility, I still keep the `config.use_magit_keybindings()` function. I can remove it if you want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ neogit.setup {
   disable_commit_confirmation = false,
   auto_refresh = true,
   disable_builtin_notifications = false,
+  use_magit_keybindings = false,
   commit_popup = {
       kind = "split",
   },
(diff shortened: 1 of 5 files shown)


TimUntersberger:
Thank you!

> P.S. for backward compatibility, I still keep the config.use_magit_keybindings() function. I can remove it if you want.

Can you remove the function in a `BREAKING: ....` commit? Thanks!

Everything else looks good.

TimUntersberger:
Thank you!

oscarbailey-tc:
Can this issue be closed? #244

PR #271, opened 2022-01-24
feat: allow user defined functions and adhoc commands in key mappings

skebanga:
This change allows user to specify custom functions/vim commands in keybindings

If the specified value in the mapping config exists in `func_map`, that is used (current behaviour)

If it doesn't then if the specified keybinding is a function, that is used directly, otherwise if it is a string it is treated as a `vim.cmd`

For an example use case please see below:

```lua
function M.setup()
  local neogit = require('neogit')

  local function quit()
    if #vim.api.nvim_list_bufs() == 1 then
      vim.cmd("enew")
    else
      vim.cmd("BufDel")  -- functionality from "ojroques/nvim-bufdel"
    end
  end

  neogit.setup({
    kind = "replace",

    mappings = {
      status = {
        ["q"] = quit, -- uses my local quit function
        ["<Esc>"] = quit
      },
    },
  })
end
```


TimUntersberger:
Thank you!

skebanga:
Should be fixed now

Issue #158, opened 2021-07-08
Cursor blinking when changing buffers


gegoune:
I have noticed that when changing buffers with snap or fzf's `:Buffers` cursor blinks quite fast for about a second before calming down and starting to blink at its normal pace. Did you see anything like that?

TimUntersberger:
Do you mean it blinks fast inside the NeogitStatus buffer?

gegoune:
Ah, no, apologies for half decent issue report. It blinks on when switching between any buffers, regular buffers, not Neogit buffer.

gegoune:
I have found out that if buffers I am switching between are for files not in git repository that is not an issue:

![ok](https://user-images.githubusercontent.com/69750637/124981956-d9b45380-e03e-11eb-8b7b-0e6ffc82a4b8.gif)

but as soon as  move those files to git repository (committed them in that example) I get the flicker:

![notok](https://user-images.githubusercontent.com/69750637/124982243-24ce6680-e03f-11eb-8828-931b85f98b60.gif)

TimUntersberger:
Tbh, I can't really reproduce this on my machine.

Does this also occur if you switch buffers without using the plugin?

gegoune:
It also happens with Snap, but I think not with as much flicker. Hard to tell for sure.

As to switching without fzf. I was about to respond saying that it doesn't happen, but it does. It's harder to see in my minimalistic test repo but in my dot files repository it is noticeable when using any of those:

- `:b#`
-  tpope/vim-unimpaired's `[b` and `]b` or `[f` and `]f`

But I must say that it does not feel as bad using above methods.
Can you think of a way to make it more palpable, easier to debug?

gegoune:
I have found the culprit:

```
luatree  User
    NeogitStatusRefreshed
              NvimTreeRefresh
```

This autocommand to refresh ivim-tree's git status. Seems to be firing on but enter too?

TimUntersberger:
> I have found the culprit:

Just to make sure: If you remove this line you no longer encounter the flicker no matter what you do?

gegoune:
That is correct.

TimUntersberger:
Well `NeogitStatusRefreshed` gets emitted whenever we refresh the status and this happens frequently.

Maybe the culprit is more the `NvimTreeRefresh` command?

gegoune:
[...]

Issue #266, opened 2022-01-10
Load changed files lazily


Shatur:
**Is your feature request related to a problem? Please describe.**
When a lot of files are changed in the repository, then when you open Neogit, a lot of `rg` processes are started, which slows down the editor.

**Describe the solution you'd like**
I would suggest to load changes lazily only when user opens the changed file in Neogit window.

CKolkey:
This has been implemented.

PR #273, opened 2022-02-12
fix: short circuit refresh on dirvish

DonnieWest:
Resolves an issue where, if we're in a large repo, dirvish will refresh Neogit aggressively. For me, this causes memory to go out of control because of `git ls-files` being run on a large repo over and over

### DIFF
--- a/plugin/neogit.vim
+++ b/plugin/neogit.vim
@@ -7,6 +7,10 @@ if !luaeval("require 'neogit.bootstrap'")
   finish
 endif
 
+function! neogit#refresh_manually(file)
+  call luaeval('(function() require "neogit.status".refresh_manually(_A) end)()', a:file)
+endfunction
+
 function! s:refresh(file)
   if match(bufname(), "^\\(Neogit.*\\|.git/COMMIT_EDITMSG\\)$") == 0
     return
--- a/README.md
+++ b/README.md
@@ -267,6 +267,19 @@ Neogit uses 'p' for pulling instead of 'F'.
 
 Set `use_magit_keybindings = true` in your call to [`setup`](#configuration) to use magit-style keybindings.
 
+## Refreshing Neogit
+
+If you would like to refresh Neogit manually, you can use `neogit#refresh_manually` in Vimscript or `require('neogit').refresh_manually` in lua. They both require a single file parameter.
+
+This allows you to refresh Neogit on your own custom events
+
+```vim
+augroup DefaultRefreshEvents
+  au!
+  au BufWritePost,BufEnter,FocusGained,ShellCmdPost,VimResume * call <SID>neogit#refresh_manually(expand('<afile>'))
+augroup END
+```
+
 ## Todo
 
 **Note: This file is no longer being updated.**
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -38,6 +38,7 @@ local neogit = {
   end,
   dispatch_reset = status.dispatch_reset,
   refresh = status.refresh,
+  refresh_manually = status.refresh_manually,
   dispatch_refresh = status.dispatch_refresh,
   refresh_viml_compat = status.refresh_viml_compat,
   close = status.close,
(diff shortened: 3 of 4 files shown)


DonnieWest:
Thinking on this more, it might make more sense to create a mechanism to disable refresh conditionally. Or just disable this entirely and leave it to the user

That would also potentially solve #266 since it would allow the user to determine exactly when they'd like it to refresh

I'm happy to submit a PR with whatever solution you think is best

gegoune:
Probably related: https://github.com/TimUntersberger/neogit/issues/158

I am still having that change applied, @TimUntersberger. Maybe that issue needs reopening for visibility?

TimUntersberger:
> Thinking on this more, it might make more sense to create a mechanism to disable refresh conditionally. Or just disable this entirely and leave it to the user

This is already possible.

```lua
neogit.setup {
  auto_refresh = false
}
```

@gegoune I pinned the issue.

gegoune:
Thanks! What does `auto_refresh` do exactly? Couldn't find it in documentation.

DonnieWest:
@gegoune my fix here wouldn't resolve #158 since it exclusively fixes the issue with dirvish and no other plugin

DonnieWest:
@TimUntersberger setting `auto_refresh` doesn't seem to fix my issue, largely because it still tries to fetch the files on the path here: https://github.com/TimUntersberger/neogit/blob/c8a320359cea86834f62225849a75632258a7503/lua/neogit/status.lua#L395-L402 despite not actually using those to refresh Neogit. Presumably, we could move the config check up in the function to solve this

Unfortunately, that doesn't give the user control on when to refresh Neogit, since it would prevent the user from calling `refresh_viml_compat` manually. Opting out of this in a way the user can control would require not setting the autocmd here https://github.com/TimUntersberger/neogit/blob/c8a320359cea86834f62225849a75632258a7503/plugin/neogit.vim#L19 somehow. 

I'd propose:

1. Fix `refresh_viml_compat` so it doesn't fetch git information when `auto_refresh` is set
2. Either:
  * Allow the user to opt out of autocmds with a global variable of some sort. It's ugly but it's also happening outside of lua :/
  * Expose a function that's like `refresh_viml_compat` but without the `auto_refresh` check

TimUntersberger:
@gegoune 

> What does auto_refresh do exactly? Couldn't find it in documentation.

Neogit refreshes its internal state after specific events, which can be expensive depending on the repository size. Disabling `auto_refresh` will make it so you have to manually refresh the status after you open it.

@DonnieWest 

> Presumably, we could move the config check up in the function to solve this

100% think we should do this, I don't really know why the config check is so low in the function.

> Unfortunately, that doesn't give the user control on when to refresh Neogit, since it would prevent the user from calling refresh_viml_compat manually. 

I think this is a none issue, since `refresh_viml_compat` is an internal function of neogit and not provided as a public API.

DonnieWest:
@TimUntersberger would you be fine with me exposing that functionality somehow? Maybe a `manual_refresh` function?

TimUntersberger:
@DonnieWest sure 👍

gegoune:
> @gegoune my fix here wouldn't resolve #158 since it exclusively fixes the issue with dirvish and no other plugin

I was thinking more in line of having more generic fix for this issue but I don't know neogit's internals at all.

plugin/neogit.vim
```diff
@@ -7,10 +7,18 @@ if !luaeval("require 'neogit.bootstrap'")
...
+function! neogit#refresh_manually(file)
+  call luaeval('(function() require "neogit.status".refresh_manually(_A) end)()', a:file)
+endfunction
+
 function! s:refresh(file)
   if match(bufname(), "^\\(Neogit.*\\|.git/COMMIT_EDITMSG\\)$") == 0
     return
   endif
+
+  if match("dirvish", &filetype) == 0
```
TimUntersberger, line 19: I think this is no longer required right?

DonnieWest: Done :)

TimUntersberger:
Could you also please add the `refresh_manually` function to the documentation?

DonnieWest:
@TimUntersberger Sorry, just now getting back to this.

Would you like the `refresh_manually` export and `neogit#refresh_manually` vimscript function to both be documented? Should they each get their own section?

DonnieWest:
@TimUntersberger sorry for the delay, how does that look?

PR #277, opened 2022-03-05
remove default kind "tab" in open function

JSteitz:
The open function still overrides the default kind option from the setup.


TimUntersberger:
Thank you!

PR #285, opened 2022-04-13
refactor(floating): use builtin border system

vigoux:
This greatly simplifies the floating window creation system.

This also relies more on the colorscheme definition for consistency,
and avoids random calls to `vim.wo`.


TimUntersberger:
LGTM. Thank you!

PR #301, opened 2022-06-09
fix(status): Adjustments to 'Close' and 'GoToFile'.

sindrets:
- Avoids the "Already only one window" message that appears when you close a neogit tabpage, and there's only one window in the tabpage.
- Avoids errors when closing a neogit tabpage and it's the last tabpage.
- Avoids errors when closing a neogit tabpage, `'hidden'` is disabled and there are buffers with unsaved changes open in the tabpage.
- Avoids unnecessary writes on GoToFile (only write the current file if `'hidden'` is disabled and it's a normal, writeable buffer).

### DIFF
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -82,6 +82,11 @@ function Buffer:close(force)
   if force == nil then
     force = false
   end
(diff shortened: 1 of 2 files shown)


TimUntersberger:
Thank you!

PR #303, opened 2022-06-16
Add Rebase option to help

doblak:
Rebase option is missing from help popup, added it as a final item of first column


TimUntersberger:
thank you!

PR #313, opened 2022-07-12
status: also display unmerged files

emmanueltouzery:
we currently don't display unmerged files (eg conflicted files) in git status. this fixes it.

### DIFF
--- a/lua/neogit/lib/git/status.lua
+++ b/lua/neogit/lib/git/status.lua
@@ -38,6 +38,13 @@ local function update_status(state)
           table.insert(untracked_files, {
             name = rest,
           })
+        elseif kind == "u" then
+          local mode, _, _, _, _, _, _, _, _, name =
+            rest:match("(..) (....) (%d+) (%d+) (%d+) (%d+) (%w+) (%w+) (%w+) (.+)")
+          table.insert(untracked_files, {
+            mode = mode,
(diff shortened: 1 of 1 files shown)


TimUntersberger:
What happens when you press `tab` on the untracked file? Does it work?

emmanueltouzery:
no, it does nothing. then again the file is conflicted, has conflict markers so... I was more interested in checking that pressing enter does take me to the file, which works.
do you think there's value in tab expanding the changes? i can try adding that maybe.

emmanueltouzery:
I checked a bit, it's relatively expected that `tab` won't do anything, since:
```lua
         if f.diff and not file.folded then
```

and we don't populate `diff` in this case.

For the other types of files, we populate the diff from:
```lua
    staged_files = Collection.new(state.staged.items or {}):key_by('name'),
    unstaged_files = Collection.new(state.unstaged.items or {}):key_by('name'),
```

but i'm not sure where to get the diff from in this case.

emmanueltouzery:
I'm not sure where is the `state` populated. If you want that I try to fill in the diff, maybe let me know where that comes from.

TimUntersberger:
Okay, then I think this should be fine 👍

TimUntersberger:
Thank you :)

Issue #320, opened 2022-07-22
Customizing core.commentChar causes issues


skbolton:
### Description

In my git config I change the comment char to be a `;` (# is a bad default since I type a lot of markdown in my commits).

When I launch the commit popup I can tell that things aren't syntax highlighted correctly

![image](https://user-images.githubusercontent.com/13260877/180347186-a64dc071-c78d-436e-ad85-f3c9c353f3a8.png)

Another is that if I abort the commit and try to not commit it saves my comments as the commit message.

### Neovim version

NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info

### Operating system and version

Arch Linux 5.18 kernel

### Steps to reproduce

1. Make changes to repo
2. Open commit buffer
3. Close buffer without editing the message

### Expected behavior

* My comment char `;` should highlight all the lines as comments
* Not adding a commit message should abort message (I think this is a neogit feature?)

I am trying out Neogit as an alternative to fugitive. Not holding this over anyone's head just commenting that my setup works in Fugitive.

### Actual behavior

Highlighting is incorrect
Closing buffer creates a commit message with comments as the message

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing neogit and dependencies.")
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
```

[...]

PR #321, opened 2022-07-22
WIP - Handle user commentChar

skbolton:
Related to #320


skbolton:
I am going to close this. I am not a daily user of neogit and so I don't think I will be the one to add this feature.

PR #324, opened 2022-07-30
feat: add diffview support for log view

EpiCanard:
Description :
This PR add a mapping `d` in LogView to open diffview.nvim with the diff of the commit under the cursor.


TimUntersberger:
Thank you!

PR #331, opened 2022-08-07
fix(diffview): Use the `^!` notation to describe single-commit ranges.

sindrets:
The `<SHA>^..<SHA>` format is unnecessarily verbose and doesn't work if
`<SHA>` has no parent commit. The `<REV>^!` format more accurately
describes the range of a single commit, and works regardless of the
number of parents.

### DIFF
--- a/lua/neogit/integrations/diffview.lua
+++ b/lua/neogit/integrations/diffview.lua
@@ -113,10 +113,10 @@ function M.open(section_name, item_name)
 
   if section_name == "recent" or section_name == "unmerged" or section_name == "log" then
     local commit_id = item_name:match("[a-f0-9]+")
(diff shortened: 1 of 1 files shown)


TimUntersberger:
Damn nice! Thanks!

EpiCanard:
Oh, I didn't know this way of looking at a specific commit.

I was sure I tested when there is no parent commit but it seems I didn't test well.

Thanks for fixing my mistakes.

sindrets:
@EpiCanard No worries! You can read more about `^!` and other notation for specifying git revisions from `man git-rev-parse` under `SPECIFYING REVISIONS` and `SPECIFYING RANGES`.

PR #337, opened 2022-08-20
Close popups with <esc>

psakalo:
Hi, and thanks everyone for awesome work on this plugin!

I recently started playing with it, and first thing I noticed is that it took me some time to figure out that popup should be closed with `q`. I think having `<esc>` as an option would be more convenient.


TimUntersberger:
Hmm... I think that is a valid default binding. Thanks!

Issue #258, opened 2021-12-09
NeogitLogPopup actions beside Log current are failing


aleprovencio:
### Description

`Log current` is working fine, but `Log HEAD`, `Log all branches` and `Log all references` return error.

### Neovim version

NVIM v0.6.0
Build type: Release
LuaJIT 2.0.5

### Operating system and version

Archlinux

### Steps to reproduce

1. call `:Neogit`
2. hit `L`
3. hit `b` or `a` or `h`

### Expected behavior

normal output

### Actual behavior

E5108: Error executing lua ...ck/packer/start/plenary.nvim/lua/plenary/async/async.lua:14: attempt to call global 'error' (a boolean value)
stack traceback:
        ...ck/packer/start/plenary.nvim/lua/plenary/async/async.lua:14: in function 'callback_or_next'
        ...ck/packer/start/plenary.nvim/lua/plenary/async/async.lua:40: in function 'step'
        ...ck/packer/start/plenary.nvim/lua/plenary/async/async.lua:43: in function 'execute'
        ...ck/packer/start/plenary.nvim/lua/plenary/async/async.lua:113: in function 'callback'
        ...m/site/pack/packer/start/neogit/lua/neogit/lib/popup.lua:214: in function 'cb'
        .../site/pack/packer/start/neogit/lua/neogit/lib/buffer.lua:298: in function <.../site/pack/packer/start/neogit/lua/neogit/lib/buffer.
lua:297>
        .../packer/start/neogit/lua/neogit/lib/mappings_manager.lua:9: in function 'invoke'
        [string ":lua"]:1: in main chunk

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
[...]

PR #343, opened 2022-09-12
Log popup issues

levouh:
See #258 for some details on the issue. Not really sure what the solution is, but this seems like it might be a way to do it. Looking for feedback on what the right way is so I can make the changes. Having a proper log viewer would be awesome.

### DIFF
--- a/lua/neogit/lib/popup.lua
+++ b/lua/neogit/lib/popup.lua
@@ -31,6 +31,9 @@ end
 
 function M:get_arguments()
   local flags = {}
+  for _, _switch in pairs(self.state._switches) do
+    table.insert(flags, "--" .. _switch.cli)
+  end
   for _, switch in pairs(self.state.switches) do
     if switch.enabled and switch.parse ~= false then
       table.insert(flags, "--" .. switch.cli)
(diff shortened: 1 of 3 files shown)


levouh:
@TimUntersberger would love some feedback if you have the time.

TimUntersberger:
Hey @levouh,

I just read your comments on the issue and I'm not sure what the purpose of the `_switch` is. Your initial solution is the correct way to handle this imo. 

As @chuck-flowers pointed out, neogit always should be parsing the log as a graph, since it contains a lot of additional information. The graph switch is intended for changing the visual representation.

Also thank you for opening a PR.

levouh:
Yes, so as I interpreted it there are two kinds of "flags" here. Those for `neogit`, and those for `git`.

Some have options, like `--max-files=...` and some are just "togglea" (or switches).

Rather than copy `"--graph"` to every line in this portion of the code (something others then have to remember to do if more things are added), `_switch` serves as an "internal switch" to the `git` flags that `neogit` uses (could be named better for sure).

The part that I wasn't sure about is that seem more like they are for the `popup` than the actual "command generator", however this _seemed_ like the only place to add flag handling like this.

TimUntersberger:
I'd suggest to just extract ```git.cli.log.format("fuller").args("--graph", unpack(popup:get_arguments()))``` into a function.

This is such a simple problem that I don't think adding some complex solution is worth it.

Issue #339, opened 2022-08-29
Provide Visual Feedback if Operation is In Progress


atomdmac:
**Is your feature request related to a problem? Please describe.**
I work with a few repos that have fairly long-running pre-commit hooks.  I love that Neogit doesn't block Neovim while that operation is running but I wish I was able to see the output of the long-running command (or at least some visual indication that there is an operation in progress).

**Describe the solution you'd like**
Ideally, output from the Git command would be piped to a new temporary buffer so it can be observed while the command runs.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

aleprovencio:
That is right @atomdmac, and would like to add that I was having a great experience with `neogit` until I had to deal with a `pre-commit` enabled repos, which then turned into something far from bearable.

I've also seen #329, if you all agree we can make another issue focusing only on `pre-commit` integration.

coldis:
In lazygit, the pre-commit hook is run before giving the user the opportunity to enter the commit message. Personally I liked this flow as there was no risk of writing a commit message that was lost because the commit failed.

ten3roberts:
There are a few ways to handle the precommit hook.

Depending on the needs, a statusline module could show a simple spinner, or a terminal buffer could be opened which shows the running process.

aleprovencio:
I guess the crux is really integrating into `neogit`'s workflow, just like you're doing for rebasing on https://github.com/TimUntersberger/neogit/pull/346

As for the interface, I think of it as a secondary issue and would like anything that just don't break the workflow inside neogit.

I've seen this [plugin](https://github.com/Ttibsi/pre-commit.nvim), but IMO it still misses a lot of polish.

ten3roberts:
I've experimented with a terminal buffer which opens if a command takes more than 1 seconds to run, such as a commit hook running, so the user can se the output.

This buffer also applies to other commands which may be slow

aleprovencio:
I was not able to issue a `:Neogit` command using your repo, any hints?

ten3roberts:
Unfortunately no, the branch is a bit work in progress, so not everything is working.

I will continue and polish the rough edges these following days

PR #346, opened 2022-09-17
Add rebase continue and abort, reworked commit view, and more

ten3roberts:
This PR also includes a slight rework in the commit view, allowing it to prepopulate the editor with "Merged branch master with origin/master" which happens if you commit during a merge. The previous behavior was an empty commit message.

Additionally, there is now a rebasing section in the status view, showing the rebasing head, as well as the interactive steps remaining.

I have also fixed auto refreshing of the status view when pulling and rebasing, so that it properly shows the merge/rebase conflicts if merging/pulling/rebasing failed due to a conflict.

Let me know what you think


ten3roberts:
image
image

Rebase in progress view, and commit during merge view

lua/neogit/popups/rebase.lua
```diff
@@ -10,6 +10,12 @@ function M.create()
   local p = popup
     .builder()
     :name("NeogitRebasePopup")
+    :action("r", "Continue rebase", function()
```
EpiCanard, line 13: The binding `c` isn't more intuitive ? c for continue.

lua/neogit/popups/rebase.lua
```diff
@@ -10,6 +10,12 @@ function M.create()
...
+      rebase.continue()
+    end)
+    :action("a", "Abort rebase", function()
```
EpiCanard, line 16: Can you add also the option for --skip ?

lua/neogit/popups/rebase.lua
EpiCanard: You can use `action_if` to only display these actions when there is rebase in progress ?

ten3roberts: I used `rr` as that is the binding in vim-fugitive, partly because it has the same double tap as `cc` for commit.

EpiCanard: I never used vim-fugitive so I was just thinking of a binding more intuitive for any users

ten3roberts: Alright, I've added action_if and skip as well

EpiCanard: Thank you for your work !

TimUntersberger:
thank you for your work! I'd suggest keeping the mapping for rebase continue as `r`, since all of the keybindings are ideally a copy of magit (they use r).

ten3roberts:
Right, I will revert that change and use rr, as well as fix the lints.

Do you run selene via lsp in your editor, or only as a CI?

TimUntersberger:
I usually just let the CI handle it.

ten3roberts:
Fixed the unused `get_commit_view` function.

It was marked with a selene allow comment, but it seemed to complain anyway.

ten3roberts:
On another note: while writing this feature I notices the `hl` module was quite messy.

Highlights are defined in two different places, and uses bg color matching to detect if a highlight has been overridden, which doesn't work if you define a highlight by `M.link`, as I initially attempted.

I think a better idea would be to use a simple `highlight default {name} {value}`, which makes it easy for a user to provide their own

ten3roberts:
I discovered that rebase elsewhere didn't refresh, and fixed it.

Other than that, rebasing is now working magically, to the point that I now find myself using tmp commits and actualy squashing them before pushing; something I've never done before.

I would like to contribute more to this project, and if so what is the current issue prioritization?

TimUntersberger:
> Other than that, rebasing is now working magically, to the point that I now find myself using tmp commits and actualy squashing them before pushing; something I've never done before.

That sound awesome!

> I would like to contribute more to this project

Thank you haha

> what is the current issue prioritization?

I haven't really developed anything in a few months, so I don't really have a prioritization. I usually just worked on whatever I needed the most for my own workflow. Right now I'm just overseeing PRs. 

So just work on whatever you need the most yourself. (Though I guess merge conflicts would be a major improvement)

aleprovencio:
Sorry about being a little nosy here!

Since `diffview` recently released it's new "merge tool mode", isn't this "merge conflicts" problem already addressed @TimUntersberger ?

@ten3roberts, I think that a `pre-commit` integration would be really useful, and affects a lot of people. Please, take a look at #339 !

TimUntersberger:
@ten3roberts if you are happy with the PR then I'm going to merge this now. Thanks again!

ten3roberts:
Yes, I am happy with the PR. Looking to pick up other issues such as the pre-commit hook one soon

PR #357, opened 2022-09-29
Mirror magit behaviour for push elsewhere

vheon:
According to magit documentation https://magit.vc/manual/magit/Pushing.html#Pushing the "elsewhere" is to push the current branch to anther remote while "other" is to push a given branch to a given remote which is what we had as "elsewhere" before.


vheon:
@ten3roberts @TimUntersberger is this something that the plugin is not interested in? I currently keep a fork of the plugin only for this since I work with multiple local remotes and the current "elsewhere" option forces me to type also the current branch name.

ten3roberts:
Oh yes, absolutely.

How is magits behaviour for pull, as pulling from.

I also think we should change the pull-elsewhere to also use the current remote but a different branch. Now you are prompted to *type* origin and branch, instead of picking i branch from a popup.

I often use this to pull-rebase from master

vheon:
> How is magits behaviour for pull, as pulling from.
>
> I also think we should change the pull-elsewhere to also use the current remote but a different branch. Now you are prompted to type origin and branch, instead of picking i branch from a popup.

yes magic behaviour for pulling is also different:

The `e` postfix
https://github.com/magit/magit/blob/2a5815fee1ed091751b8bbfa242f7a78d7e5e039/lisp/magit-pull.el#L61

calls this

https://github.com/magit/magit/blob/2a5815fee1ed091751b8bbfa242f7a78d7e5e039/lisp/magit-pull.el#L154-L161

which asks the branch from the minibuffer but uses the "remote" section of the tracking branch

I guess we can tackle that in a separate PR?

ten3roberts:
Yes, it is probably a good idea to tackle it in a separate PR since I've modified and fixed that part in an ongoing PR.

Otherwise, thanks for the work

Issue #356, opened 2022-09-29
Environment issues when committing


itmecho:
### Description

When I commit, I get issues where my pre-commit hook fails to run because it can't find the a required binary. The binary is in my PATH and if I run `lua print(vim.inspect(vim.fn.executable('pnpm')))` then it returns `1` as expected.

I think this is a problem with the environment rather than my pre-commit hook as it was working until I ran `PackerSync` a few days ago.

Also, if I use the `--no-verify` flag, it continues as expected but then when I try to write and quit the commit message, I get another error because it failed to GPG sign the commit. This is because the git command can't access the gpg socket environment variable.

### Neovim version

```
NVIM v0.8.0-dev-1173-gc815aadfc
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /bin/cc -DNVIM_TS_HAS_SET_MATCH_LIMIT -DNVIM_TS_HAS_SET_ALLOCATOR -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/iain/src/neovim/build/cmake.config -I/home/iain/src/neovim/src -I/home/iain/src/neovim/.deps/usr/include -I/usr/include -I/home/iain/src/neovim/build/src/nvim/auto -I/home/iain/src/neovim/build/include
Compiled by iain@orion

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/home/iain/.local/share/nvim"

Run :checkhealth for more info
```

### Operating system and version

```
Void Linux 5.18.19_1 #1 SMP PREEMPT_DYNAMIC Thu Aug 25 14:36:55 UTC 2022 x86_64 GNU/Linux
```

### Steps to reproduce

1. Add a basic script somewhere in your PATH. For me that's `~/.local/bin`.
```sh
[...]

Issue #353, opened 2022-09-28
error when using different GIT_DIR


tastytea:
### Description

When i set `GIT_DIR=/home/tastytea/.dotfiles GIT_WORK_TREE=/home/tastytea` and launch `:Neogit`, it shows an error and the NeogitStatus buffer is empty.

It worked fine until i updated my plugins yesterday. tpope/vim-fugitive and lewis6991/gitsigns.nvim still work, so it doesn't seem to be a problem with my repo.

### Neovim version

```
NVIM v0.7.2
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/x86_64-pc-linux-gnu-gcc -march=native -mtune=native -O2 -pipe -g1 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -DNVIM_TS_HAS_SET_ALLOCATOR -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -Wimplicit-fallthrough -Wsuggest-attribute=pure -Wsuggest-attribute=const -Wsuggest-attribute=malloc -Wsuggest-attribute=cold -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/var/tmp/portage/app-editors/neovim-0.7.2/work/neovim-0.7.2_build/config -I/var/tmp/portage/app-editors/neovim-0.7.2/work/neovim-0.7.2/src -I/usr/include -I/var/tmp/portage/app-editors/neovim-0.7.2/work/neovim-0.7.2_build/src/nvim/auto -I/var/tmp/portage/app-editors/neovim-0.7.2/work/neovim-0.7.2_build/include
Compiled by portage@localhost

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "/etc/vim/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

### Operating system and version

Gentoo Linux

[...]

Issue #347, opened 2022-09-24
Unable to use Neogit after update


Icy-Thought:
### Description

Re-installed neovim yesterday and after installing all the plugins required for my setup I had noticed that neogit was not operational as it once had been. 

That is to say, it would not work with neither the default configurations (`require("neogit").setup()`) nor with my customized neogit setup.

### Neovim version

NVIM v0.8.0-dev
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by nixbld (nix overlay: neovim-nightly)

### Operating system and version

NixOS (`nixos-unstable`)

### Steps to reproduce

1. Launch Neovim.
2. Call neogit through usual commands.
3. Greeted with error.

### Expected behavior

A working neogit setup.

### Actual behavior

```
Error executing luv callback:
.../nvim/site/pack/packer/opt/neogit/lua/neogit/process.lua:57: ENOENT: no such file or directory
stack traceback:
        [C]: in function 'error'
        .../nvim/site/pack/packer/opt/neogit/lua/neogit/process.lua:57: in function 'returned_functi
on'
        ...pack/packer/opt/plenary.nvim/lua/plenary/async/async.lua:31: in function 'callback_or_nex
t'
        ...pack/packer/opt/plenary.nvim/lua/plenary/async/async.lua:45: in function 'cb'
        .../nvim/site/pack/packer/opt/neogit/lua/neogit/process.lua:16: in function 'raise_if_fully_
closed'
        .../nvim/site/pack/packer/opt/neogit/lua/neogit/process.lua:50: in function <.../nvim/site/p
ack/packer/opt/neogit/lua/neogit/process.lua:44>
```

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
[...]

PR #358, opened 2022-09-30
Add a console window to show output of slow git commands

ten3roberts:
This PR adds a verbosity option to processes. Verbose processes will show both stdout and stderr to the console, while non-verbose ones, like `git log` and `git rev-parse` will only show stderr. If a command takes more than `console_timeout` time and an interactive target, such as a commit buffer has not been reached, the console will open, showing what is happening.

This happens for slow commits hooks, or pre-commit hooks which fails, allowing the user to very easily see what ran, what is said, and why it is taking so long.

This is slightly akin to the already existing log, but a more user facing and realtime approach.

This implements the features discussed in #339


ten3roberts:
It feels great to be able to rebase and remove tmp commits before opening a pr. No more messy commit history :tada:

ten3roberts:
Let me know how this feature could be improved, and if there should be a statusline indicator or similar as well

gegoune:
@ten3roberts My commit just failed after writing and saving commit message. pre-commit's output with all 'passes' was present in the terminal with line at the end:

```
  env: python: No such file or directory
```
This is not happening in regular terminal with `git commit` and also wasn't happening on master branch. Can test it more if you point me in the right direction.

Another small point - output wasn't coloured, is it possible to make it look the same as in regular terminal?

ten3roberts:
Yes.

Those are exactly the two painpoints I've been encountering.

Colors are easy, we just need to enable pty.


The python env issue is because `libuv.spawn` doesn't merge the env.

Using the `jobstart` api would allow for both pty behavior, env merging, and sync/blocking calls. Currently, jobstart are used for call_blocking, and interactive_git_command (git push), while `uv.spawn` is used for the rest of the background commands.

The only thing I've found is that jobstart replaces nulls with newlines, which means that the `-z` flag used in git status does not work. Have not been able to find out why jobstart does this.

@TimUntersberger is it ok to replace the process.lua usage of libuv.spawn with jobstart, which would also allow for more code reuse, as the console command can't be used for call_sync/blocking, and the git error logs aren't working for the `to_job` cli commands.

If so, I can include the change to jobstart in this pr

ten3roberts:
The python issue is related to the environment not being merged à'la #353, #356, and #347, which is again fixed by jobstart, which was used previously by `interactive_git_command` rather than the `.call()` which is used for others

TimUntersberger:
Wow this is really cool! Thank you for your work again! Do you want to be added as a collaborator?

> is it ok to replace the process.lua usage of libuv.spawn with jobstart? It would allow for more code reuse, as the console command can't be used for call_sync/blocking, and the git error logs aren't working for the to_job cli commands.

👍

ten3roberts:
Another preference discussion. The console window is currently global for all commits with headers, showing each git command. Non verbose commands which do not output anything to stderr are not shown.

Because it is global you can see the context that lead up to the failing command, such as a in progress rebase followed by a pull.

Should I keep it this way or do one console window per command? The advantage with keeping it global is that a background buffer won't need to be created for each git command in vain in case it may output. The disadvantage is that it could become slightly more cluttered if many commands are marked as verbose.

The console window is a tty terminal, so colors and all should work, we just need to tell the subprocess that is has access to a pty (which will come with jobstart migration).

> Wow this is really cool! Thank you for your work again! Do you want to be added as a collaborator?
> 
> > is it ok to replace the process.lua usage of libuv.spawn with jobstart? It would allow for more code reuse, as the console command can't be used for call_sync/blocking, and the git error logs aren't working for the to_job cli commands.
> 
> +1

It would be great to be added as a collaborator :smile:

ten3roberts:
There is now a keybind `#` to open the console buffer. The log buffer is still useful for viewing the execution order of commands, and what has been done, while the console is mostly useful to view what went wrong, or a failed precommit hook

ten3roberts:
I have now finished the console feature, as well as making some refactors and dead clone cleanups. Most notably, removal of the two different process/job apis, as wel as merging the `interactive_git_command` to also use the same api instead of a custom jobstart invocation, which isn't handled by the logs.

Let me know what you think, and if you encounter any issues

TimUntersberger:
Thank you for your work! I also added you as a collaborator now.

Issue #360, opened 2022-10-01
Errors when untracked files are present in submodules


anachronic:
### Description

If any repo has submodules that have untracked files inside, then Neogit spits one error per submodule that has untracked files. I'll send a PR shortly that suppresses the error, but does nothing more.

Happy to contribute if possible!

### Neovim version

NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui

### Operating system and version

Arch Linux x86_64

### Steps to reproduce

1. get any git repo
2. add a submodule: `git submodule add https://github.com/nvim-lua/plenary.nvim`
3. commit the submodule
4. cd into the submodule: `cd plenary.nvim`
5. create any file: `touch somefile`
6. go into neogit status and see the error

### Expected behavior

Neogit doesn't error, and *maybe* it could show that there are untracked files in submodules like git does? I'm happy with just no errors though

### Actual behavior

Errors with the following

```
Error executing luv callback:
...pack/bundle/opt/plenary.nvim/lua/plenary/async/async.lua:18: The coroutine failed with this message: .../nvim/pack/bundle/opt/neogit/lua/neogit/lib/git/diff.lua:63: attempt to index a nil value
stack traceback:
        [C]: in function 'error'
        ...pack/bundle/opt/plenary.nvim/lua/plenary/async/async.lua:18: in function 'callback_or_next'
        ...pack/bundle/opt/plenary.nvim/lua/plenary/async/async.lua:45: in function 'cb'
        ...onfig/nvim/pack/bundle/opt/neogit/lua/neogit/process.lua:16: in function 'raise_if_fully_closed'
        ...onfig/nvim/pack/bundle/opt/neogit/lua/neogit/process.lua:51: in function <...onfig/nvim/pack/bundle/opt/neogit/lua/neogit/process.lua:45>
```

👆 that's once per submodule with untracked files

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
[...]

PR #361, opened 2022-10-01
fix: check diff header length before parsing

anachronic:
Was causing #360. This should at least get rid of the error.


TimUntersberger:
Thank you!

PR #362, opened 2022-10-01
chore: update stylua-action version

anachronic:
1.0.0 is broken and is causing the pipeline to fail.

This PR fixes back the pipeline.


TimUntersberger:
thank you!

PR #363, opened 2022-10-01
Pass SSH agent related env vars to client processes

xanderio:
Without this patch git is unable to contact the ssh agent for commit singing with an ssh key.

### DIFF
--- a/lua/neogit/process.lua
+++ b/lua/neogit/process.lua
@@ -36,6 +36,8 @@ local function spawn(options, cb)
     table.insert(params.env, string.format("%s=%s", "GNUPGHOME", os.getenv("GNUPGHOME") or ""))
     table.insert(params.env, string.format("%s=%s", "NVIM", vim.v.servername))
     table.insert(params.env, string.format("%s=%s", "PATH", os.getenv("PATH")))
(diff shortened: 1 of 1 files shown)


lua/neogit/process.lua
```diff
@@ -36,6 +36,8 @@ local function spawn(options, cb)
     table.insert(params.env, string.format("%s=%s", "GNUPGHOME", os.getenv("GNUPGHOME") or ""))
     table.insert(params.env, string.format("%s=%s", "NVIM", vim.v.servername))
     table.insert(params.env, string.format("%s=%s", "PATH", os.getenv("PATH")))
+    table.insert(params.env, string.format("%s=%s", "SSH_AUTH_SOCK", os.getenv("SSH_AUTH_SOCK" or "")))
```
TimUntersberger, line 39: shouldn't this be `os.getenv("SSH_AUTH_SOCK") or ""`? same on the other line

xanderio: Yes that is a mistake on my part, I'll amend the commit with this fixed.

TimUntersberger:
Thank you!

PR #364, opened 2022-10-01
feat: preserve cursor text position when jumping to file from hunk

Weypare:
When jumping to file from hunk, cursor will be set to the same text position as it was on the hunk in the status buffer.


TimUntersberger:
Thank you!

emmanueltouzery:
i wanted this feature for a long time, thanks a lot @Weypare!!

PR #365, opened 2022-10-02
chore: use uniform vim.cmd formatting

tobealive:
The double quote pattern is already dominant in the repo. This commit makes it uniform.
Keeping the multi brackets would come in handy for complex multiline commands or those in need of escape characters. 
But neither is the case here.


TimUntersberger:
thank you!

PR #366, opened 2022-10-02
fix: commits get sent despite denying prompt

tobealive:
This PR deals with the bug that commits get sent no matter what a user is answering to the `"Are you sure you want to commit?"` confirmation prompt. Especially when accidentally entering a `commit --amend` and it gets commited desipte canceling, it gets a bit annoying because the state before the commit cannot be restored so easily.

It looks like the reason for this behavior is a somewhat messy conditional statement in the commit editor. 

~~Also, a visual feedback on cancellation might be nice - as it is done with the other events. But there doesn't seem to be a way to get a return value that can be used others with the other notification messages without doing some architetural changes. So I'm helping myself out with an autocmd.~~

edited: 2022/11/04

### DIFF
--- a/lua/neogit/buffers/commit_editor/init.lua
+++ b/lua/neogit/buffers/commit_editor/init.lua
@@ -31,7 +31,6 @@ function M.new(content, filename, on_unload)
 end
 
 function M:open()
-  local written = false
   self.buffer = Buffer.create {
     name = self.filename,
     filetype = "NeogitCommitMessage",
@@ -40,21 +39,17 @@ function M:open()
     modifiable = true,
     readonly = false,
     autocmds = {
-      ["BufWritePost"] = function()
-        written = true
-      end,
       ["BufUnload"] = function()
-        if written then
-          if
-            config.values.disable_commit_confirmation
-            or input.get_confirmation("Are you sure you want to commit?")
-          then
-            vim.cmd("silent g/^#/d | silent w!")
-          end
+        if
+          not config.values.disable_commit_confirmation
+          and not input.get_confirmation("Are you sure you want to commit?")
+        then
+          vim.cmd("silent %d | silent w!")
+          return
         end
-
+        vim.cmd("silent g/^#/d | silent w!")
         if self.on_unload then
-          self.on_unload(written)
+          self.on_unload()
         end
         require("neogit.process").defer_show_preview_buffers()
       end,
@@ -69,7 +64,7 @@ function M:open()
     initialize = function(buffer)
       buffer:set_lines(0, -1, false, self.content)
       if not config.values.disable_insert_on_commit then
-        vim.cmd(":startinsert")
+        vim.cmd("startinsert")
       end
 
       -- NOTE: This avoids the user having to force to save the contents of the buffer.


TimUntersberger:
> Also, a visual feedback on cancellation would be nice - as it is done with the other events. But there doesn't seem to be a way to get a return value that can be used others with the other notification messages without doing some architetural changes. So I'm helping myself out with an autocmd.

There shouldn't be a need for visual feedback. If you say no it just doesn't do anything, so there is no async part to it.

lua/neogit/buffers/commit_editor/init.lua
```diff
@@ -40,24 +40,24 @@ function M:open()
     modifiable = true,
     readonly = false,
     autocmds = {
-      ["BufWritePost"] = function()
-        written = true
```
TimUntersberger, line 44: why did you remove the written condition?

lua/neogit/buffers/commit_editor/init.lua
```diff
@@ -40,24 +40,24 @@ function M:open()
...
-            config.values.disable_commit_confirmation
-            or input.get_confirmation("Are you sure you want to commit?")
-          then
-            vim.cmd([[
-              silent g/^#/d
-              silent w!
-            ]])
-          end
+        if
+          not config.values.disable_commit_confirmation
```
TimUntersberger, line 45: this condition doesn't do the same thing. if you disable the commit confirmation you never enter into this condition

tobealive: > > Also, a visual feedback on cancellation would be nice - as it is done with the other events. But there doesn't seem to be a way to get a return value that can be used others with the other notification messages without doing some architetural changes. So I'm helping myself out with an autocmd.
> 
> There shouldn't be a need for visual feedback. If you say no it just doesn't do anything, so there is no async part to it.

Perfectly fine. Tell me if you like it via rebase or as commit, and I'll remove it. Or if you prefer go ahead push the changes and add yourself as co-author.

tobealive:
> > Also, a visual feedback on cancellation would be nice - as it is done with the other events. But there doesn't seem to be a way to get a return value that can be used others with the other notification messages without doing some architetural changes. So I'm helping myself out with an autocmd.
> 
> There shouldn't be a need for visual feedback. If you say no it just doesn't do anything, so there is no async part to it.

That's just fine. Tell me if you like it via rebase or as commit, and I'll remove it. Or if you prefer go ahead push the changes and add yourself as co-author.

lua/neogit/buffers/commit_editor/init.lua
tobealive: yes that's right. but isn't it the goal that you won't get the confirmation message if you disable it?

lua/neogit/buffers/commit_editor/init.lua
tobealive: As far as I could see, we could achieve the desired functionality with a negating guard clause condition. Which is less nested with less code. Also, I've seen the crook in here that this is the reason why commits are always send. As it would always be true due to the hard-coded save commands. 

And the rebase editor also has none: https://github.com/TimUntersberger/neogit/blob/3323380e4a110ec3bfa4d2a681f4a9634a239e87/lua/neogit/buffers/rebase_editor/init.lua#L27-L30 so I'll burned it.

Alternatively it should work to use the original with less modification and additionally add an else statement to return the function early.

tobealive:
@TimUntersberger 

> There shouldn't be a need for visual feedback. If you say no it just doesn't do anything, so there is no async part to it.

I removed the notification when a commit gets canceled.

> this condition doesn't do the same thing. if you disable the commit confirmation you never enter into this condition

Checking the functionality, I think we should be fine. Please clarify any other changes that need to be made, so I can understand and carry out the desired result if I am missing something.

tobealive:
Should be fixed with the recent updates. Therefore closing.

PR #371, opened 2022-10-12
fix: avoid error message in non-git directories

### LINKED ISSUE #349: After saving a file in directory without version control, an error occurs in the output
### Description

After saving a file in directory without version control, an error occurs in the output: `fatal: not a git repository (or any of the parent directories): .git`

### Neovim version

NVIM v0.8.0-dev
Build type: Release
LuaJIT 2.1.0-beta3

macOS 12.6

### Steps to reproduce

Steps to reproduce:
1. Edit any file in the directory without version control;
2. Save and exit (`:wq`)
3. Get an error in the output.

### Expected behavior

_No response_

### Actual behavior

```console
$ nvim test.txt +wq

fatal: not a git repository (or any of the parent directories): .git
```

bug

### Minimal config

```Lua
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
```

nanozuki:
close #349

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -433,6 +433,9 @@ local function refresh_viml_compat(fname)
   if not config.values.auto_refresh then
     return
   end
+  if #vim.fs.find(".git/", { upward = true }) == 0 then -- not a git repository
+    return
+  end
 
   refresh_manually(fname)
 end


nanozuki:
Using `vim.fs.find` to detect git repository. If current cwd is not is or in a git repository, don't refresh.

conrad760 (commented):
THANKS!!

TimUntersberger:
It seems very inefficient to traverse a lot of directories on each refresh. Isn't there a better way to check whether the cwd is in a git repo?

nanozuki:
If we want to optimize this, we can hook autocmd into the VimEnter and DirChanged events to do a check of the git directory. 
Let me try, btw, can I use the API in neovim 0.7? (autocmd)

nanozuki:
But If a directory changed from non-git directory to git directory, the 'autocmd' plan won't detect this change.

nanozuki:
@TimUntersberger I add some code to evaluate the elapsed time of finding directories:

```
  local t = os.clock()
  local not_git = #vim.fs.find(".git/", { upward = true }) == 0
  vim.notify(string.format("find git repo, cost time: %.3fs", os.clock() - t))
```
In these three situations:
- In a directory, not a git repo
- At a git repo
- At a subdirectory in a git repo.

The elapsed time is around 0.1ms (about 0.05ms - 0.2ms ).

I think this code is inefficient enough. 

You can check this code at: https://github.com/nanozuki/neogit/tree/benchmark

PR #380, opened 2022-10-25
fix: missing Neogit command

### LINKED ISSUE #379: :Neogit command is missing
### Description

I recently updated neogit and the `:Neogit` command is gone. Opening neogit using the lua api still works though (i.e `:lua require('neogit').open()`).

It seems to be caused by [this change](https://github.com/TimUntersberger/neogit/pull/358/commits/0ae9fc91b102b3c2f70de767b7003b8b8038dee3#diff-55b344b98ef0e2242a2dd20a8888e3b01e696c8b9707fcacff6a49d9075fc46a) from PR #358 where the whole `plugin/neogit.vim` file, which defined the command, got removed. I suspect the file got rewritten in Lua, but the part that defined the command at the bottom got missed.

### Neovim version

NVIM v0.8.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@HMBRW-A-001-M1-004.local

macOS 12.6

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. type `:Neogit`

### Expected behavior

I expect `:Neogit` to start neogit

### Actual behavior

`:Neogit` command is not defined.

### Minimal config

```Lua
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
```

ten3roberts:
Fixes #379, an overlook of #358


ten3roberts:
This makes the Neogit command accessible after `neogit.setup` is called.

I am unsure if it should be kept that way or if it should always be sources, regardless if setup was called.

Any opinions would be greatly appreciated.

Other than that, the fix is done, and it would just be a matter of moving the `neogit.commands.setup` to `plugin/neogit.lua`

TimUntersberger (approved):
IMO moving the `neogit.commands.setup` call into `plugin/neogit.lua` feels cleaner, but I'm fine with both. 👍

ten3roberts:
The neogit command has now been moved to `plugin/`.

The highlighting issues encountered on OS X should now also be fixed

kristiankupsch:
Hi. Sadly some of the highlighting issues on OS X are still there.
image

Issue #384, opened 2022-10-27
Strange prefix `^D` all over the place


strash:
### Description

Also appears with minimal config.

### Neovim version

NVIM v0.8.0
Build type: Release
LuaJIT 2.1.0-beta3

### Operating system and version

macOS 13.0

### Steps to reproduce

I actually don't know how to reproduce it

### Actual behavior

<img width="911" alt="Screenshot 2022-10-27 at 19 22 39" src="https://user-images.githubusercontent.com/14233263/198345522-3a192cd5-4ded-4a04-92b5-9cfee03957eb.png">

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing neogit and dependencies.")
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
```

strash:
Duplicate of #383

snowair:
![企业微信截图_638b781c-d594-46e8-be7b-387f809058f5](https://user-images.githubusercontent.com/1695133/198579833-ab3c2771-b97f-48eb-8b4e-4e4212a6b087.png)

emanuel2718:
Same behavior with default configuration

### Neovim version
NVIM v0.8.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew


![image](https://user-images.githubusercontent.com/55965894/198706742-5b348dcb-03c1-4fdb-beae-faaf95a004c5.png)

danilshvalov:
Hi! Could you test #388, please?

strash:
@danilshvalov, hi! This fixed the problem for me.
[...]

Issue #387, opened 2022-10-30
Diffview integration fails with an error


Utkarsh-khambra:
### Description

I recently updated Neogit from commit 3fab92b376655277eec15d36b75f850176b6f4d4 to 1acb13c07b34622fe1054695afcecff537d9a00a and I noticed that whenever I press `d` to see the diffs I either get an error like this 
![Screenshot from 2022-10-30 12-48-06](https://user-images.githubusercontent.com/29328319/198867036-3d1716fd-99de-497c-9d65-a35897fcae27.png)

error
```
fatal: path 'src/terminal/Line.h' is in the index, but not 'terminal/Line.h'
hint: Did you mean ':0:src/terminal/Line.h' aka ':0:./terminal/Line.h'?

```
when my cwd is not the root of git directory.
or a wrong diff as shown.
![Screenshot from 2022-10-30 12-46-59](https://user-images.githubusercontent.com/29328319/198867043-aa429a2f-9dac-48da-b942-f0726d0dc483.png)
(The actual change here is an addition of line `#include <regex>` but it shows that more lines have been added here)

Invoking Diffview manually with command `DiffviewOpen master` shows the correct diffs. Commit 3fab92b376655277eec15d36b75f850176b6f4d4 is working fine for me. 

### Neovim version

NVIM v0.9.0-dev-135-g96cf385a7
Build type: Release
LuaJIT 2.1.0-beta3

### Operating system and version

Fedora 36

### Steps to reproduce

1. Update to latest commit
2.  Open neovim from directory that is not a root of your git directory.
3.  Run neogit and press d to see diffs, this should show the error
4.  Or run neovim from root of git directory to see the wrong diffs.

### Actual behavior

I get errors when I try to see diffs for files when neovim is not started in root of git directory or wrong diffs when neovim is started in root of git directory.

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			{
				"TimUntersberger/neogit",
				requires = {
					{ "nvim-lua/plenary.nvim" },
					{
						"sindrets/diffview.nvim",
						config = function()
							require("diffview").setup({ use_icons = false, diff_binaries = true })
						end,
					},
				},
[...]

PR #388, opened 2022-10-30
fix(process): remove sending extra codes

### LINKED ISSUE #382: Color codes in git output
### Description

Follow up of #379 , sort of.

#380 removed the `^D^H^H` prefixes to "everything" on Linux for me (which is awesome)

On macOS, they're not quite gone yet. They've changed from `^D^H^H` to `^D`.

### Neovim version

NVIM v0.8.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@HMBRW-A-001-M1-004.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.8.0/share/nvim"

macOS 13

### Steps to reproduce

1. nvim -u minimal.lua in a git repo
2. :Neogit
3. observe `^D` in unpulled, unmerged commits and at the start of the commit message on Head: and Push:

### Expected behavior

No `^D` characters

### Actual behavior

image

### Minimal config

```Lua
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
```

danilshvalov:
Fixes #382, #384

### DIFF
--- a/lua/neogit/process.lua
+++ b/lua/neogit/process.lua
@@ -367,7 +367,6 @@ function Process:close_stdin()
   -- Send eof
   if self.stdin then
     self.stdin = nil
-    vim.api.nvim_chan_send(self.job, "\04")
     vim.fn.chanclose(self.job, "stdin")
   end
 end


TimUntersberger:
This is supposed to be a fix for a bug introduced in https://github.com/TimUntersberger/neogit/commit/541d7a1dc643b364eeecaa62f68b57f4c738d8b9. @ten3roberts

ten3roberts:
While writing the original code, the chanclose did not "close" stdin on the child process point of view.

This means that a pager will keep waiting for page down, or that input prompts will keep waiting for input.

The solution was to send EOT. I then disabled the git pager, but removing the EOT may cause some commands to never finish.

But yes, it can definitely be removed, and if issues crop up about freezing/unresponsive or long commands in terms of async, it may be due to it.

I am currently experiencing solving issue #387 which is caused by the pty emulation causing git to not show the whole file, but stops after a while. Setting `pty =false` solves the incorrect diffs, though will cause precommits to not emit color codes.

I am not sure if it is due to a vim/neovim issue for jobstart and on_stdout, or on gits side when running in a pty.

I am thinking the best of both worlds is to only have `commit`, `rebase`, `push`, and `pull` use `pty=true` by making it an option in the cli builder like `show_stdout`.

Regarding the issue of the output not being captured in full; I have discovered a similar issue in https://github.com/ten3roberts/recipe.nvim where a `fn.termopen` will not show the last output of a command if it terminates immediately afterwards, like a testing command.

ten3roberts:
@TimUntersberger 

How ok are you that I review and merge some PRs that aren't controversial?

I.e; do you give consent that I handle merging and issue closings?

TimUntersberger:
> @TimUntersberger
> 
> How ok are you that I review and merge some PRs that aren't controversial?
> 
> I.e; do you give consent that I handle merging and issue closings?

Go ahead :)

MartyBeGood:
no more stray `^D`s on macOS, no change in behavior on Linux! 👍

PR #391, opened 2022-11-02
Disable editing when initially opening NeogitStatus

FredeEB:
I stumbled upon this when I opened Neogit today and found through bisection that `45c1ef6e2c89920f0cca` introduced what seems like a bug to me.

When I run `:NeogitStatus<cr>` my cursor is in insert-mode, which shouldn't ever be possible in a readonly buffer. What I've done in the commit, fixes the issue, but I have no idea if it's the correct way to do it. From what I can tell, the first parameter to call is `verbose`, and the second one is unused after the change mentioned above.

I'll gladly take any suggestions

### DIFF
--- a/lua/neogit/lib/git/push.lua
+++ b/lua/neogit/lib/git/push.lua
@@ -17,7 +17,7 @@ local function update_unmerged(state)
     return
   end
 
-  local result = cli.log.oneline.for_range("@{upstream}..").show_popup(false).call(true, true):trim().stdout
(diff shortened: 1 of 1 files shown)


ten3roberts:
Is it because the console buffer opens?

Does the repo it happens in have an upstream?

FredeEB:
I don't recall, i looked at it very briefly. The diff of the commit i mentioned was small enough that I could "brute-force" a patch that fixed the issue, but it's very likely that it's just hiding a symptom

FredeEB:
To answer your question (I posted my previous comment while at work). Yes, the repo has an upstream commit

PR #407, opened 2022-11-29
fix: bufferline and 'No Name' bufs

caarlos0:
Signed-off-by: Carlos A Becker <caarlos0@users.noreply.github.com>

### DIFF
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -302,7 +302,7 @@ end
 function Buffer.create(config)
   config = config or {}
   local kind = config.kind or "split"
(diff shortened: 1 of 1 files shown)


ten3roberts:
How does this fix work with the other popup buffers? It seems very strange that the handle returned from nvim_create_buf is one index off.

What seems more likely is that two buffers are created and the last one is used, but the first one if forgotten, or something similar

agronskiy:
My ["fix" in the linked thread](https://github.com/TimUntersberger/neogit/issues/389#issuecomment-1314375284) was rather intended to surface the potential problem and invite someone knowledgable for deeper investigation, TBH I would also not push it per se, since it seems too much of a hack.

PR #414, opened 2022-12-23
Fix autocomplete 'arg_lead' matched as regex and add autocompletion to pull from elsewhere action

laytan:


laytan:
Closed in favour of #421 which is its own branch

PR #415, opened 2022-12-24
Remove NeogitCommitMessage filetype

### LINKED ISSUE #405: Use git commit file type for commits
Neovim has a git commit file type. Several plugins, like cmp-git use this. However, neogit uses its own git file type, preventing cmp-git and others from working.

Neogit should use the built-in file type 
Adjust other plugins to use the neogit type. This seems inefficient.

vinibispo:
Close #405 
- Remove NeogitCommitMessage filetype
- Turn it back to gitcommit filetype

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,13 @@ neogit.setup {
   console_timeout = 2000,
   -- Automatically show console if a command takes more than console_timeout milliseconds
   auto_show_console = true,
+  -- Persist the values of switches/options within and across sessions
+  remember_settings = true,
+  -- Scope persisted settings on a per-project basis
+  use_per_project_settings = true,
+  -- Array-like table of settings to never persist. Uses format "Filetype--cli-value"
+  --   ie: `{ "NeogitCommitPopup--author", "NeogitCommitPopup--no-verify" }`
+  ignored_settings = {},
   -- Change the default way of opening the commit popup
   commit_popup = {
     kind = "split",
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -4,6 +4,7 @@ local lib = require("neogit.lib")
 local signs = require("neogit.lib.signs")
 local hl = require("neogit.lib.hl")
 local status = require("neogit.status")
+local state = require("neogit.lib.state")
 
 local neogit = {
   lib = require("neogit.lib"),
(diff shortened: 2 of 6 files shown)


syntax/NeogitStatus.vim
```diff
@@ -5,9 +5,9 @@ endif
 source $VIMRUNTIME/syntax/gitrebase.vim
 
 syn match NeogitObjectId /^[a-z0-9]\{7,}\>/
-syn match NeogitCommitMessage /.*/ contained
-syn match NeogitBranch /\S\+/ contained nextgroup=NeogitCommitMessage
-syn match NeogitRemote /\S\+/ contained nextgroup=NeogitCommitMessage
+syn match gitcommit /.*/ contained
+syn match NeogitBranch /\S\+/ contained nextgroup=gitcommit
+syn match NeogitRemote /\S\+/ contained nextgroup=gitcommit
```
gegoune, line 10: I am not entirely sure but those probably can just be removed. Nvim comes with gitcommit syntax as well as there is treesitter parser for gitconmit now.

vinibispo: I'll try removing

CKolkey: Yea, these can be removed altogether.

CKolkey:
This change would prevent `lua/neogit/autocmds.lua:19` from working properly. Not sure the consequences of that.

gegoune:
```lua
      if api.nvim_buf_get_option(o.buf, "filetype"):find("Neogit") then
        return
      end
```
This could/should also check for `gitcommit` in that case. Not sure what implications of not checking that are, refreshing Neogit on`gitcommit` buffers' save makes little sense here I think.

CKolkey:
Yea. I've been running with this patch for a week or two now without issue, and without addressing the above autocmd.

syntax/NeogitStatus.vim
CKolkey: An update on this: These should _not_ be removed, and should be left unchanged. I was mistaken before: they have nothing to do with the NeogitGitCommit popup. These three lines are just for syntax highlighting on the status buffer of the branch name/remote name.

vinibispo:
Sorry guys, I'll promise that I'll fix it

PR #416, opened 2022-12-28
Add config for automatically showing console

bradsherman:
The log for slow running commands is nice, but I don't want it to pop up all the time and disrupt my flow. This allows users to still collect logs for slow running commands according to their `console_timeout` value, but will only automatically show the console (as it does now) if `auto_show_console` is `true`. I tested this by using my branch in my local setup and making a really small `console_timeout` value and setting `auto_show_console` to `true` first to confirm current behavior and then setting it to `false` to ensure that the console didn't pop up. I also pressed `#` to make sure that the log lines were still being accumulated.


ten3roberts:
Looks awesome.

Is this ready to merge, or do you want to add anything else?

bradsherman:
@ten3roberts Thanks! Glad we are on the same page. I am ready for it to be merged if it looks good to you.

Issue #392, opened 2022-11-03
Should branches be sorted by last used/checked out or last commit


ten3roberts:
**Is your feature request related to a problem? Please describe.**
When picking and switching between branches it can be difficult to find the branch you recently worked on if you have many branches as it may become buried

**Describe the solution you'd like**
This could be implemented as an option in the config, but the question is what the default order should be

Issue #144, opened 2021-06-15
Branches sorting by latest checkout


timsofteng:
Hey! Is it possible to sort git branches by time of latest checkout into it?

TimUntersberger:
Do you mean the autocompletion?

timsofteng:
> Do you mean the autocompletion?

Completion is necessary as well but no I mean sorting. Try to use `lazygit`. Create several branches and checkout to the las and. And then to last one. You will see sorting orged by last checkout in list of branches.

I believe in current realization it could be implemented with popup or with `telescope`. E.g you are in command line then press some keybinding which open telescope popup. And on enter it return branch which user need.

But the main question is sorting.

timsofteng:
Hello!
Any news about this feature?

TimUntersberger:
> Any news about this feature?

I haven't worked on this yet, but I am willing to accept any PRs.

timsofteng:
@TimUntersberger unfortunately I'm not good in lua.
It something like additional option to `git`.
Additionally it would be cool to have alternative to `:G branch` in `fugitive` to see all branches and checkout between them.

timsofteng:
@ten3roberts Do you think it is completed? How can I sort branches by last checkout?

CKolkey:
Sure :)

```
 -- Value used for `--sort` option for `git branch` command
  -- By default, branches will be sorted by commit date descending
  -- Flag description: https://git-scm.com/docs/git-branch#Documentation/git-branch.txt---sortltkeygt
  -- Sorting keys: https://git-scm.com/docs/git-for-each-ref#_options
  sort_branches = "-committerdate",
```

Field names are here: https://git-scm.com/docs/git-for-each-ref#_field_names

But you'll want to use the default value of `-committerdate`

https://stackoverflow.com/questions/5188320/how-can-i-get-a-list-of-git-branches-ordered-by-most-recent-commit

timsofteng:
@CKolkey Sorry but I'm not sure I've get what I mean. Commit date and date of last checkout are different things. I could checkout to another branch but do nothing but I expect it on the top of the list.

PR #433, opened 2023-01-12
Add custom option that allows to sort branches

ivaaaan:
When I checkout, I want to see branches sorted by commit date. 
This change allows user to specify their own sort settings.

Flag description: https://git-scm.com/docs/git-branch#Documentation/git-branch.txt---sortltkeygt
Sorting keys: https://git-scm.com/docs/git-for-each-ref#_options

#144 #392

### DIFF
--- a/lua/neogit/config.lua
+++ b/lua/neogit/config.lua
@@ -9,6 +9,7 @@ M.values = {
   disable_insert_on_commit = true,
   use_magit_keybindings = false,
   auto_refresh = true,
+  sort_branches = "-commiterdate",
   kind = "tab",
   -- The time after which an output console is shown for slow running commands
   console_timeout = 2000,
--- a/lua/neogit/config.lua
+++ b/lua/neogit/config.lua
@@ -9,7 +9,7 @@ M.values = {
   disable_insert_on_commit = true,
   use_magit_keybindings = false,
   auto_refresh = true,
-  sort_branches = "-commiterdate",
+  sort_branches = "-committerdate",
   kind = "tab",
   -- The time after which an output console is shown for slow running commands
   console_timeout = 2000,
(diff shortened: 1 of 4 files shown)


lua/neogit/config.lua
```diff
@@ -9,6 +9,7 @@ M.values = {
   disable_insert_on_commit = true,
   use_magit_keybindings = false,
   auto_refresh = true,
+  sort_branches = "-commiterdate",
```
ivaaaan, line 12: Maybe this one should be changed to `nil`/something else.

ten3roberts:
This looks fantastic!

Is there anything more you'd like to add or update, since it has been some time you implemented the feature and it may thus need some tweaking?

Otherwise, I am more than happy to merge

ten3roberts (approved):

ivaaaan:
Thanks for the review :) Not really, the only thing I don't like is that it shows remote branches next to the local branches. I would hide remote branches, but not sure how it will affect others. 
image

asymmetric:
Maybe documenting this could be a good idea? Or is this self-documenting for end-users through the UI somehow?

ten3roberts:
> Maybe documenting this could be a good idea? Or is this self-documenting for end-users through the UI somehow?

Agreed, perhaps a mention in the readme would suffice, otherwise it is quite self documenting.

ten3roberts:
Closes: #392

ivaaaan:
I have updated  README

ten3roberts (approved):

ten3roberts:
I can merge this after we fix the stylua linting.

Thanks for the PR, as this is something I've been needing a long time

ivaaaan:
@ten3roberts linting seems to be fine now, at least locally :) 
Thanks for your work.

timsofteng:
@ivaaaan may you descibe please how to sort branches by last checkout as like lazygit do?

Issue #439, opened 2023-01-28
Canceling an amend still sends a commit


ttytm:
### Description

I would like to be able to cancel amending to the last commit and do not have it commited anyway.

Especially while amending it's important to have cancellation working, as it can get time consuming to restore the previous state.

For now, I quit the editor(lol :smiling_face_with_tear: ) if I enter a commit amend dialog and want to stop the amend process.

### Neovim version

stable v0.8.x | nightly

### Operating system and version

Linux

### Steps to reproduce

Trigger commit amend in neogit <kbd>ca</kbd> then e.g., <kbd>q</kbd> to quit.

### Expected behavior

Commit won't get amended when just quitting without saving / confirming.

### Actual behavior

Commit amend happens anyway.

### Minimal config

```Lua
-
```

ten3roberts:
Does this issue still happen?

tobealive:
I'm using #441 since months as it also solves this.

But it still does happen checking back on current master 36ffabb09974d3cdc3f3b31a639280da36a2b79f

PR #441, opened 2023-01-30
 fix: compatibility issue with noice

ttytm:
This PR fixes some issues that are related to using the `"BufUnload"` event as trigger for the commit action. 

It adds a dedicated action for sending a commit. This allows for things like checking the status of the confirmation **first** and then processing further logic (like sending a close request to the buffer). Instead of having all the logic bound to when the buffer is closed. This simplifies and allows more control over what is happening.

**The cost of this is that a commit needs to be sent explicitly instead by closing the buffer e.g. `:wq`**. This explicit way is also the way magit is doing it  https://github.com/NeogitOrg/neogit/pull/441#issuecomment-1419974326.

The action is mappable and defaults to `<leader>c`. (A user can assign multiple mappings, for me personally this also contains `;q` which I use as my usual "quick-confirm-quit" mapping instead of `:wq`). And while adding a "commit" related mapping table, the ability to customize the "close" mapping - which is currently hard coded to `"q"` - was added as well.

```lua
mappings = {
  commit = {
    close = { "q" },
    send = { "<leader>c" },
  },
  ...
}
```

Last, the setting `disable_commit_close_on_deny` was added. It allows to configure if the commit buffer should stay open when a user denies the confirmation dialog of a commit send action. To not deprecate / break anything, it follows the current scheme of options. Personally I would prefer:
```lua
commit_confirmation = {
  enabled = true,
  close_on_deny = false,
}
-- instead of
disable_commit_confirmation = false,
disable_commit_close_on_deny = false,
```

Let me know how it works for you and if any changes are desired.

---
<em>edit(2023.07.15): update description after rebase on master</em>

### DIFF
--- a/lua/neogit/buffers/commit_editor/init.lua
+++ b/lua/neogit/buffers/commit_editor/init.lua
@@ -29,8 +29,44 @@ function M.new(filename, on_unload)
   return instance
 end
 
+local confirmed = config.values.disable_commit_confirmation
+
+-- Define mappable actions
+local map_actions = {
+  send = function(buffer)
+    if not confirmed then
+      confirmed = input.get_confirmation("Are you sure you want to commit?")
+      if confirmed and vim.bo.mod then
+        vim.cmd("silent w!")
+      end
+    end
+    if not confirmed and config.values.disable_commit_close_on_deny then
+      return
+    end
+    buffer:close(true)
+  end,
+  close = function(buffer)
+    buffer:close(true)
+  end,
+}
+
+-- Assign actions to keys (single keys or tables of keys)
+local mappings = { n = {} }
+for action, command in pairs(map_actions) do
+  local action_mapping = config.values.mappings.commit[action]
+
+  if type(action_mapping) == "string" then
+    mappings.n[action_mapping] = command
+  elseif type(action_mapping) == "table" then
+    for _, map in pairs(action_mapping) do
+      mappings.n[map] = command
+    end
+  end
+end
+
 function M:open()
-  local should_commit = false
+  confirmed = config.values.disable_commit_confirmation
+
   self.buffer = Buffer.create {
     name = self.filename,
     filetype = "NeogitCommitMessage",
@@ -41,41 +77,21 @@ function M:open()
     readonly = false,
     autocmds = {
       ["BufUnload"] = function(o)
-        local buf = Buffer.create {
-          name = o.buf,
-        }
-        if not should_commit and buf:get_option("modified") then
-          if
-            not config.values.disable_commit_confirmation
-            and not input.get_confirmation("Are you sure you want to commit?")
-          then
-            -- Clear the buffer, without filling the register
-            buf:clear()
-            buf:call(function()
-              vim.cmd("silent w!")
-            end)
-          end
+        if not confirmed then
+          vim.api.nvim_buf_set_lines(o.buf, 0, -1, false, {})
+          vim.api.nvim_buf_call(o.buf, function()
+            vim.cmd("silent w!")
+          end)
         end
 
-        if self.on_unload and not should_commit then
-          self.on_unload(true)
+        if self.on_unload then
+          self.on_unload()
         end
 
         require("neogit.process").defer_show_preview_buffers()
       end,
     },
-    mappings = {
-      n = {
-        ["q"] = function(buffer)
-          if not buffer:get_option("modified") then
-            buffer:close(true)
-          elseif input.get_confirmation("Commit message hasn't been saved. Abort?") then
-            should_commit = true
-            buffer:close(true)
-          end
-        end,
-      },
-    },
+    mappings = mappings,
     initialize = function(buffer)
       vim.api.nvim_buf_call(buffer.handle, function()
         local disable_insert = config.values.disable_insert_on_commit
--- a/lua/neogit/config.lua
+++ b/lua/neogit/config.lua
@@ -269,6 +269,10 @@ function M.get_default_values()
       "NeogitCommitPopup--allow-empty",
     },
     mappings = {
+      commit = {
+        close = { "q" },
+        send = { "<leader>c" },
+      },
       finder = {
         ["<cr>"] = "Select",
         ["<c-c>"] = "Close",
@@ -723,7 +727,7 @@ function M.setup(opts)
     M.values = vim.tbl_deep_extend("force", M.values, opts)
   end
 
-  local config_errs = M.validate_config()
+  --[[ local config_errs = M.validate_config()
   if vim.tbl_count(config_errs) > 0 then
     local header = "====Neogit Configuration Errors===="
     local header_message = {
@@ -750,7 +754,7 @@ function M.setup(opts)
       ),
       vim.log.levels.ERROR
     )
-  end
+  end ]]
 end
 
 return M


lua/neogit/buffers/commit_editor/init.lua
```diff
@@ -40,25 +39,24 @@ function M:open()
     modifiable = true,
     readonly = false,
     autocmds = {
-      ["BufWritePre"] = function()
-        written = true
-      end,
```
tobealive, line 45: Removed as there is no need to mimic vims `modified`.

lua/neogit/buffers/commit_editor/init.lua
```diff
@@ -40,25 +39,24 @@ function M:open()
...
+        then
+          -- Clear the buffer, without filling the register
+          vim.api.nvim_buf_set_lines(o.buf, 0, -1, false, {})
+          vim.api.nvim_buf_call(o.buf, function()
+            vim.cmd("silent w!")
+          end)
         end
 
         if self.on_unload then
-          self.on_unload(written)
```
tobealive, line 61: Can't be anything else then `true` at this point. As we check for `written` above.
Going through the code base it's unclear to me why the bool value is passed here at all.  If you have the change to clarify maybe it gives the insight for further improvements.

lua/neogit/buffers/commit_editor/init.lua
```diff
@@ -40,25 +39,24 @@ function M:open()
...
       ["BufUnload"] = function(o)
-        if written then
-          if
-            not config.values.disable_commit_confirmation
-            and not input.get_confirmation("Are you sure you want to commit?")
-          then
-            -- Clear the buffer, without filling the register
-            vim.api.nvim_buf_set_lines(o.buf, 0, -1, false, {})
-            vim.api.nvim_buf_call(o.buf, function()
-              vim.cmd("silent w!")
```
tobealive, line 55: Why not just `return`? Doing exploratory testing through some scenarios I couldn't find a difference using this vs a simple return.

tobealive:
The second commit applies the changes stated in the third part of the review above. 

As a nice side effect a commit amend won't get sent anymore when canceled. Fixes #439

tobealive:
The third commit improves compatibility with noice. By correctly showing the confirmation dialog.



Related issue: https://github.com/folke/noice.nvim/issues/232

Please let me know what you think and if there should be any changes.

lua/neogit/buffers/commit_editor/init.lua
```diff
@@ -40,25 +39,21 @@ function M:open()
...
+          -- Clear the buffer, without filling the register
+          vim.api.nvim_buf_set_lines(o.buf, 0, -1, false, {})
+          vim.api.nvim_buf_call(o.buf, function()
+            vim.cmd("silent w!")
+          end)
         end
 
         if self.on_unload then
-          self.on_unload(written)
+          self.on_unload(vim.bo.mod)
```
tobealive: should be 
```lua
self.on_unload(not vim.bo.mod)
```

even though this should get the right buffer. If you want to narrow it down to no other possibility than the unloaded buffer, we could add `vim.bo[o.buf].mod`

abusch:
I haven't tried this yet, but I'm curious: what happens if I do `:wq` instead of `<leader>c`?

tobealive:
It would close the commit split without committing. 

The committing action must be triggered before the close event if a confirmation dialog is to work as a popup. 
At least this was the only option I could see without a bigger architectural change. That's why this adds the dedicated action for sending commits, hoping that it is not to far from the spirit of magit.

To give an insight from my personal config: 
`;q` and `;w` is mapped to speed up their command line equivalents (`;` still triggers a repeat after timeoutlen if one would use it for `f`,`t`, etc.).
```lua
-- Quick Commands
map("n", ";w", "<Cmd>w<CR>")
map("n", ";q", "<Cmd>confirm quit<CR>")
``` 
For neogit now i have mapped `;q` to the send commit action. Which sends the commit and closes the buffer after confirmation.

abusch:
FWIW, this is also the approach `magit` uses:

```
C-c C-c (with-editor-finish)
Finish the current editing session by returning with exit code 0. Git then creates the commit using the message it finds in the file.

C-c C-k (with-editor-cancel)
Cancel the current editing session by returning with exit code 1. Git then cancels the commit, but leaves the file untouched.
```

TimothyPescatore:
Any update here?

wrightjjw:
This additionally looks like it could be a solution to our discussion in #552. It does change the core functionality of `:wq`... not a fan tbh. maybe a community poll to see how people would feel about that?

CKolkey:
I've been thinking about it, and seeing as noice is explicitly experimental and says up front that stuff might break... going out of our way to support it seems silly.

tobealive:
@wrightjjw yes, it would solve it. Using this branch since months, it also solves the linked commit amend issue. Also @CKolkey's reasoning that noice is experimental is valid. 

So I wouldn't sell this PR anymore as fixing noice mainly, but the other issues and compatibility noice gets fixed along the way.

I'll rebase it on current master to get all the sweet recent commits and re-author the description to be a little easier to follow. Maybe a community poll would still be a thing, but deciding that is beyond my powers.

CKolkey:
If it solves the amend issue, then _that_ is something I'd say is worth looking into.

ttytm:
@CKolkey any update in regards of merging this change? I'm still relying on the branch and would work through the merge conflicts to make it ready.

CKolkey:
So, I think the biggest issue I have with this is that it changes the default experience so a user can't just save the buffer and close it. I don't have a lot of time right now to think about this, but two things come to mind:
1. `BufWritePost` autocmd to invoke the "send" action
2. use a conditional to check for `noice` integration, and disable the `BufUnload` autocmd if it's present.

ttytm:
> So, I think the biggest issue I have with this is that it changes the default experience so a user can't just save the buffer and close it. I don't have a lot of time right now to think about this, but two things come to mind:
> 
> 1. `BufWritePost` autocmd to invoke the "send" action
> 2. use a conditional to check for `noice` integration, and disable the `BufUnload` autocmd if it's present.

BurwritePost would also change default behavior. Personally I wouldn't want a send action to be invoked after it.

No. 2 is genius I think.

CKolkey:
I think this is addressed here: https://github.com/NeogitOrg/neogit/pull/1038

No more confirmations in the autocmd. Thanks for the ideas and discussion.

ttytm:
@CKolkey thanks for the work. I tried the changes. Cancelling an amend will still send a commit. 

Repor:
- <kbd>ca</kbd> to ammend:
- Change something in the commit message - don't save the buffer
- <kbd>q</kbd> to quit. Hit <kbd>n</kbd>o in the saves changes confirmation dialog.
- See the commit sha changing

This was not the case with the PR.

CKolkey:
Ahh, yeah, 'Close' isn't going to abort. Only 'Abort' (default c-c c-k, like Magit) will do that. I kept Close around to not break people's workflows too much, but I think you're better off with the more explicit commands. Do you think thats reasonable?

ttytm:
Yep it's playing nicely with neogit now.

I think the fix of aborting when closing is canceled is fairly simple. I just opened a PR.

CKolkey:
Alright, thats a good point.

Issue #445, opened 2023-02-13
Allow NeogitCosole to have its kind set.


Kraust:
**Is your feature request related to a problem? Please describe.**
I have a use cases where I am using `kind=floating` and both commit_popup and popup also set to `kind=floating`. Whenever the NeogitConsole pops up (e.g. during a pre-commit hook) it is obstructed by the floating window. This process also makes the floating window lose focus without closing which may be unintentional.

**Describe the solution you'd like**
Allowing the NeogitConsole to be rendered in the floating window in a similar way to how `commit_popup=floating` works. If no error is reported, transition to the commit message dialog, otherwise, keep the console open and allow the user to exit the floating window or go back to status.

**Describe alternatives you've considered**
Allow the option of automatically closing the NeogitConsole while also being able to set the length of the popup dialog timeout as its currently too short to properly read the error message in my case.

**Additional context**

ten3roberts:
As for the alternative solution. If you disable the builtin notification system in config `disable_builtin_notifications=true`, you can use `nvim-notify` and `:Notifications` to view the failed commands.

That is what I do most of the time

PR #446, opened 2023-02-14
Added the ability to set the kind of buffer used for preview_buffer.

Kraust:
Ended up resolving #445 by myself. Basically, I added a new configuration option preview_buffer.kind which will allow the user to select what kind of buffer the preview_buffer/NeogitConsole renders as. While doing this I also resolved an issue where if the preview_buffer as a floating window was unfocused it could not be re-focused again after it loses focus.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,10 @@ neogit.setup {
   commit_popup = {
     kind = "split",
   },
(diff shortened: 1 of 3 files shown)


ten3roberts:
Looks great, there just seems to be a small formatting issue.

PR #448, opened 2023-02-23
feat: Implement fetch popup

pritchett:
This adds a popup with what I believe to be the core functionality of fetching.


NarrowCode:
Hey, this seems to be a very nice addition.
Is there anything that prevents this PR from being merged?

lua/neogit/popups/fetch.lua
```diff
@@ -0,0 +1,57 @@
...
+    end)
+    :action("a", "Fetch from all remotes", function(popup)
+      local args = popup:get_arguments()
+      table.insert(args, "--all")
+
+      fetch_from("all remotes", "", "", args)
+    end)
+    :action("e", "Fetch from elsewhere", function(popup)
+      local remote = input.get_user_input("remote: ")
+      local branch = git.branch.prompt_for_branch()
```
ten3roberts, line 47: FYI there is a non-blocking and much prettier split based branch selection now

https://github.com/TimUntersberger/neogit/blob/efc1764060e64cc8e81687319d5223b3246e5871/lua/neogit/popups/merge.lua#L26-L37

ten3roberts (changes requested):

lua/neogit/popups/fetch.lua
CKolkey: IMO its easier to leave as is - I completely replaced the branch picker interface with telescope. For simplicity's sake I made it a required dependency, so we don't need to maintain two parallel implementations. Updating this PR just adds a conflict to resolve 🤷

ten3roberts: Good point, was just a consideration, and is not strictly necessary to merge. There are many other places where the old branch picker is used.

CKolkey: Totally fair - the new branch picker _is_ an improvement. I was thinking about it on my way home from work and it could be a good strategy to roll out the dependency - do something like:
```lua
local ok, tele = pcall(require, "telescope")
if ok then
  -- Use telescope picker
else
  -- Print a notice to add dependency and fallback to OG picker
end
```

ten3roberts (approved):

Issue #404, opened 2022-11-18
Neogit leaves empty buffer every time it's open


Strongleong:
### Description

Every time I open Neogit it creates new empty buffer.

Just opened file "test.txt":
![image](https://user-images.githubusercontent.com/17177071/202617438-5aff9494-7e45-43c6-81f4-9b77d0bdf90c.png)

Opened and closed Neogit:
![image](https://user-images.githubusercontent.com/17177071/202617498-9ea3639c-ec13-4d70-b51e-ae469b3db708.png)

Opened and closed Neogit for the second time:
![image](https://user-images.githubusercontent.com/17177071/202617522-cf1bc57f-0902-4c02-bffd-e73f1a0eb5b0.png)

Bug reproduces with "minimal.lua" config

P.S. Sorry for bad English :sweat_smile: 

### Neovim version

NVIM v0.8.1
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

### Operating system and version

Arch Linux

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. check `:buffers`
3. open Neogit `:Neogit`
4. check `:buffers` again, note that second buffer appears
5. close Neogit `q`
6. check `:buffers` again, note that second buffer did not disappeared

### Expected behavior

No additional buffers created

### Actual behavior

Empty additional buffers created every time I open Neogit

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing neogit and dependencies.")
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
```

[...]

PR #449, opened 2023-02-25
chore: close buffer by its underlying window, when available

Fabian-programmer:
the changes from #404 created problems, when https://github.com/tiagovla/scope.nvim was also used.

Closing a buffer created a new No Name buffer + window. Which screwed the window layout.
The changes address closing the buffer by its underlying window, when available.

Probably it also fixes this:
https://github.com/TimUntersberger/neogit/issues/402
https://github.com/TimUntersberger/neogit/issues/235
I tested it and the behaviour was expected. (I didn't test if this PR solved these issues)

### DIFF
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -92,14 +92,20 @@ function Buffer:close(force)
   if force == nil then
     force = false
   end
-  if self.kind == "tab" then
-    -- `silent!` as this might throw errors if 'hidden' is disabled.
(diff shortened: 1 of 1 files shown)


Fabian-programmer:
@ten3roberts would you like to review it? :)

ten3roberts:
Looks great and thanks for the help.

I had a hunch the "close previous buffer" approach by #404 may break some other plugin.

If the other mentioned issues become fixed by this as well, that would be fantastic; only time will tell.

PR #450, opened 2023-02-25
Make all file modes equal length in Status

CKolkey:
This is a fairly small change, the effect of which is to vertically align filepaths in the status window:

Screenshot 2023-02-25 at 13 04 42

Compare that with how it is now:
Screenshot 2023-02-25 at 13 07 18

Much more readable!

(ignore the colors - thats just some fun I'm having)

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -85,14 +85,14 @@ local function get_current_section_item()
 end
 
 local mode_to_text = {
-  M  = "Modified     ",
-  N  = "New file     ",
-  A  = "Added        ",
-  D  = "Deleted      ",
-  C  = "Copied       ",
-  U  = "Updated      ",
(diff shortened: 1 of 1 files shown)


gegoune:
How about slightly smaller gap? It might matter for those with small screens.

CKolkey:
The only reason for that size is the "Both Modified" label. I kinda agree its a bit big.

gegoune:
I am not a maintainer here so don't have any day, but didn't take that case into consideration. Thanks for explaining. It's a super simple yet very qol improving pr!

CKolkey:
Hmm. What if we just do something like:

```lua
if vim.o.columns < 80 then
  label = vim.trim(label)
end
```

(Nothing meaningful about 80, could be anything really.)

ten3roberts:
That works as well

PR #453, opened 2023-02-28
Initial implementation for persisting state

CKolkey:
Adds a persistence layer for all switches/options. Users can disable this entirely via config `remember_settings`, it can be project-scoped (default) or global, and individual settings can be ignored by users, if they so desire.

I'm not sure how to expose the "reset" function right now, but it can be called directly if needed.


ten3roberts:
Project scoping as a default seems great.

Is there anything more you'd like to add, or should this be merged?

CKolkey:
I'm happy with it if you are :)

CKolkey:
That works nicely. Ok, good to go!

Fabian-programmer:
Thank you for the feature.
I just have an annotation. 
I'd say that a sane configuration is to disable it by default.
If someone force-pushes after a rebase. The force push setting is now active for the next commit.
For force-pushing I certainly want to opt-in instead of opt-out.

What do you think @CKolkey @ten3roberts

CKolkey:
Yea, I think thats right. In fact, I have force push and force with lease disabled in my config.

PR #454, opened 2023-02-28
Refactor diff.lua to speed up large files in index

CKolkey:
I did some profiling for the general sluggishness when a very large file is present in the index, and it came up that `parse_diff()` was the culprit. From there, I started teasing the different functions out of that larger one to find where the problem came from, and low-and-behold, it was building hunks.

So, with some metatable magic, we can defer computing the hunks until somebody tries to actually _render_ it. That way, if you are just doing some staging/committing with the folds closed, you never pay for something that you never see.

I did change a couple things beyond that, though:

1. `parse_diff()` took a second argument, `with_stats`, but I couldn't find a single caller who actually passed that argument. So, it's removed.
2. Replacing calls to `vim.startswith("...")` with `<string>:match("^...")`. Profiling called these out as hot-spots, as you can see below.

I haven't done a lot with metatables/lazy-evaluation in Lua before, so if there are ways to improve it, I'm all ears.

To test the effect of this, on master, make a file with 10,000-20,000+ lines, open neogit, and stage it. Wait. Close neogit, reopen it, wait some more. Then checkout this branch, and observe the lack of waiting.

For compairison, here's a profile of staging and unstaging a (folded) 100,000 line file on current master:
Screenshot 2023-03-01 at 11 06 25

Mind you, we never actually _rendered_ the diff, so all this is wasted work.

Here's this branch, same action, same file:
Screenshot 2023-03-01 at 11 06 59

Of course, if the user goes to render the diff, then all this work still needs to be performed. But one problem at a time :)

### DIFF
--- a/lua/neogit/lib/git/diff.lua
+++ b/lua/neogit/lib/git/diff.lua
@@ -32,72 +32,82 @@ local function parse_diff_stats(raw)
   return stats
 end
 
-local function parse_diff(output, with_stats)
-  local diff = {
-    kind = "modified",
-    lines = {},
-    file = "",
-    info = {},
-    hunks = {},
-    stats = {},
-  }
+local function build_diff_header(output)
+  local header = {}
   local start_idx = 1
 
-  if with_stats then
-    diff.stats = parse_diff_stats(output[1])
-    start_idx = 3
-  end
-
-  do
-    local header = {}
-
-    for i = start_idx, #output do
-      if output[i]:match("^@@@*.*@@@*") then
-        start_idx = i
-        break
-      end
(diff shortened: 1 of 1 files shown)


CKolkey:
I'll sort out the failing tests tomorrow :)

CKolkey:
Specs seem good, so as far as I can tell this is good to go. Lemme know if you want anything changed/added.

ten3roberts:
Looks good, let me know if you are ready to merge

CKolkey:
Yup! This guy seems totally stable to me :)

PR #455, opened 2023-03-01
Performance Tuning for Status Buffer with Large Files

CKolkey:
I did some more performance tuning with very large hunks (100,000 lines). It turns out that adding signs for context was _incredibly_ expensive, and was rerun on `CursorMove`. To fix this up, I did a few things:

1. Use decorator API to handle decorating the buffer.
2. Replace vim.regex usage with lua string comparison. Since lua interns strings, comparing them is just checking if two pointers are the same, with is blindingly faster than parsing regular expressions/patterns.
3. Added `NeogitDiffContext` highlight group. This will fill the background of a diff when it's _not_ the current context, if it isn't an add/delete line.

### DIFF
--- a/lua/neogit/buffers/common.lua
+++ b/lua/neogit/buffers/common.lua
@@ -12,8 +12,8 @@ local range = util.range
 
 local M = {}
 
-local diff_add_matcher = vim.regex("^+")
-local diff_delete_matcher = vim.regex("^-")
+local diff_add_start = "+"
+local diff_delete_start = "-"
 
 M.Diff = Component.new(function(diff)
   local hunk_props = map(diff.hunks, function(hunk)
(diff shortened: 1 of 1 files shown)


CKolkey:
Think it's good to go :)

CKolkey:
Let's call this good-to-go :)

mknurs:
Hey, this fixes all the highlighting bugs I've been having. But why remove the CursorMoved callback? Shouldn't the hunks be highlighted as the cursor moves to them?

CKolkey:
Yes! Check out the function "set_decoration_provider" in status.lua. Specifically "on_win". That function gets called every time the buffer highlights need to be repainted, and the cursor moving triggers a repaint. This has a couple advantages:
1) repainting signs is more expensive than repainting extmarks (idk why)
2) the existing implementation isn't clamped to visible lines, so we repaint ALL signs EVERY time the cursor moves. This results in poor performance with large diffs
3) the decoration provider api can also be used in the commit view, so Im working on a way to have a single "show context" approach that works in every buffer

PR #461, opened 2023-03-06
feat: use install action for selene

ten3roberts:


CKolkey:
I think we can probably close this?

PR #466, opened 2023-03-08
Cherry pick popup

CKolkey:
There are still some places missing cherry-pickability, but for now you can use `A` from the status buffer to open commit list. If your cursor is over a commit, it will be chosen on the list, and if you are in visual mode with 1 or more commits selected, they will be automatically used.

On the log buffer, normal mode is working (single commit), visual mode isn't.

More to come, but I'm going to put a pin in this for now.


MaxKiv:
Would be cool to have this feature! What is its status?

CKolkey:
Working on merging a bunch of things in right now, and this is included :) (see #503 )

PR #468, opened 2023-03-09
MD5 Hash state file to prevent issues on windows

CKolkey:
Pretty simple, just pass the `cwd` string into md5 to get a valid filepath for windows/mac/etc


CKolkey:
This is probably the way to go: https://github.com/TimUntersberger/neogit/pull/470

PR #474, opened 2023-03-13
Add telescope integration for fuzzy-searching branches

CKolkey:
Adds optional telescope integration, used when searching for a branch


CKolkey:
I've gotta rebase this, it's only 4 files -_-

PR #485, opened 2023-04-14
fix: partition diff checks to prevent too many open files error

laytan:
If there are many staged or unstaged files in the repo, each of these would have a `git diff` check performed on them. This caused the error "too many open files" because too many async requests were being done at a time.

This change partitions these `git diff` calls to be done in chunks of 50, which during testing did not cause the error.

### DIFF
--- a/lua/neogit/lib/collection.lua
+++ b/lua/neogit/lib/collection.lua
@@ -68,6 +68,19 @@ function M.find(tbl, func)
   return nil
 end
 
+function M.partition(tbl, max)
+  local res = { {} }
+  local i = 1
+  for j, item in ipairs(tbl) do
+    if j % max == 0 then
+      table.insert(res, {})
+      i = i + 1
+    end
+    table.insert(res[i], item)
+  end
+  return res
+end
+
 return setmetatable(M, {
   __call = function(_, tbl)
     return M.new(tbl)
--- a/lua/neogit/lib/git/diff.lua
+++ b/lua/neogit/lib/git/diff.lua
@@ -207,8 +207,6 @@ end
 function diff.register(meta)
   meta.load_diffs = function(repo, filter)
     filter = filter or false
-    local executions = {}
-
     if type(filter) == "table" then
       filter = ItemFilter.new(Collection.new(filter):map(function(item)
         local section, file = item:match("^([^:]+):(.*)$")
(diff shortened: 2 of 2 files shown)


CKolkey:
Check out what I did here: https://github.com/CKolkey/neogit/blob/master/lua/neogit/lib/git/diff.lua#L182

First, each diff is lazily loaded when the hunk is viewed (when the `#diff` value on the hunk's table is called), and since the actual CLI call and parsing is in a closure, the CLI is only called at the moment the diff is needed. 

Not that there's anything wrong with this implementation.. I've just spent a lot of time optimising this part of the plugin :)

Hopefully I'll be able to upstream all my work at some point.. But you can test out my fork with thousands of files, each with thousands of lines changed, and loading diffs is quick, the UI doesn't freeze... It's fast as hell, quite frankly.

laytan:
Damn @CKolkey nice work over there, a bunch of changes I was missing (cherry pick, fuzzy, annoying things fixed)!

I hope this can get into upstream at some point, so many changes are probably hard to get in at once though, I will switch to your fork for now.

I did find 2 issues with it I thought I should mention (you might already know):

1. The popup screens (other than commit) seem weird, I don't know what it is doing but see these screenshots, for commits and for branches:
Screenshot 2023-04-14 at 22 25 12

Screenshot 2023-04-14 at 22 27 09

2. Diff highlighting looks weird (black background) any hovered/selected item also has a black background now
 
Screenshot 2023-04-14 at 22 27 56

Anyways, thanks for letting me know about your changes!

CKolkey:
Thanks! I've been having fun. 

Per your comments:

1. Im not sure I understand the issue? Maybe you could clarify? The images look fine
2. Yea, thats shouldn't be black. I was messing with the highlights yesterday, extracting everything from the syntax files and putting them all together in hl.lua. Is your config available to test out?

laytan:
@CKolkey Thanks, the background color is better now. For the screenshots I thought it was showing "verbose" stuff under the variables of the branch popup, when the commit popup kind of has explanations instead.

But I have never seen or used magit so it might be on me and I will look into how to use that part.

CKolkey:
Ahh, ok, yea it's not that it's "verbose", more that neogit (upstream) just hasn't implemented everything yet. The popups (as implemented in my fork) are as close to the magit version as I could make them. The variables exposed are just git variables for the given situation, so you can have even more control over how git handles things :)

As for the commit buffer, I always just use [q](https://github.com/CKolkey/neogit/blob/master/lua/neogit/buffers/commit_editor/init.lua#L69) to close it. And I haven't changed anything there, so I'm not entirely sure why it would be different than the upstream version. Do you have any issues if you just use `q`?

PR #491, opened 2023-05-01
Remove incorrect merge with the diffview config.

asmodeus812:
This fixes the following issue https://github.com/TimUntersberger/neogit/issues/440. @ten3roberts


asmodeus812:
Now diffview will always work with whatever configuration the user has already configured it with.

PR #500, opened 2023-05-09
Upstreaming 1/n - lib/git/*

CKolkey:
@ten3roberts I'm thinking about how to split this all up - welcome your feedback on it. It seems like it's almost entirely additions, so hopefully nothing breaks.

That said, I think the easiest way to handle this is to do this branch-stack style, and maybe merge top-down? I donno. Lets try that, and you can tell me if it works :)

edit:
I don't think I can stack them properly since the base branch is on my fork, and I don't have write access to this repo. Annoying. I'll try to figure out a nicer way, but for now, here are the next steps in the stack:
- https://github.com/CKolkey/neogit/pull/28
- https://github.com/CKolkey/neogit/pull/29
- https://github.com/CKolkey/neogit/pull/30


ten3roberts:
Great.

I am not familiar with what merge top-down is, so please enlighten me.

Othewise it looks great.

What method are you using to split apart the branches into separate ones? Rebase from BASE and discard or something else?

CKolkey:
Oh, by top-down I just mean that if we have:

`master` <- `branch1` <- `branch2` <- `branch3`

Where the arrows indicate the base of the branch.

We would start by merging `branch3` into `branch2`
...then `branch2` into `branch1` (which now includes `branch3`)
...then `branch1` into `master` (which includes `branch2` and `branch3`). 

Does that make sense?

And to do this, all I did was pull a fresh master from this repo (`upstream` from my point of view), and run: `git reset --soft upstream master`, which gave me a state where all my changes were uncomitted. Then, I  created a new branch based on `upstream/master`, and started committing files. Once that branch felt like "enough", new branch off current, continue.

The only real downside to this approach is that git history is lost, but if that makes it easier to review then I won't lose any sleep :shrug:

lua/neogit/lib/git.lua
ten3roberts, line 1: Don't we already have `merge` or is it just git showing the diff based off a BASE before I implemented that

lua/neogit/lib/git/merge.lua
```diff
@@ -1,52 +1,38 @@
...
-end
-
 function M.merge(branch, args)
   a.util.scheduler()
-  local git = require("neogit.lib.git")
-  local result = merge_command(git.cli.merge.args(branch).arg_list(args))
+  local result = merge_command(cli.merge.args(branch).arg_list(args))
   if result.code ~= 0 then
-    notif.create("Rebasing failed. Resolve conflicts before continuing", vim.log.levels.ERROR)
+    notif.create("Merging failed. Resolve conflicts before continuing", vim.log.levels.ERROR)
```
ten3roberts, line 20: Thank you.

You've caught me copying rebase

lua/neogit/lib/git.lua
ten3roberts: Seems to be there to allow `require"neogit.git".merge` rather than `require"neogit.git.merge" so that you only need a top `git` require?

ten3roberts (approved):
Seems like there are lots of robustness improvements and checks, which is fantastic as neogit crashing or doing the wrong thing could mean you lose work or miss a commit.

lua/neogit/lib/git.lua
CKolkey: Yup, precisely. I thought it was getting a little messy to have the individual git libs required in the popups, so a longer-term goal of mine is to update all the popups to only import `git` and navigate from there.

lua/neogit/lib/git/merge.lua
CKolkey: Haha, no worries ;)

CKolkey:
> That is clever and sounds like a good way to go about it. I am always on the lookup for new workflows.
> 
> Wouldn't this PR merge (master <- branch1) <- branch2 so that at the point of merging branch2 the diff of branch1 is already in master?

Yea, the workflow kinda broke because I don't have write access to this repo. branch2, 3, and 4 are on my fork. I'll have to think a little if there's a nice way to pull this off, or if we just have to go about this differently.

CKolkey:
Ahh, that's a good idea. Let's do that - lemme know when you make a branch and I'll update the base of this one to point there.

CKolkey:
Awesome, done. I'll have all the branches point to that one, then we can just go in order I guess :)

PR #501, opened 2023-05-11
Upstreaming 2/n - Update lib/*

CKolkey:
This expands on the non-git libs. Includes commits from https://github.com/TimUntersberger/neogit/pull/500, which should vanish once that's merged into the ckolkey branch :)


lua/neogit.lua
```diff
@@ -59,14 +59,14 @@ local neogit = {
     if opts ~= nil then
       config.values = vim.tbl_deep_extend("force", config.values, opts)
     end
-    if not config.values.disable_signs then
-      signs.setup()
-    end
+
```
ten3roberts, line 62: This unconditionally sets up signs

The README still mentions `disable_signs`

lua/neogit/lib/popup/lib.lua
ten3roberts, line 1: If I've understood main correctly these are actually defined in `lua/neogit/lib/popup.lua` and just duplicates here?

ten3roberts (changes requested):

lua/neogit.lua
CKolkey: The check was just moved into the `setup()` function: https://github.com/TimUntersberger/neogit/pull/501/files/cad85a236022389a127454f0a69c5d5fd125d6a8#diff-26497fce652f1fc707efdf4c7a8db4f3e6f96c3e7cd4c22c9b07c7515ac237b1R21

lua/neogit/lib/popup/lib.lua
CKolkey: Yea, I think this might have been a precursor to that? But it isn't used anymore :shrug:

CKolkey:
I think the spec failure here (and on the other branches) is because the commits are split up by file. You cool with deferring that until we are looking to merge to master?

ten3roberts:
I suspected so as well, though is seems to fail due to the removed popup lib

https://github.com/TimUntersberger/neogit/actions/runs/4947105276/jobs/8846043488?pr=501#step:4:256

CKolkey:
Yea, that line gets removed next branch 🤷

ten3roberts:
Perhaps we should just include that line/commit in this PR to have this pass.

That way we know that we don't get into a cascading state where something else fails but was stopped by an earlier failure.

ten3roberts:
Is this error going to be fixed by a later PR?

https://github.com/TimUntersberger/neogit/actions/runs/4951611519/jobs/8856815744?pr=501#step:4:3470

I am sorry if I am being pedantic

CKolkey:
> Is this error going to be fixed by a later PR?
> 
> [TimUntersberger/neogit/actions/runs/4951611519/jobs/8856815744?pr=501#step:4:3470](https://github.com/TimUntersberger/neogit/actions/runs/4951611519/jobs/8856815744?pr=501#step:4:3470)

To be honest, I'm not sure. Maybe a better way of putting it is that, because some of the user-interactions are changing from `vim.fn.input()` based to telescope, the tests themselves (the mocks) just kinda don't make sense anymore, in some cases. I'll try and figure that out tonight, but on the `HEAD` of my fork, not here.


> I am sorry if I am being pedantic
No worries at all :)

PR #502, opened 2023-05-11
Upstreaming 3/n - Popup/Builder libs

CKolkey:
I thought it would be easier for these to stand in their own PR.

`popup`
- Rename `option.parse` to `option.internal` (parse made no sense)
- Add `arg.cli_prefix` since some git flags do not use `--`. Some use `++`. Some use a single `-`. It's wild.
- Add `config` type to popups. This is, you guessed it, a `git config` setting the user can change.
- Switches can now be incompatible with other switches (defined in the popup creator) to prevent invalid input states
- `option`s can now support a `choices` table, to restrict the user to a set of valid inputs
- Adds a 'heading' value to allow for different categories of actions

`builder`
- Combine `self.state.switches` and `self.state.options` to `self.state.args` so we have better control over placement and ordering, and can set more detailed headings for sections. The existing layout style (options and switches) can be maintained, but it allow much greater flexibility.
- add `new_action_group_if` for conditional version
- Add `group_heading` and `group_heading_if` to set the heading for an action group
- Add switch opts:
  - internal (renamed from 'parse')
  - incompatible - prevent bad input state with a list of mutually exclusive switches
  - key_prefix - add the ability to overwrite the default hotkey prefix, `-` with something else
  - cli_prefix - Some switches (from the UI) are _not_ switches to the cli, so we need to account for that
  - value - allow setting a default value. Note, this is distinct from an option, where the value has a `=` separating it from the flag. Git log, for example, lets you send `-Gsomething` where `something` is user input.
- Add option opts:
  - key_prefix - same as switch
  - cli_prefix - same as switch
  - choices - limit user to prespecified choices (selected using `vim.ui.select`)
  - default - a default value
- arg_heading - Adds section heading for arguments
- config_heading - adds section heading for config

Here are some screenshots for what these changes allow:
`Log Popup`
Screenshot 2023-05-15 at 15 49 18

`Commit Popup`
Screenshot 2023-05-15 at 15 49 52

`Rebase Popup`
Screenshot 2023-05-15 at 15 49 58

`Pull Popup`
Screenshot 2023-05-15 at 15 50 06

`Branch Popup`
Screenshot 2023-05-15 at 15 58 15

`Branch Config Popup`
Screenshot 2023-05-15 at 16 23 30

### DIFF
--- a/lua/neogit/lib/popup.lua
+++ b/lua/neogit/lib/popup.lua
@@ -6,12 +6,19 @@ local logger = require("neogit.logger")
 local util = require("neogit.lib.util")
 local config = require("neogit.config")
 local state = require("neogit.lib.state")
+local input = require("neogit.lib.input")
+
+local branch = require("neogit.lib.git.branch")
+local config_lib = require("neogit.lib.git.config")
 
 local col = Ui.col
 local row = Ui.row
 local text = Ui.text
 local Component = Ui.Component
 local map = util.map
+local filter_map = util.filter_map
+local build_reverse_lookup = util.build_reverse_lookup
+local intersperse = util.intersperse
 local List = common.List
 local Grid = common.Grid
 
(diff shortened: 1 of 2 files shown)


CKolkey:
@ten3roberts lemme know if you have any questions.

lua/neogit/lib/popup/builder.lua
```diff
@@ -31,46 +33,163 @@ function M:env(x)
...
+    return self:group_heading(heading)
   end
 
-  table.insert(self.state.switches, {
-    id = "-" .. key,
+  return self
+end
+
+---@param opts.internal boolean Whether the switch is internal to neogit or should be included in the cli command.
+--                              If `true` we don't include it in the cli comand.
```
ten3roberts, line 63: Perhaps we should add incompatible here as well,

or in general maybe document the popup a little more?

ten3roberts (changes requested):

lua/neogit/lib/popup/builder.lua
CKolkey: I'm trying to just add things on an as-needed basis (for the most part, at least). Haven't found a place where two options need to be exclusive yet.

I can add more documentation, thats a good idea.

lua/neogit/lib/popup.lua
```diff
@@ -81,97 +89,280 @@ local function get_highlight_for_option(option)
...
+      for _ = 1, #value do
+        table.remove(component.children)
+      end
+
+      -- insert new items to row
+      for _, text in ipairs(value) do
+        table.insert(component.children, text)
+      end
+    else
+      print("Unhandled component value type! (" .. type(value) .. ")")
```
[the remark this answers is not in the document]
CKolkey: Makes sense.

ten3roberts:
Do you feel that this is ready to merge, or do you want to add more things?

CKolkey:
Im happy with this if you are - onwards and upwards, as they say

PR #503, opened 2023-05-11
Upstreaming 4/n - Add (optional) Telescope dependency and `vim.ui.select` fuzzy finder

CKolkey:


CKolkey:
Hmm. Yea, I guess it comes down to "out of the box experience" vs. "ecosystem". I think right now I've leaned towards the former, but then again I'm also using telescope so it wasn't really a big shift. I think it could work fine to let the `FuzzyFinderBuffer` class decide whether or not to use telescope based on a config setting (like the diffview integration), as long as the public api is the same for both. You mentioned something in that issue about `a.wrap()`ing the function so it could return the value without a callback? Maybe you could give me a hand with the implementation of that, if you like.

I have some commits to cherry pick over here that add user-configurability to the key mappings, so it will (the telescope finder) at the very least be completely customizable.

lua/neogit/lib/finder.lua
```diff
@@ -0,0 +1,108 @@
...
+  return self
+end
+
+---Adds a select action - NOT OPTIONAL
+---@param action function
+---@return Finder
+function Finder:add_select_action(action)
+  self.select_action = action
+  return self
+end
```
CKolkey, line 88: These should probably just get added via the constructor.

ten3roberts:
> Hmm. Yea, I guess it comes down to "out of the box experience" vs. "ecosystem". I think right now I've leaned towards the former, but then again I'm also using telescope so it wasn't really a big shift. I think it could work fine to let the `FuzzyFinderBuffer` class decide whether or not to use telescope based on a config setting (like the diffview integration), as long as the public api is the same for both. You mentioned something in that issue about `a.wrap()`ing the function so it could return the value without a callback? Maybe you could give me a hand with the implementation of that, if you like.
> 
> I have some commits to cherry pick over here that add user-configurability to the key mappings, so it will (the telescope finder) at the very least be completely customizable.

I hope you don't mind it; I pushed a small commit which shows the usage of vim.ui.select in the `:find` method of the Finder, as well as making telescope an optional dependency. I am not sure I completely understand where you use telescope in an async manner, as the `:find` method just seems to finish immediately and then later call the provided `select_action` function at some future point in time. I added  a small `test` to the file as well so you can try it for yourself.

For the record, here is what it looks like for me:

Screenshot 2023-05-24 at 16 24 09

And after pressing enter:

Screenshot 2023-05-24 at 16 24 43

CKolkey:
Yea, no worries at all. This class hasn't been used anywhere yet, that'll be over the next few pr's, so it would be reasonable (i think) to defer some of the finer details

ten3roberts:
Do you want me to annotate functions and create the `open_async` I talked about?

CKolkey:
Sure! Once this is merged we can get into the individual popups, where I think the implementation will make sense/be testable

ten3roberts:
Did you also get a `attempt to concatenate nil value column_width` for e.g PushPopup?

Anyhow, that error is fixed and merge in as well

ten3roberts:
There is also a new fancy async api for the buffers, which don't require the user to supply their function in the constructor which is then not used. Check it out

CKolkey:
> Did you also get a `attempt to concatenate nil value column_width` for e.g PushPopup?
> 
> Anyhow, that error is fixed and merge in as well

Saw the fix - seems reasonable :) I think I may have seen that issue once or twice, but it's not something I've encountered enough to ever dig into it.

Not sure which async API you mean - maybe I'm just being blind right now? 😅

ten3roberts:
No worries

At the bottom of the `fuzzy_finder.lua` there is an `open_async` function which returns the value directly and does not use a callback.

It is a similar behavior to the channel based `open_sync` you made. Just made a few apis available to fit the best use case and to see what fits.

ten3roberts:
Overall I am quite happy with the state of this PR and the optional telescope dependency.

Let me know if we should merge this and move on to the next.

Thanks for your awesome contributions :heart:

ten3roberts:
Pardon the inconvenience.

You are not blind, I did `Pp` and pushed to `TimUntersberger/neogit/merge-upstream-4` and created a new branch rather than the other remote that I used to checkout your branched, I.e pushed to something different than pull.

Now the changes are on this PR

You think me of all people should know how to push using neogit.

I don't quite recall, but did your upstream changes show what push will push to? This has caused me to push to the wrong remote more than a couple of times.

lua/neogit/buffers/fuzzy_finder.lua
```diff
@@ -0,0 +1,80 @@
...
+  self:open(...)
+
+  rx()
+  return result
+end
+
+---@param opts FinderOpts
+---@return any|nil
+--- Asynchronously prompt the user for the selection, and return the selected item or nil if aborted.
+function M:open_async(opts)
```
ten3roberts, line 61: This is the suggested impl for the comment above

lua/neogit/lib/finder.lua
```diff
@@ -0,0 +1,178 @@
+local config = require("neogit.config")
+local async = require("plenary.async")
+
+local function telescope_mappings(on_select, allow_multi)
+  local action_state = require("telescope.actions.state")
```
ten3roberts, line 5: Imports are placed here to only import telescope if enabled

lua/neogit/lib/notification.lua
```diff
@@ -9,7 +9,7 @@ local function create(message, level, delay)
   end
 
   if config.values.disable_builtin_notifications then
-    vim.notify(message, level, { title = "Neogit", icon = "" })
+    vim.notify(message, level, { title = "Neogit", icon = "󰊢" })
```
ten3roberts, line 12: This will disappear once CKolKey branch is merged into master

lua/neogit/lib/finder.lua
```diff
@@ -0,0 +1,178 @@
...
+    table.insert(self.entries, entry)
+  end
+  return self
+end
+
+---Engages finder and invokes `on_select` with the item or items, or nil if aborted
+---@param on_select fun(item: any|nil)
+function Finder:find(on_select)
+  if config.values.use_telescope then
+    local pickers = require("telescope.pickers")
```
ten3roberts, line 126: Same here, only import in the telescope enabled branch rather than the top of the file

lua/neogit/lib/finder.lua
```diff
@@ -0,0 +1,178 @@
+local config = require("neogit.config")
+local async = require("plenary.async")
+
+local function telescope_mappings(on_select, allow_multi)
+  local action_state = require("telescope.actions.state")
+  local actions = require("telescope.actions")
+  --- Lift the picker select action to a item select action
+  local function select_action(prompt_bufnr)
```
ten3roberts, line 8: I moved this from the `fuzzy_finder.lua` since it is only needed for telescope, the `vim.ui.select` calls the select function with the item directly

lua/neogit/buffers/fuzzy_finder.lua
```diff
@@ -0,0 +1,80 @@
...
+    allow_multi = false,
+    layout_config = { height = buffer_height(#self.list) },
+  }
+
+  Finder.create(opts):add_entries(self.list):find(action)
+end
+
+-- Opens finder in such a way that selected value can be returned to the main thread
+-- without the need to use a callback to process the selection.
+function M:open_sync(...)
```
ten3roberts, line 43: You may have other opinions on this, but from what I've seen `sync` refers to the whole editor blocking on the call, like a prompt using the older style vimscript function like `confirm()` or other ex line prompts. Libuv uses this notation for example, where `sync` methods just freeze the whole editor in place until it is done, which is generally not preferred.

Whereas `async` only suspends the current task and keeps all other things turning.

The plenary async tasks `async.run` `async.void` and everything run inside of them spawn a lua coroutine and when awaiting for a result the coroutine is suspended, and resumed when a value is available, which is what neogit uses as much as possible.

The third method is callbacks, which is the style that vim.loop (https://github.com/luvit/luv) uses.

These can be adapted to the coroutine/task/green thread style using `plenary.async.wrap`, which is how `find_async` uses vim.ui.select which is callback style.

I do not know which language you are the most familiar to, but the plenary async model is quite similar to go's async model.

Let me know if you have any opinions on the naming

ten3roberts:
Hey Cameron, what do you think of the api I added? Is it inline with what you thought or should we remove it? I don't want to go intrude too much on *your* work

CKolkey:
Hey! Haven't had a chance to look at this over the weekend - it was a holiday here. I'll respond to the individual comments you left, but generally I've got no issue. And, honestly, until we start _consuming_ the API's it's a bit hard to say. If it doesn't work down the road, then we just change it again, simple as that ^_^

lua/neogit/buffers/fuzzy_finder.lua
CKolkey: I think I just had a misunderstanding of how sync/async/blocking nomenclature worked. I mostly write ruby at the dayjob, so the only real concurrency I've dealt with is... pretty much none, actually. Which is also to say that I appreciate the pointers ;) 

So, if it makes more sense to rename the method, then I'm all for it.

lua/neogit/buffers/fuzzy_finder.lua
```diff
@@ -0,0 +1,88 @@
...
+    end
+  end
+
+  Finder.create(opts):add_entries(self.list):add_select_action(select_action):find()
+end
+
+-- Opens finder in such a way that selected value can be returned to the main thread
+-- without the need to use a callback to process the selection.
+function M:open_sync(...)
+  local tx, rx = require("plenary.async.control").channel.oneshot()
```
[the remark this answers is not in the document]
CKolkey: Yea, we can give that a shot. I _think_ I tried doing something like that while trying to get this working.. I tried tons of different things, really. Saw a lot of "attempt to yield across c-function call boundary" errors. But if this can work more simply, then I'm all for it.

asmodeus812:
@CKolkey i second that telescope should be optional, also i was trying out your fork but telescope is required there, is there any chance you could upstream this change to your fork as well ?

CKolkey:
> @CKolkey i second that telescope should be optional, also i was trying out your fork but telescope is required there, is there any chance you could upstream this change to your fork as well ?

Upstream the change downstream? You're a madman 🤣 

...but yeah, I think I'll wind up doing that since we changed some API's here, just to make sure everything is working/renamed correctly.

asmodeus812:
Yeah :D, I mean at this point your fork seems much more complete, and is getting quite good !

ten3roberts:
Yes.

We will hope and see if @TimUntersberger will come back soon.

I think the best thing would be to make this an organization so that we can more easily add contributors as this is quite a large plugin.

I use Neogit every day and maintain where I can, and I am able to manage PRs and issues, but I lack the ability to change repository settings such as contributors

CKolkey:
Agreed :) - Anything else you want to see here @ten3roberts before merging?

ten3roberts:
No it all looks great, and as you said we will be able to refine it when we do use it, as well as after the fact as I have a habit of polishing neovim plugins while compiling other things

CKolkey:
Haha, the joy of compiled languages. Lets merge this, then knock out the next few (should be easier ones) - then get to the real fun :)

PR #504, opened 2023-05-11
Upstreaming 5/n - Cleanup syntax files

CKolkey:
This is just a general cleanup of the syntax files. A lot of the sign definitions have been extracted to a lua file, as well as highlight definitions. Makes it easier to do dynamic highlights/signs and prevent issues for users who have syntax disabled.


CKolkey:
The failing spec is just because of the same issues that are corrected in the base branch.

ten3roberts:
Is this related to the sections like HEAD: {...} and headlines like `Recent commits` having a different color?

ten3roberts:
So if I am understanding this all right this just removes now unnused syntax files in favor of the already merged `nvim_sign_define` `nvim_set_hl` etc.

How related are the upstreaming changes to

https://github.com/TimUntersberger/neogit/pull/415?
and
https://github.com/TimUntersberger/neogit/issues/405?

CKolkey:
You are 100% right, and not really related at all.

The next PR in the stack addresses filetypes, and I came to the conclusion we can have the best of both worlds: custom filetypes for neogit buffers, and treesitter parsers turned on (if available).

ten3roberts:
Agreed, it is advantageous to make use of treesitter rather than our own hand rolled regexes

PR #505, opened 2023-05-11
Upstreaming 6/n - Update FTPlugin

CKolkey:
- Use treesitter parsers if available
- Start in insert-mode if the line is empty
- Start at the top line always


CKolkey:
@ten3roberts Lemme know if you have any questions.

ftplugin/NeogitCommitMessage.lua
```diff
@@ -0,0 +1,16 @@
+if vim.b.did_ftplugin then
+  return
+end
+
+vim.cmd.source("$VIMRUNTIME/ftplugin/gitcommit.vim")
+vim.cmd([[exec 'norm! gg']])
```
ten3roberts, line 6: In which situations it the cursor not at the top?

ftplugin/NeogitCommitMessage.lua
```diff
@@ -0,0 +1,16 @@
...
+
+if vim.fn.prevnonblank(".") ~= vim.fn.line(".") then
+  vim.cmd([[startinsert]])
```
ten3roberts, line 9: Does this respect the config option `disable_insert_on_commit`?

ftplugin/NeogitRebaseTodo.vim
ten3roberts, line 1: Great, I also found the rebase todo to be supreflous when adding that feature

ten3roberts (changes requested):

ftplugin/NeogitCommitMessage.lua
CKolkey: Probably none for most people, but if you have a "remember cursor location in buffer" autocmd/plugin, this would attempt to defeat that.

ftplugin/NeogitCommitMessage.lua
CKolkey: No.. Probably should. Or just be sure to call `stopinsert` if it's set to true in the buffer constructor.

ten3roberts: startinsert is already called in [`commit_](https://github.com/TimUntersberger/neogit/blob/c8af6d8429ba5b4bd5686084bf70d0d1483d0a18/lua/neogit/buffers/commit_editor/init.lua#L77)

So I do not think it is needed here

ten3roberts: I do not think `stopinsert` is the right either as we will have called insert enter autocommands, for a short while. Those could then have `vim.schedule`d something which will break in the next tick as we are no longer in insert mode.

I think it will cause more harm than good when considering that insert modes changes tend to lead to many interactions with other plugins. I know of quite a few other plugins who have had odd interactions due to there functions.

CKolkey: Hmm. Alright, great points. Maybe a better course would be to add this as a config option, ie, `smart_insert` and move it into the buffer constructor.

ten3roberts: It is already there in the constructor :)

On another note: I'd really prefer if insert_on_commit is left as optional since I prefer to not enter insert mode automatically

CKolkey: Totally fair - just haven't had time to fix up this PR yet. I'll try to get that all working tonight :)

ftplugin/NeogitCommitMessage.lua
CKolkey: Ditched this. Probably not needed after all.

ten3roberts (approved):

ten3roberts:
Looks great, thanks for you contribution

Issue #495, opened 2023-05-02
Initialize git repository


ten3roberts:
It would be nice for neogit to be able to initialize a repository if one doesn't exist (or clone, similar to https://magit.vc/manual/2.90.0/magit/Repository-Setup.html).

_Originally posted by @CKolkey in https://github.com/TimUntersberger/neogit/issues/487#issuecomment-1517999648_

milkias17:
I'd really like to work on this. What would be an acceptable pr? How did you imagine getting user input for initializing a repo, I thought we can just prompt a user, would you like to initialize a git repo if they tried to open the neogit window inside a non-git repo. In what form should we take their input, would a simple vim.fn.input suffice?

CKolkey:
Sure! You can use https://magit.vc/manual/magit/Creating-Repository.html as a reference. Essentially you'll be making a function bound to `I` on the status buffer. Heres where thats set up: https://github.com/TimUntersberger/neogit/blob/master/lua/neogit/status.lua#L879

If you follow how emacs magit works, pressing `I` will prompt the user to pick a folder to initialize a repo in. 

Finally, here's a reference for the implementation of such a function: https://github.com/magit/magit/blob/9c66514b2c879a8959b676393467b1f24be71303/lisp/magit-status.el#L205

For now, `vim.fn.input` should be fine.

milkias17:
Ok great thanks a lot for the reply. I'll try my best to achieve this👍

CKolkey:
Sounds good, and feel free to ping me with questions :)

PR #511, opened 2023-05-16
New repo

milkias17:
Implement feature #495

- Now, if the user invokes neogit inside a non-git repository, neogit asks the user if they want to initialize a new repository, if they oblige it will create and open neogit.
- Now neogit also has an I command inside the status buffer which allows the user to initialize a repository inside a directory they want.


CKolkey:
Hey! Great start. Check out the `lib/git/` folder. I think it would make sense to have and api like `git.init.create("name")`, so you could use that lib/function in both locations.

Also check out how the `git.cli` lib provides a declarative interface for the git cli. That'll let you use the process system, which does error handling and will make sure to add the command to the log view.

Also, check out the `help` popup - the `I` entry belongs there too

milkias17:
Hello. Thanks a lot for the feedback. I've tried to implement the 3 things you mentioned:

- Not there is a `git.init.create` function inside `lib/git` folder.
- I've stopped using vim.fn.system calls and used `git.cli` lib to run the `git init` command.
- I've also inserted an entry into the help popup for I.

Please do let me know if there's still more things I need to address or if there's something I overlooked from your previous comment.

lua/neogit.lua
```diff
@@ -26,8 +26,18 @@ local neogit = {
     opts = opts or {}
 
     if not cli.git_is_repository_sync() then
-      notification.create("The current working directory is not a git repository", vim.log.levels.ERROR)
-      return
+      local create = vim.fn.input({
```
CKolkey, line 29: This could use `vim.fn.confirm()` - there are some other examples of that in the codebase. Or, better yet, `require("neogit.lib.input").get_confirmation(...)`

lua/neogit/status.lua
```diff
@@ -859,6 +859,46 @@ local discard = function()
   vim.cmd("checktime")
 end
 
+local init_repo = function()
```
CKolkey, line 862: I think this would fit better in `neogit.lib.git.init` so that the status buffer isn't exporting a git operation

CKolkey (changes requested):
Minor stuff - just move the git operation fn, and use the input lib's get_confirmation, then I think it's looking good :)

milkias17:
@ten3roberts Any comments on this to get this merged in?

lua/neogit/lib/git/init.lua
```diff
@@ -0,0 +1,58 @@
...
+    text = "",
+    cancelreturn = "",
+    completion = "dir",
+  }
+  if directory == "" then
+    return
+  end
+
+  -- git init doesn't understand ~
+  if string.find(directory, "~") == 1 then
```
ten3roberts, line 29: Perhaps we should use vim.fn.fnamemodify(directory, ":p") instead to resolve all of this

ten3roberts (changes requested):
Looks like a great addition.

I left a single concern, and the linting would also need to be fixed, but we can merge this in.

Thanks for the help

lua/neogit/lib/git/init.lua
milkias17: yeah that's much better.. I've changed that in the latest commit.

ten3roberts:
Great.

Do you think we could fix the recent commit parsing error popup in thist PR, or will that be much too large?
Screenshot 2023-05-22 at 16 55 05

milkias17:
Yeah I can try to fix it. I'll see what I can do.

milkias17:
@ten3roberts The solution i thought of was to add a suppress_error option inside the process class, so whenever we spawn a process, we can pass that flag. Then based on the flag to not show errors. Would that be an acceptable solution or?

If so, are there cases where refreshing the status buffer should throw errors?

ten3roberts:
> @ten3roberts The solution i thought of was to add a suppress_error option inside the process class, so whenever we spawn a process, we can pass that flag. Then based on the flag to not show errors. Would that be an acceptable solution or?
> 
> If so, are there cases where refreshing the status buffer should throw errors?

There is a field in process options called `on_error` which if true or a function which returns true, will handle the error and prevent it from being printed.

I seem to have removed that accidentally at some point, likely due to the rework that allowed output of the command to be captured and logged, which meant I didn't have to intercept to error to show that rebasing failed due to conflicts,

...anyhow, it is now back in master, so you can just supply your own function for logging.

I think the other existing functions should still show errors to not give a false positive that the buffer was properly refreshed. If there in the future shows to be another error that is obvious given the context and we can silence that as well.

ten3roberts:
The commit of interest

https://github.com/TimUntersberger/neogit/commit/bcf4b6bb203d89af49ee22b64a06d5571bd3f39c

milkias17:
@ten3roberts I've made it so that neogit doesn't show a popup when `repo.update_recent` is called as that seems to be where this emanates from.

ten3roberts (approved):

ten3roberts:
Fantastic, let me know if you are ready to merge :)

milkias17:
Yeah  I mean if there's nothing you think should be improved or changed I think it's ready.

PR #516, opened 2023-05-29
Use default vim notify icons when disable_builtin_notifications are true

Fabian-programmer:
Not every font has a nerd font 3.0 patch yet. I suggest to use default icons.

This is how it looks with nvim-notify:
grafik

Without notify visualization like noice or nvim-notify the icon isn't shown anyway? At least i didn't find it.

@aleprovencio what do you think?

### DIFF
--- a/lua/neogit/lib/notification.lua
+++ b/lua/neogit/lib/notification.lua
@@ -9,7 +9,7 @@ local function create(message, level, delay)
   end
 
   if config.values.disable_builtin_notifications then
-    vim.notify(message, level, { title = "Neogit", icon = "󰊢" })
+    vim.notify(message, level, { title = "Neogit" })
     return nil
   end
 


CKolkey:
Just as an aside, the nerdfont 3.0 transition is such a mess 😂 Like, I completely get the reasoning, and I'm not saying it could have been handled -better-, but given the nature of fonts, and no programatic way to determine if a user has updated... not to mention, how many people even realized they needed to update a font? 

I'm not envious.

aleprovencio:
I get it guys, the main reason I updated the icon was because the old one was not in the range anymore so that ugly square was being generated.

I'm totally OK with this change considering your compatibility issues raised here.

CKolkey:
Makes sense. I think the compromise would be to expose the icon as a user configurable option, and provide a default (imo the 3.0 version)

PR #531, opened 2023-06-18
Merge Upstream - 7

CKolkey:
Ok @ten3roberts , think I got it now.

### DIFF
--- a/lua/neogit/lib/finder.lua
+++ b/lua/neogit/lib/finder.lua
@@ -1,9 +1,10 @@
 local config = require("neogit.config")
-local async = require("plenary.async")
+local a = require("plenary.async")
 
 local function telescope_mappings(on_select, allow_multi)
   local action_state = require("telescope.actions.state")
   local actions = require("telescope.actions")
+
   --- Lift the picker select action to a item select action
   local function select_action(prompt_bufnr)
     local selection = {}
@@ -30,7 +31,6 @@ local function telescope_mappings(on_select, allow_multi)
     end
   end
 
-  local actions = require("telescope.actions")
   return function(_, map)
     local commands = {
       ["Select"] = select_action,
@@ -150,7 +150,7 @@ end
 
 ---@type async fun(self: Finder): any|nil
 --- Asynchronously prompt the user for the selection, and return the selected item or nil if aborted.
-Finder.find_async = async.wrap(Finder.find, 2)
+Finder.find_async = a.wrap(Finder.find, 2)
 
 ---Builds Finder instance
 ---@param opts table|nil
@@ -161,7 +161,7 @@ end
 
 --- Example usage
 function Finder.test()
-  async.run(function()
+  a.run(function()
     local f = Finder:create()
     f:add_entries { "a", "b", "c" }
 
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -284,7 +284,19 @@ local configurations = {
     },
   },
 
-  fetch = config {},
+  fetch = config {
+    options = {
+      recurse_submodules = "--recurse-submodules",
+      verbose = "--verbose",
+    },
+    aliases = {
+      jobs = function(tbl)
+        return function(n)
+          return tbl.args("--jobs=" .. tostring(n))
+        end
+      end,
+    },
+  },
 
   ["read-tree"] = config {
     flags = {
@@ -388,6 +400,16 @@ local configurations = {
       deduplicate = "--deduplicate",
       exclude_standard = "--exclude-standard",
       full_name = "--full-name",
+    },
+  },
+
+  ["ls-remote"] = config {
+    aliases = {
+      remote = function(tbl)
+        return function(remote)
+          return tbl.args(remote)
+        end
+      end,
     },
   },
 
@@ -677,12 +699,13 @@ local function new_builder(subcommand)
     env = {},
   }
 
-  local function to_process(verbose, suppress_error)
-    -- Disable the pager so that the commands don't stop and wait for pagination
-    local cmd = { "git", "--no-pager", "-c", "color.ui=always", "--no-optional-locks", subcommand }
+  local function to_process(verbose, suppress_error, ignore_code)
+    local cmd = {}
+
     for _, o in ipairs(state.options) do
       table.insert(cmd, o)
     end
+
     for _, arg in ipairs(state.arguments) do
       if arg ~= "" then
         table.insert(cmd, arg)
@@ -701,6 +724,9 @@ local function new_builder(subcommand)
       table.insert(cmd, 1, state.prefix)
     end
 
+    -- Disable the pager so that the commands don't stop and wait for pagination
+    cmd = util.merge({ "git", "--no-pager", "-c", "color.ui=always", "--no-optional-locks", subcommand }, cmd)
+
     logger.debug(string.format("[CLI]: Executing '%s %s'", subcommand, table.concat(cmd, " ")))
 
     return process.new {
@@ -709,6 +735,7 @@ local function new_builder(subcommand)
       env = state.env,
       pty = state.in_pty,
       verbose = verbose,
+      ignore_code = ignore_code,
       on_error = suppress_error,
     }
   end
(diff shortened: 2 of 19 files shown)


lua/neogit/lib/finder.lua
```diff
@@ -1,9 +1,10 @@
 local config = require("neogit.config")
-local async = require("plenary.async")
```
CKolkey, line 2: I changed these back to `a` to be consistent with the rest of the project.

lua/neogit/lib/git/cli.lua
```diff
@@ -701,6 +724,9 @@ local function new_builder(subcommand)
       table.insert(cmd, 1, state.prefix)
     end
 
+    -- Disable the pager so that the commands don't stop and wait for pagination
+    cmd = util.merge({ "git", "--no-pager", "-c", "color.ui=always", "--no-optional-locks", subcommand }, cmd)
```
CKolkey, line 728: This got moved down to fix a bug: `prefix` would insert at index 1, which was before the string `git`.

lua/neogit/lib/git/repository.lua
```diff
@@ -1,16 +1,8 @@
...
-
-local meta = {
-  __index = {},
-}
-
-local modules = { "status", "diff", "stash", "pull", "push", "log", "rebase", "cherry_pick", "merge" }
-for _, m in ipairs(modules) do
-  require("neogit.lib.git." .. m).register(meta.__index)
-end
+local a = require("plenary.async")
```
CKolkey, line 1: This module was rewritten to act as a singleton, instead of being bound as an instance on the status module. This resulted in (imo) better encapsulation, and let us remove many random imports of the status buffer in different libs just to get a stateful repo. 

It also meant I could extract the state-updating logic from the status buffer, which is a huge win, since that guy needs to get slimmed down and reimplemented using the buffer lib at some point.

ten3roberts: Yes I saw that and thought the same.

Also neat to move the refresh logic to the repository module.

lua/neogit/lib/finder.lua
ten3roberts: The issue with `a` is that is can be confused with `local a = vim.api` so the names clash when moving code around.

I think being slighly longer with `async` and `api` to not confuse them in shortening is a good idea.

lua/neogit/lib/git/repository.lua
CKolkey: I thought it made a lot more sense to say "Hey repo, refresh yourself" vs "Hey repo... I'm refreshing you"

lua/neogit/lib/finder.lua
CKolkey: Totally fair - in that case, why don't we do a project-wide update after all this is merged?

lua/neogit/lib/git/push.lua
```diff
@@ -14,7 +14,7 @@ function M.push_interactive(remote, branch, args)
 end
 
 local function update_unmerged(state)
-  if not state.upstream.branch then
+  if not state.upstream.ref then
```
ten3roberts, line 17: Why the general change from`branch` to `ref`?

Is it due to a collision with something else called `branch` or just to be consistent with git's naming?

lua/neogit/lib/git/repository.lua
```diff
@@ -45,20 +38,148 @@ M.create = function(_path)
...
+    for name, fn in pairs(self.lib) do
+      table.insert(refreshes, function()
+        logger.debug("[REPO]: Refreshing " .. name)
+        fn(self.state)
+      end)
+    end
+  end
+
+  logger.debug(string.format("[REPO]: Running %d refresh(es)", #refreshes))
+  a.util.join(refreshes)
```
ten3roberts, line 157: I have a wip timeout of individual refresh jobs, in case on hangs or doesn't complete.

Will rebase that ontop of this later.

I noticed that `merge.abort` can run indefinitely due to a race, that is why 😊

lua/neogit/lib/git/stash.lua
```diff
@@ -65,54 +59,67 @@ local function perform_stash(include)
...
+
+function M.stash_all(args)
+  cli.stash.arg_list(args).call()
+  -- this should work, but for some reason doesn't.
+  --return perform_stash({ worktree = true, index = true })
 end
 
-local function list()
-  return cli.reflog.show.format("%h").args("stash").call():trim().stdout
+function M.stash_index()
```
ten3roberts, line 77: I also prefer the `M.` syntax rather than the exports table at the end.

Makes *goto definition* jump to it immediately rather than landing on the field of the exports table.

Good work

CKolkey: Exactly!

lua/neogit/lib/git/cli.lua
```diff
@@ -709,6 +735,7 @@ local function new_builder(subcommand)
       env = state.env,
       pty = state.in_pty,
       verbose = verbose,
+      ignore_code = ignore_code,
```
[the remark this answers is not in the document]
CKolkey: Oh, thats not a duplicate. I wound up implementing my own version of `on_error` because I didn't know there was an `on_error`. But `call_sync_ignoring_exit_code` and `call_ignoring_exit_code` (and the callers) need to be updated to use `external_errors` properly.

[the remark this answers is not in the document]
ten3roberts: Yes, not a duplicate per say, but a duplicate in functionality.

If you want to we can just leave it as is for now and I can merge those implementations together after merging everything, so that it is less work for you and you can keep focus on what's important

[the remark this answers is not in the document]
CKolkey: Agreed, its due for a cleanup, and yea I'd like to just get this all merged and come back 😅 Even whats in PR's is already behind.. i made the revert popup two days ago, and added a force delete option for deleting unmerged branches last night...

lua/neogit/lib/git/repository.lua
CKolkey: Ah, very cool idea. Want to just do that via PR once this is merged to CKolkey? Rebasing means I have to fix the histories of all the subsequent branches 😟

lua/neogit/lib/git/push.lua
CKolkey: Partly to be more consistent with git nomenclature, and partly because I'm trying to use more internal state for things instead of calling out to the CLI. Some views make like... 5 cli calls that are exactly the same, to get stuff like branch names, or upstream names. So `upstream.ref` can be the unparsed version, and `upstream.branch` can be just the name.

ten3roberts: Oh in that case, fantastic.

Perhaps a comment that clears up their relationship (if there isn't one already) would be a good idea.

lua/neogit/lib/git/repository.lua
ten3roberts: Ah no, I meant that I rebase my (yet to be commited on a suspended machine) changes on the ckolkey branch once all is wrapped up.

Opening a PR to ckolkey now would be a good idea, unless it makes the next few PRs conflict with their merges.

lua/neogit/lib/git/push.lua
CKolkey: Yeah, I'll add documentation in the repository that lays out what everything is. Good call.

CKolkey:
> Feel free to add me as a reviewer again once you are ready.
> 
> 
> 
> You don't have to worry about requesting too many reviews 😊 

Ah, super, good to know. This PR is ready in my eyes - I'll make sure CI passes before we merge into master anyhow. Easier that way.

lua/neogit/lib/git/repository.lua
CKolkey: Hmm. Maybe. But if you open it then I can at least have a look at what you're up to and plan accordingly :)

ten3roberts:
Nice. Should we merge this now and move on to the next set of sweet features?

CKolkey:
Lets do it :D

PR #532, opened 2023-06-18
Upstreaming -- 8 Popups

CKolkey:
Popups!

The big change here is to extract the inline functions from the builder and put them into their own module. This is for two reasons:
1. It's cleaner to look at.
2. This makes the actions testable without needing the builder! Since now they're just functions.


MaxKiv:
༼ つ ◕_◕ ༽つ @CKolkey Cameron take my energy ༼ つ ◕_◕ ༽つ

CKolkey:
@ten3roberts I can't request a review formally, but I can do it as a comment, informally ;)

lua/neogit/config.lua
```diff
@@ -109,8 +109,11 @@ M.values = {
       ["c"] = "CommitPopup",
       ["L"] = "LogPopup",
       ["Z"] = "StashPopup",
+      ["A"] = "CherryPickPopup",
```
ten3roberts, line 112: It seems that these mappings are not set up in `neogit/lua/status.lua`

They are present on your master branch, just not in this branch

lua/neogit/popups/rebase/actions.lua
```diff
@@ -0,0 +1,71 @@
...
+
+function M.onto_base(popup)
+  git.rebase.rebase_onto(M.base_branch(), popup:get_arguments())
+  a.util.scheduler()
+  status.refresh(true, "rebase_master")
+end
+
+-- TODO: Set upstream if unset
+function M.onto_upstream(popup)
+  git.rebase.rebase_onto(git.repo.upstream.remote .. " " .. git.repo.upstream.branch, popup:get_arguments())
```
ten3roberts, line 23: Won't this be treated as one argument `"upstream branch"` rather than `"upstream", "branch"`?

Running rebase locally requires those to be two separate shell arguments, but there may be something I've missed

doc/neogit.txt
```diff
@@ -100,8 +100,18 @@ p                       Open pull popup
 
                                                 *neogit_P*
 P                       Open push popup
+
                                                 *neogit_f*
 f                       Open fetch popup
+
```
ten3roberts, line 106: Fantastic job on keeping the documentation up to date :+1: 

It is usually the part most forgotten about

lua/neogit/popups/cherry_pick/init.lua
```diff
@@ -0,0 +1,40 @@
...
+    -- :option("s", "strategy", "", "Strategy")
+    -- :option("S", "gpg-sign", "", "Sign using gpg")
+    :group_heading_if(not in_progress, "Apply here")
+    :action_if(not in_progress, "A", "Pick", actions.pick)
+    :action_if(not in_progress, "a", "Apply", actions.apply)
+    :action_if(not in_progress, "h", "Harvest")
+    :action_if(not in_progress, "m", "Squash")
+    :new_action_group_if(not in_progress, "Apply elsewhere")
+    :action_if(not in_progress, "d", "Donate")
+    :action_if(not in_progress, "n", "Spinout")
```
ten3roberts, line 26: I suppose these will be filled in at a later point in time

ten3roberts (changes requested):
I left some small comments for things that caught my eye.

It seems as if most of the diff in this branch is just splitting popups into an `init` and `action` pair, which in my opinon makes it a lot easier to see what the popup can do, keybindings etc.

Is there any change to *existing* popups, rather than functions having moved? It is difficult to diff when it is all green and moved about.

Overall, great job. I am looking forwards to seeing this in master

lua/neogit/popups/cherry_pick/init.lua
ten3roberts: Reference to nomenclature: https://magit.vc/manual/magit/Cherry-Picking.html#index-A-h

CKolkey: Yeah.. i've tried to make sense of how these are implemented in Magit and.. I'm putting it off for now. Maybe someone beats me to it 😅 But for the sake of previewing whats to come, and layout, I added them anyways.

lua/neogit/popups/rebase/actions.lua
CKolkey: I think it was so I didnt need to change the function signature of `rebase_onto`, since you can also use it without a remote. Maybe its a better idea to pass an opts table with a remote key, and use that if present

ten3roberts: I think an table is the best here for versitility.

Does the rebase command work at all now with the "combined" argument?

lua/neogit/config.lua
CKolkey: Yea, the status buffer is in an upcoming branch. I didn't wanna co-mingle things in this PR, since it's already a lot.

lua/neogit/popups/rebase/actions.lua
CKolkey: After a bit of reading, it seems like git will happily accept the fully-qualified ref name, f.ex `refs/remotes/origin/master` in place of `origin master`. I think that's the simplest way to go about this.

Not sure what you mean by "combined" - could you elaborate?

CKolkey:
> I left some small comments for things that caught my eye.
> 
> Most of the diff in this branch is just splitting popups into an `init` and `action` pair, which both you and I agree makes it a lot easier to see what the popup can do, keybindings etc.
> 
> Is there any change to _existing_ popups, rather than functions having moved? It is difficult to diff when it is all green and moved about.

Yeah, thats fair. I definitely have fixed up stuff... But I've done a lot, so it's hard to recall all the details. The things that got changed were to match the behaviour of Magit. For example, on the branch popup, "delete remote" and "delete local" just becomes "delete" and lists both local and remote refs, and then has some logic added about unmerged commits.

You can check out this commit: https://github.com/CKolkey/neogit/blame/e80b0d1e251507b737be1a6aefe9810cef741f48/lua/neogit/popups/branch.lua

and see the blame/history for each popup file up to there. Alternatively, you can take my word that any change made was minor, or a fix in behaviour, to more closely match Magit ;)


> Overall, great job. I am looking forwards to seeing this in master

Thanks :D Me too!

lua/neogit/popups/rebase/actions.lua
ten3roberts: That sounds better, so `branch + "/" + commit` rather than `branch + " " + commit`

Testing the existing state locally with `ru` gives

image

I'll make a note about it and you or I will come around to fix it later, don't put any pressure on it now

ten3roberts (approved):

PR #533, opened 2023-06-18
Upstreaming  9 - Buffers

CKolkey:
### Branch Select View
- Removed. Now handled by either telescope or `vim.ui.select`

### Commit Select View
- Changed UI to `tab` to better match Magit
- Adds common bindings (esc, q)
- Replaces index-based selection with component api
- Goto commit via vim's search function

### Commit View
- Replace autocmd based highlighting with decorator. the `:update()` call is _very_ expensive.
- Add cherry-pick popup when viewing commit (auto-selects the commit)
- Goto hunk mappings

### Common
- String comparison is faster than `vim.regex` since lua interns strings
- Add common components (extracted from other views)

### Fuzzy Finder
- Didn't need this - the `wrap` api is better.

### Git Command History
- Add common bindings (esc)
- Reduce noise by cleaning command string

### Log View
- Add Cherry Pick popup to visual and normal modes
- Cleanup

### Reflog View
- Add reflog view


lua/neogit/buffers/commit_select_view/ui.lua
```diff
@@ -1,43 +1,20 @@
...
-        text(" "),
-        text(show_graph and ("* "):rep(commit.level + 1) or "* ", { highlight = "Character" }),
-        text(" "),
-        text(table.concat(commit.description)),
-      }
-    )
-  end
-
-  return res
+---@param commits CommitLogEntry[]
```
ten3roberts, line 8: Does this in some way retain the *commit under cursor* functionality?

What it did was if you had the cursor on the `Recent Commits` view or alike it would show that commit at the top so that you can navigate to a commit in the log and e.g invoke rebase from there without chasing your commit again.

lua/neogit/buffers/commit_select_view/init.lua
```diff
@@ -72,8 +75,14 @@ function M:open(action)
         end
       end,
     },
+    after = function()
+      if commit_at_cursor then
+        vim.fn.search(commit_at_cursor.oid)
```
ten3roberts, line 80: I see, so this will instead go to the line where the hoverer commit appears.

What happens if it doesn't show up in the buffer, most likely due to a different number of commits being selected, as the `Recent Commits` section could load more commits than this

lua/neogit/buffers/commit_select_view/ui.lua
CKolkey: Yes! Two ways:
- if you have a commit(s) selected in visual mode, those will be selected automatically.
- otherwise its using `vim.fn.search` in the `after` callback to locate the commit.

lua/neogit/buffers/commit_select_view/init.lua
CKolkey: I default to 256 commits in the select view, and 10 for the recent commits, so, I guess a user could show 257+ commits in the recent view, but... I doubt they would.

PR #534, opened 2023-06-18
Upstreaming  10 - Status

CKolkey:


ten3roberts:
It seems as github has cached and does not recalculate the diff based on the target branch having merges.

Happened yesterday as well with upstreaming 8, and waiting solved the issue then.

CKolkey:
Yeah, the fix is to rebase onto master, then rebase onto the correct target again.

lua/neogit/status.lua
```diff
@@ -937,6 +980,63 @@ local cmd_func_map = function()
         vim.cmd("split " .. item.name)
       end
     end,
+    ["GoToPreviousHunkHeader"] = function()
```
ten3roberts, line 983: Ok, I *really* like this feature ❤️

ten3roberts (approved):

lua/neogit/status.lua
CKolkey: :D Yeah! It's _slightly_ wonky when you're at the bottom/top of a file, but thats just an issue of fine-tuning

PR #535, opened 2023-06-18
Upstreaming  11 - The rest

CKolkey:


.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
+
```
ten3roberts, line 2: Which plugin do you use for indent detection?

I've seen both modeline, editorconfig, vim-sleuth, as well as recent lua alternatives.

todo.md
```diff
@@ -22,11 +21,10 @@
 ## Popups
 
 * chore: rename popup -> transient
-* fix: default values for options don't get displayed on initial draw
-
-## Stash
-
-* feat: delete stash
+* refactor: action/switch/option/config should all exist on the same list within state, allowing better control over
```
ten3roberts, line 24: Absolutely love this. Having a "list" rather than an implementation makes it much easier to a see and audit

ten3roberts (approved):
Note: formatting and `<esc>` to hide process console

.editorconfig
CKolkey: Just using the (now) build-in editorconfig support + https://github.com/nmac427/guess-indent.nvim

todo.md
CKolkey: Agreed. It's _slightly_ hacky right now with the different columns/rows and headers. Works fine, but I totally agree, a list of components is much simpler and lets you do more.

Currently, something I want to do (and can't) is to have _multiple_ 2D-grids within a popup. The current implementation limits you to the actions group at the bottom, and this works well enough that I would consider it a very low-priority fix, but still.. can't do it.

CKolkey:
Alright, with this merged to CKolkey, I'll:

- backport all the changes to my fork
- open one last PR with any remaining changes _not yet on CKolkey_ (I did do some more stuff before I made a development branch on my fork)
- Open final PR with the full commit history

PR #540, opened 2023-07-02
feat! move neogit to https://github.com/NeogitOrg/neogit

ten3roberts:
Neogit has now moved to an organization to ensure the longevity of this project and make it more accessible for maintainers.

This PR notifies users who are still on the `TimUnterberger/neogit` remote using neogits own api, and is consequently most  likely the best example of *dog fooding* I've encountered


CKolkey:
Damn, that was fast :D

PR #542, opened 2023-07-03
Replace repository name in README with NeogitOrg

okuramasafumi:


ten3roberts:
Oh thank you. Forgot about those links ❤️

Issue #552, opened 2023-07-07
`q` should not abort the commit if it contains a written commit message.


ten3roberts:
I've encountered this multiple times because I try `:wq` to save and quit to commit, but accidentally just hit `q` which makes me lose the commit message.

Having a prompt if there are unsaved changes would be preferred

wrightjjw:
Been looking into this. Looks like the commit logic is handled in a `BufUnload` autocmd in the commit editor ([buffers/commit_editor/init.lua](lua/neogit/buffers/commit_editor/init.lua) starting at line 46), which includes the "Are you sure you want to commit?" prompt. Btw, saying no to that discards the commit too - maybe that should also keep the buffer open in case someone wants to reword a message at the last minute.

Anyone know of a way to cancel the autocmd's event during its execution? That would probably be the easiest way to manage this, but I'm not sure if it's possible... A different approach may require reworking the logic outside an autocmd, which could be a rather major task.

By the way, you can also use `ZZ` to commit, which afaik would be equivalent to `:wq` in this case. Would probably cut down on the typos! 😃

CKolkey:
This is slightly tangential, but for the longest time I've had an autocmd set up thats essentially: 
```
autocmd InsertLeave * update
```

So, you leave insert mode, and it write the buffer to disk. The reasoning behind this is pretty simple: if I made changes, I probably wanted to. And writes take a negligible amount of time.

I'm bringing this up because _my_ experience is exactly what we're describing as the ideal - dont let a user accidentally close the buffer with an unwritten commit message, discarding the commit.. because it's almost never in an unwritten state.

It would be trivial to add such an autocmd thats buffer-local, and could even gate it behind a config flag.

ten3roberts:
I am trying to use `ZZ` for commiting, but `:wq` and `q` for closing are *so* deeply engrained in my muscle memory that I do it in our team's discord chat, and close.other popups with `q`.

I think I need a firmware update as well.

Good point that aborting the commit should not discard the message entirely, hadn't thought about it. What I usually do when I remember I forgot to stage a file when writing the commit, I commit anyway and then amend instead.

There are definitely some good questions and considerations regarding the workflow, and I am open to all suggestions.

wrightjjw:
> This is slightly tangential, but for the longest time I've had an autocmd set up thats essentially:
> 
> ```
> autocmd InsertLeave * update
> ```
> 
[...]

PR #553, opened 2023-07-08
wip: Keep the commit dialog open after exit

wrightjjw:
Here's where I got last night working on #552, thought I may as well put it out there. Ideally we stop the buffer from closing at this point, but I don't know if that's possible to do from inside the `BufUnload` autocmd.


CKolkey:
I think this is addressed here: https://github.com/NeogitOrg/neogit/pull/1038

PR #558, opened 2023-07-13
Fix: 557

### LINKED ISSUE #557: Neogit error on startup: not a git repository
Feel free to hide this pass by comment (or should I open an issue for that?).

Well done on all your hard work!

I am getting an error when running setup (at nvim's startup) when not in git repo:
```
git rev-parse --show-toplevel:
fatal: not a git repository (or any of the parent directories): .git
Open the console for details
```

_Originally posted by @gegoune in https://github.com/NeogitOrg/neogit/issues/554#issuecomment-1633955161_

CKolkey:
Check for '.git/' folder before running, otherwise the CLI will return 1, and send a noisy alert to the user that they're not in a repo

fixes #557

### DIFF
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -3,6 +3,7 @@ local logger = require("neogit.logger")
 local a = require("plenary.async")
 local process = require("neogit.process")
 local util = require("neogit.lib.util")
+local Path = require("plenary.path")
 
 local function config(setup)
   setup = setup or {}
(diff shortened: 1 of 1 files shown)


gegoune:
Can confirm it works with that PR. Thanks for quick fix!

Mange:
This makes neogit impossible to use inside of a directory in a project.

For example:

```
monorepo/
	.git/
	project1/
		<I'm here>
	project2/
```

I think it would be better to look for the Git repo directory the same was as Git does, including checking parent directories and `GIT_DIR`.

The command `git rev-parse --is-inside-work-tree` returns one of these things:

| Cause | Exit code | Output |
|---|---|---|
| Inside a working directory | 0 | `"true"` |
| Inside a .git directory | 0 | `"false"` |
| Outside of a repo | >0 | A long error message |


By discarding its output and checking for the status code, you should get a fully compliant check.

ten3roberts:
I wonder why it didn't make an error. Your suggestion about the rev-parse is great, although prior to the merged neogit was able to detect a git directory (and prompt for one if it didn't exist.

The `is_git_dir` should not thrown an error, but rather return it as a false. What exactly had it work before but not after?

PR #559, opened 2023-07-13
Refactor: prevent double setup

CKolkey:
Currently, if a user calls `setup()` in their config, it gets executed twice, since it's called automatically in this file. No real reason for that - we can check if it's setup in the `open` function anyways, which solves the same issue of trying to open the plugin before being setup.

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -6,6 +6,7 @@ local hl = require("neogit.lib.hl")
 local status = require("neogit.status")
 local state = require("neogit.lib.state")
 local input = require("neogit.lib.input")
+local logger = require("neogit.logger")
 
 local cli = require("neogit.lib.git.cli")
 local notification = require("neogit.lib.notification")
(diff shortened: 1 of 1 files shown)


ten3roberts:
I think perhaps that we should not do auto setup at all and make it explicit.

Most (if not all) neovim plugins require explicit setup.

If a setup is done automatically in the open function, it is still quite error prone as you can use other neogit functionalities before that auto setup. So in essence there are now 3 different states of setups, where 2 would be preffered

CKolkey:
Fine with me. It's a breaking change, but it's also undocumented currently that it auto-sets up. I'll just remove that bit.

ten3roberts:
Yes, it is a breaking change, but having this auto magic setup is quite surprising and may lead to inconsistent behavior or values somehow being set.

CKolkey:
Agreed. Plus... it's undocumented, and the "normal" way to do it is to require explicit setup. In fact, I was entirely unaware of this until I was reading through the debug logs and noticed setup being run twice - once without user settings, once with.

CKolkey:
Good to go then?

ten3roberts (approved):
Great.

Good to show an error if neogit is not setup, which will easy any user who have not called setup

Issue #563, opened 2023-07-13
Can't commit (with certain config)


wrightjjw:
### Description

On Windows and Fedora, I can open the commit dialog and put in a commit, but there is no syntax highlighting, and writing the commit dialog does not create a commit.

On WSL the commit dialog doesn't even open...

**Important:** I've narrowed down this problem to probably being related to some plugin I have installed, because running with the minimal config works fine. Thought I'd still open this anyway to see if I can track down a potential fix. [This](https://github.com/wrightjjw/neovim-config/blob/main/lua/plugins.lua) is my plugin config if anyone is crazy enough to look into it themselves.

### Neovim version

NVIM v0.9.1
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wvla -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -fno-common -Wno-unused-result -Wimplicit-fallthrough -fdiagnostics-color=auto -fstack-protector-strong -DUNIT_TESTING -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -I/usr/include/luajit-2.1 -I/usr/include -I/usr/include/luv -I/builddir/build/BUILD/neovim-0.9.1/redhat-linux-build/src/nvim/auto -I/builddir/build/BUILD/neovim-0.9.1/redhat-linux-build/include -I/builddir/build/BUILD/neovim-0.9.1/redhat-linux-build/cmake.config -I/builddir/build/BUILD/neovim-0.9.1/src -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include

### Operating system and version

Fedora 38

### Steps to reproduce

Stage commits, then `cc`

### Expected behavior

commit works

### Actual behavior

it doesn't

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
[...]

PR #568, opened 2023-07-13
Check if treesitter parsers are installed differently

CKolkey:
Fix for #563


cuducos (approved):
I tested it locally (without `gitcommit` tree-sitter installed) and it worked — I could commit.

CKolkey:
Thanks :)

PR #574, opened 2023-07-13
Use remotes.pushDefault value if set in branch config popup

CKolkey:
From: https://github.com/CKolkey/neogit/pull/44/files

### DIFF
--- a/lua/neogit/popups/branch_config/actions.lua
+++ b/lua/neogit/popups/branch_config/actions.lua
@@ -7,12 +7,16 @@ local FuzzyFinderBuffer = require("neogit.buffers.fuzzy_finder")
 local M = {}
 
 function M.remotes_for_config()
-  local remotes = {
-    { display = "", value = "" },
-  }
-
+  local remotes = {}
   for _, name in ipairs(git.remote.list()) do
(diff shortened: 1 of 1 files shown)


ten3roberts:
I suppose this will serve to prefer `origin` in the picker when pushing a new branch?

Won't the remote be duplicated if it already exists?

Additionally, for some repos I have mutliple remotes (one for each fork to checkout PRs). Telescope will favor the initial ordering, so if the `remotes.pushDefault` should be "default" by being the initially selected item it should perhaps be inserted at the beginning of the list.

CKolkey:
Alright, I can break this down. This changes the behaviour of this config on the branch-config popup:

Screenshot 2023-07-13 at 23 10 06

With the current behaviour, the config `remote.pushDefault` is ignored. You can press `p` on that popup to cycle through remotes, with a "none" option available.

Screenshot 2023-07-13 at 23 23 06

If the git config `remote.pushDefault` is set, I noticed that in magit, instead of a none option, it showed `remote.pushDefault:origin` (where origin is the value of that config). This is just to reflect how _git_ works internally - if that value is set, then an _unset_ value for the branch will use the default.

So, this'll never show up in a telescope picker - it's only for the config.

PR #577, opened 2023-07-13
fix: correctly detect the git root location

### LINKED ISSUE #575: Typing `:Neogit` shows nothing in the `NeogitStatus` buffer in symlinked directories not containg a `.git` root
### Description

# Problem
The Neogit Status buffer does not show any content for a symlinked directory belonging to a git repo that does not contain a `.git` directory within its upper paths.

See this short video:

https://github.com/NeogitOrg/neogit/assets/58627896/a7dbd5b6-80a5-4a93-9b91-18356fc3a360

This commit, 3149cd6fda556c9022b3e9dd03aee85a854be478, leads to this behavior. I have determined the *likely* reason this occurs.

Plenary is recursively searching the path for `.git` in upper level directories, but if you've symlinked a directory inside a git repo there is, of course, not a `.git` directory within the symlink.

Take for instance, a structure like so:
```
📁 example/
├── 📁 .git/
└── 📁 sub-dir/
   └── 🗎 empty_file
```
Now if I were to run `Neogit` within this directory, the path traversal would work because there is a `.git` directory in the upper level of the `example` directory.

If I symlink `sub-dir` to somewhere else and then attempt to run `Neogit` within that directory, `Plenary.Path` doesn't find a upper level `.git` directory leading to the issue.

# Solution
Instead of attempting to recursively locate a `.git` directory, a potential solution would be to check the return code of `process.new`.

Something like:
```lua
local function git_root()
  local process = process.new({ cmd = { "git", "rev-parse", "--show-toplevel" } }):spawn_blocking()

  if process ~= nil and process.code == 0 then
    return process.stdout[1]
  else
    return ""
  end
end
```

Patch for that:
```patch
diff --git a/lua/neogit/lib/git/cli.lua b/lua/neogit/lib/git/cli.lua
index 4c43258..45544f5 100644
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -3,7 +3,6 @@ local logger = require("neogit.logger")
 local a = require("plenary.async")
 local process = require("neogit.process")
 local util = require("neogit.lib.util")
-local Path = require("plenary.path")
 
 local function config(setup)
   setup = setup or {}
@@ -449,9 +448,10 @@ local configurations = {
 
 -- TODO: Consider returning a Path object, since consumers of this function tend to need that anyways.
 local function git_root()
-  local dir = Path.new(".git")
-  if dir:exists() and dir:is_dir() then
-    return process.new({ cmd = { "git", "rev-parse", "--show-toplevel" } }):spawn_blocking().stdout[1]
+  local process = process.new({ cmd = { "git", "rev-parse", "--show-toplevel" } }):spawn_blocking()
+
+  if process ~= nil and process.code == 0 then
+    return process.stdout[1]
   else
     return ""
   end
end
```

The above resolves the error I was running into and still ensures that noisy output does not appear, as was the goal of the commit 3149cd6fda556c9022b3e9dd03aee85a854be478.

If the above looks acceptable, I will gladly open a PR to integrate this.

### Neovim version

NVIM v0.10.0-dev-658+g06694203e
Build type: Release
LuaJIT 2.1.0-beta3

Arch, Kernel 6.4.3 (`Linux orion 6.4.3-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 11 Jul 2023 05:13:39 +0000 x86_64 GNU/Linux`)

### Steps to reproduce

# Some manual work required

1. Run the following bash commands (this will create the correct environment to expose the bug):
```bash
mkdir -p "example/sub-dir" &&
	touch "example/sub-dir/empty_file" &&
	pushd "example" && git init && popd &&
	ln -s "example/sub-dir" "sub-dir-symlink"
```
2. Change directory into the `sub-dir-symlink` directory
3. Create the `minimal-init.lua` with the minimal config content
4. `nvim --clean -nu minimal-init.lua`
5. Run the `Neogit` command within Neovim

# Alternative, full setup via bash commands
1. Run this big ol' monster and let it set everything up for you
```bash
mkdir -p "example/sub-dir" &&
	touch "example/sub-dir/empty_file" &&
	pushd "example" && git init && popd &&
	ln -s "example/sub-dir" "sub-dir-symlink"

cat << __EOS__ > minimal-init.lua
for name, url in pairs({
    plenary = "https://github.com/nvim-lua/plenary.nvim",
}) do
	local install_path = vim.fn.fnamemodify("/tmp/nvim_issue/" .. name, ":p")
		vim.fn.system({ "git", "clone", "--depth=1", url, install_path })
	end
end

require("neogit").setup()
print("Minimal Config Ready!")
__EOS__

cd "sub-dir-symlink" && nvim --clean -nu ../../minimal-init.lua -c "Neogit"
```
This will run the `Neogit` command immediately as well, notice the lack of content within the Neogit Status buffer.

### Expected behavior

Neogit Status buffer has content.

### Actual behavior

Neogit Status buffer is empty.

### Minimal config

```Lua
for name, url in pairs({
    plenary = "https://github.com/nvim-lua/plenary.nvim",
}) do
	local install_path = vim.fn.fnamemodify("/tmp/nvim_issue/" .. name, ":p")
		vim.fn.system({ "git", "clone", "--depth=1", url, install_path })
	end
end

require("neogit").setup()
print("Minimal Config Ready!")
```

treatybreaker:
Currently the git root will not be correctly detected within symlinked directores not containing a `.git` directory within its upper paths. This PR resolves that issue by relying entirely upon `git rev-parse`. The commit 3149cd6fda556c9022b3e9dd03aee85a854be478 was created to remove "noisy output" when `git rev-parse` failed. This PR also resolves that issue by checking the return code emitted by `git rev-parse`.

See #575 for more details.

If this PR is not desired, feel free to close it out :smile:.

Closes #575

### DIFF
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -3,7 +3,6 @@ local logger = require("neogit.logger")
 local a = require("plenary.async")
 local process = require("neogit.process")
 local util = require("neogit.lib.util")
-local Path = require("plenary.path")
 
 local function config(setup)
   setup = setup or {}
@@ -449,9 +448,10 @@ local configurations = {
 
 -- TODO: Consider returning a Path object, since consumers of this function tend to need that anyways.
 local function git_root()
-  local dir = Path.new(".git")
-  if dir:exists() and dir:is_dir() then
-    return process.new({ cmd = { "git", "rev-parse", "--show-toplevel" } }):spawn_blocking().stdout[1]
+  local process = process.new({ cmd = { "git", "rev-parse", "--show-toplevel" } }):spawn_blocking()
+
+  if process ~= nil and process.code == 0 then
+    return process.stdout[1]
   else
     return ""
   end
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -448,7 +448,8 @@ local configurations = {
 
 -- TODO: Consider returning a Path object, since consumers of this function tend to need that anyways.
 local function git_root()
-  local process = process.new({ cmd = { "git", "rev-parse", "--show-toplevel" } }):spawn_blocking()
+  local process =
+    process.new({ cmd = { "git", "rev-parse", "--show-toplevel" }, ignore_code = true }):spawn_blocking()
 
   if process ~= nil and process.code == 0 then
     return process.stdout[1]


codebycorey:
Confirming this resolved my issue with my nvim config symlinked. I noticed after the update my git status was an empty buffer.

CKolkey:
Awesome turnaround on this - thanks for the contribution 🙂

Issue #550, opened 2023-07-06
Auto detect diffview and telescope integration


ten3roberts:
diffview integration is currently *explicit*, which means that users who have diffview installed may miss that there is an integration that they can enable hidden inside neogit config 2 levels deep.

The same goes for a soon to land telescope picking. Though this one is more debatable as some people may prefer to use `vim.ui.select` using e.g https://github.com/stevearc/dressing.nvim and keep it more minimal than a full telescope picker. So perhaps it shouldn't auto enable itself and *override* the `vim.ui` picker if you install telescope. In the Rust world that kind of remote influence would be discouraged 😝 

Please chime in and let me know what you think and would like to see with this.

CKolkey:
Yea, for telescope I was thinking we detect/enable by default, but respect if a user sets it to false

wrightjjw:
As a user of both plugins I would love default integration, but would appreciate an option to disable them.

asmodeus812:
@CKolkey further more there is an issue with the current setup with diffview - a default config is passed in overriding the existing user config. The way diffview integration is currently setup, the keymaps are NOT going to be extended in the indented way, i suggest the entire call to setup be simply removed, if the user has configured/installed diffview it will use whatever configuration was provided to it anyway. 

https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/integrations/diffview.lua#L145-L147

```lua
old_config = dv_config.get_config() -- this here, does not return the user config, it instead returns internal diffview transformed representation of the config passed in by the user, as table array, instead of the expected [key][value] table passed in the user space. The extend below will then not extend the config in the indented way, rather will simply override the existing user config with what we see below. 

  local config = vim.tbl_deep_extend("force", old_config, {
    keymaps = {
      view = {
        ["q"] = cb("close"),
        ["<esc>"] = cb("close"),
      },
      file_panel = {
        ["q"] = cb("close"),
        ["<esc>"] = cb("close"),
      },
    },
  })
  dv.setup(config)
```

asmodeus812:
@ten3roberts if possible in the future, could you consider fzf-lua.nvim as well, as alternative to people not using telescope ? Would be great, as of now, fzf plugin can replace the ui.select, which still works great, however having direct integration with it will allow neogit to add custom actions and stages to the picker (in cases where needed), as we do with telescope.

CKolkey:
[...]

PR #583, opened 2023-07-14
feat: auto detect diffview and telescope integration #550

milkias17:


ten3roberts:
Wow thanks.

This solution is exactly what I had in mind

PR #584, opened 2023-07-14
test: add tests for git root detection

treatybreaker:
This commit also allows the `in_prepared_repo` harness to pass its directory to the function it wraps, which is necessary to check the root directory.

This helps with validation and checks of the git root directory, should help with avoiding the problem that cropped up in #575.


CKolkey:
Keep up the great work 🥳

PR #585, opened 2023-07-14
ci: cache selene install

treatybreaker:
This speeds up the Selene job in the lint action significantly when the cache is valid. Selene's last update was 4 months ago so there's really no need to be installing it on every run. Much faster to use a cache.

Take a look at [this run](https://github.com/treatybreaker/neogit/actions/runs/5556895200/jobs/10149993953) to see the speed up. It only took the Selene job ~9 seconds to complete with the cache.


CKolkey:
Love it - this was on my TODO, so, glad you beat me to it 😍

PR #588, opened 2023-07-14
Defer entering data into console until absolutely necessary

### LINKED ISSUE #587: Neogit seems to call TermOpen via setup
### Description

One of the recent Neogit updates seems to be causing Neovim to call TermOpen when the setup function is called. I like to have my terminals in Neovim open in insert mode, so I have the following in my config: `vim.api.nvim_create_autocmd("TermOpen", { command = "startinsert" })`. 

This results in Neovim's normal buffers opening in insert mode, rather than normal mode, when it should only be applied to the terminal buffers.

I installed Neogit for the first time yesterday and didn't have this problem, but after updating my plugins today I noticed it, which is why I think a recent update is the culprit. 

### Neovim version

NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3

Arch Linux

### Steps to reproduce

1. Update Neogit to the current release with the TermOpen autocmd
2. Open Neovim

### Expected behavior

Neovim opens in normal mode

### Actual behavior

Neovim opens in insert mode

### Minimal config

```Lua
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
vim.api.nvim_create_autocmd("TermOpen", { command = "startinsert" })
```

treatybreaker:
Prior to this commit the process module would spawn a Neovim terminal in the background immediately and write whatever data it needed to. This interferes with user auto commands for `TermOpen`. See #587 for an example.

This commit defers opening the terminal until absolutely necessary and writes all the data to it at that point. This ensures relevant auto commands for `TermOpen` are properly issued when the term is going to be actually used, not when its been backgrounded.

Another thing this commit does is improve the scroll function to use in-built functionality from Neovim instead of parsing the terminal buffer. By going into insert mode, waiting a small amount of time, and then exiting it ensures the cursor is directly at the end of viewable content.

Closes #587

### DIFF
--- a/lua/neogit/process.lua
+++ b/lua/neogit/process.lua
@@ -108,40 +108,15 @@ local function create_preview_buffer()
   }
 end
 
---from https://github.com/stevearc/overseer.nvim/blob/82ed207195b58a73b9f7d013d6eb3c7d78674ac9/lua/overseer/util.lua#L119
----@param win number
-local function scroll_to_end(win)
-  local bufnr = vim.api.nvim_win_get_buf(win)
-  local lnum = vim.api.nvim_buf_line_count(bufnr)
-  local last_line = vim.api.nvim_buf_get_lines(bufnr, -2, -1, true)[1]
-  -- Hack: terminal buffers add a bunch of empty lines at the end. We need to ignore them so that
-  -- we don't end up scrolling off the end of the useful output.
-  -- This has the unfortunate effect that we may not end up tailing the output as more arrives
-  if vim.bo[bufnr].buftype == "terminal" then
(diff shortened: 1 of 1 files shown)


CKolkey:
Fantastic - it looks a lot nicer to delay that until absolutely needed. Thanks again :D

PR #589, opened 2023-07-14
fix: make neogit respect user's configuration of diffview

milkias17:
mentioned in #550

### DIFF
--- a/lua/neogit/integrations/diffview.lua
+++ b/lua/neogit/integrations/diffview.lua
@@ -142,7 +142,7 @@ local function get_local_diff_view(selected_file_name)
 end
 
 function M.open(section_name, item_name)
-  old_config = dv_config.get_config()
+  old_config = vim.deepcopy(dv_config.get_config())
 
   local config = dv_config.get_config()
 
@@ -178,7 +178,6 @@ function M.open(section_name, item_name)
   if view then
     view:open()
   end
-
   return view
 end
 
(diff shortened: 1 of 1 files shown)


milkias17:
@ten3roberts Any problems with the implementation😅?

ten3roberts (commented):
I suppose this works by not replacing the table reference, but rather works and mutates the existing reference?

milkias17:
Yup exactly. We are basically using the exact same method that [diffview.nvim](https://github.com/sindrets/diffview.nvim/blob/b0cc22f5708f6b2b4f873b44fbc5eb93020f4e0c/lua/diffview/config.lua#L628) uses to merge the default configuration with the users configuration.

ten3roberts (approved):

ten3roberts:
Is this all done or is there anything more you want to get in?

Thanks for your help, I'm looking forward to having *neogit invoked* diffview be consistent with the *user invoked* diffview 🚀

milkias17:
Yea interms of repsecting user's configuration that's about it. However the thing mentioned about having a keybinding to invoke the commit popup within the diffview buffer, should I add that or?

ten3roberts:
Perhaps that feature would be best under a config option, since some people may not want their difftview keybindings to be different and inconsistent. There is an in-progress PR which modifies the config file, so if we do this it would be best as another PR after that one.

PR #592, opened 2023-07-14
fix: ensure floating popup windows are auto closed when they lose focus

### LINKED ISSUE #591: Moving cursor out of a floating popup leaves the popup inaccessible
#### Problem
See this video:

https://github.com/NeogitOrg/neogit/assets/58627896/2af3ac97-63c8-4f1d-8155-20d69440b90a

Even when I use my mouse to *click* on the window, I'm unable to access it. Effectively this results in a floating window covering the editor until I exit Neogit.

#### Solution

I haven't looked yet, but deep in the Neogit window code a auto command should be created checking for focus. If focus is lost (`WinLeave` event) then it should be closed.

Will look something like:

```lua
vim.api.nvim_create_autocmd("WinLeave", {
    callback = function(win)
        if win.buf == popup.buf then
            vim.api.nvim_win_close(win.id, true)
            return true -- Deregisters the autocmd
        end
    end,
})
```

### Neovim version

NVIM v0.10.0-dev-672+g33e1a8cd7
Build type: Release
LuaJIT 2.1.0-beta3

Arch, Kernel 6.4.3 (`Linux orion 6.4.3-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 11 Jul 2023 05:13:39 +0000 x86_64 GNU/Linux`)

#### Manual work required

1. Ensure you are inside a git repository with a staged change
2. `nvim --clean -nu minimal.lua`
3. Run `:Neogit`
4. Open the commit popup
5. Hit `<C-W>w` to switch to the lower window
6. Notice you're unable to interact with the commit popup now covering your editor

#### Paste from Bash and let it set up an ideal environment
```bash
pushd "$(mktemp -d)"
cat << __EOS__ > minimal-init.lua
for name, url in pairs({
    plenary = "https://github.com/nvim-lua/plenary.nvim",
}) do
    local install_path = vim.fn.fnamemodify("/tmp/nvim_issue/" .. name, ":p")
        vim.fn.system({ "git", "clone", "--depth=1", url, install_path })
    end
end

require("neogit").setup({
    popup = {
        kind = "floating",
    },
    commit_popup = {
        kind = "floating",
    },
})

-- Yes, I realize that all this nesting here is a crime against god,
-- but I'm too lazy to use vim.loop.new_timer()
vim.defer_fn(function()
    vim.cmd.Neogit()
    vim.defer_fn(function()
        vim.api.nvim_feedkeys("c", "", true)
            vim.defer_fn(function()
                vim.cmd("wincmd w")
                print("Notice that you can't use the commit popup at all :(")
            end, 100)
    end, 300)
end, 300)
__EOS__

git init
printf "Example Content\n" >> example_file
nvim --clean -nu minimal-init.lua
```

This will (should) do *everything*, including opening the commit popup. There is nothing you have to do, but observe the incorrect functionality.

### Expected behavior

The commit popup window closes when it loses focus in float mode.

### Actual behavior

The commit popup window covers the editor and is unable to be interacted with.

### Minimal config

```Lua
for name, url in pairs({
    plenary = "https://github.com/nvim-lua/plenary.nvim",
}) do
    local install_path = vim.fn.fnamemodify("/tmp/nvim_issue/" .. name, ":p")
        vim.fn.system({ "git", "clone", "--depth=1", url, install_path })
    end
end

require("neogit").setup({
    popup = {
        kind = "floating",
    },
    commit_popup = {
        kind = "floating",
    },
})
print("Minimal Config Ready!")
```

treatybreaker:
Prior to this commit any floating windows that lost focus were not closed and covered the editor as well as being inaccessible in that state.

This commit ensures the window is properly cleaned up when it loses focus. This only applies to *floating* windows.

See #591 for more details.

Closes #591

### DIFF
--- a/lua/neogit/lib/popup.lua
+++ b/lua/neogit/lib/popup.lua
@@ -575,6 +575,19 @@ function M:show()
       }
     end,
   }
+
+  -- Closes the window if it loses focus and it is a floating buffer
+  if self.buffer.kind == "floating" then
+    vim.api.nvim_create_autocmd("WinLeave", {
+      callback = function(win)
+        if win.buf == self.buffer.handle then
+          -- We pcall this because it's possible the window was closed by a command invocation, e.g. "cc" for commits
+          pcall(self.close)
+        end
+      end,
+      once = true,
+    })
+  end
 end
 
 return M
--- a/lua/neogit/lib/popup.lua
+++ b/lua/neogit/lib/popup.lua
@@ -574,20 +574,15 @@ function M:show()
         },
       }
     end,
-  }
-
-  -- Closes the window if it loses focus and it is a floating buffer
-  if self.buffer.kind == "floating" then
-    vim.api.nvim_create_autocmd("WinLeave", {
-      callback = function(win)
-        if win.buf == self.buffer.handle then
+    autocmds = {
+      ["WinLeave"] = function()
+        if self.buffer.kind == "floating" then
           -- We pcall this because it's possible the window was closed by a command invocation, e.g. "cc" for commits
-          pcall(self.close)
+          pcall(self.close, self)
         end
       end,
-      once = true,
-    })
-  end
+    },
+  }
 end
 
 return M


lua/neogit/lib/popup.lua
```diff
@@ -575,6 +575,19 @@ function M:show()
...
   }
+
+  -- Closes the window if it loses focus and it is a floating buffer
+  if self.buffer.kind == "floating" then
+    vim.api.nvim_create_autocmd("WinLeave", {
+      callback = function(win)
+        if win.buf == self.buffer.handle then
+          -- We pcall this because it's possible the window was closed by a command invocation, e.g. "cc" for commits
+          pcall(self.close)
+          return true -- Deregisters the autocmd
```
gegoune, line 586: Is it the same as setting `once` flag?

treatybreaker: Huh, you're right per [the neovim source code](https://github.com/neovim/neovim/blob/33e1a8cd7042816a064c0d2bf32b6570d7e88b79/src/nvim/autocmd.c#L2083). Will change.

CKolkey:
Cool! But the buffer's `#create` function can take an `autocmds` key, like https://github.com/NeogitOrg/neogit/blob/d5f2a9670a924a9260b83ba415098ce5d88f6c07/lua/neogit/buffers/commit_editor/init.lua#L42 - I think this could probably take advantage of that, no?

treatybreaker:
This has lead me to find what is arguably a bug in Neovim upstream and unfortunately in Neogit.

Using the `nvim_create_autocmd` syntax like we are currently, Neovim never cleans up the autocmds even when the buffers they're attached to are gone.
This syntax:
https://github.com/NeogitOrg/neogit/blob/33016ad20b92b7eb83d27880b4abfbce53b467ad/lua/neogit/lib/buffer.lua#L405

Using that is equivalent to this:
```lua
api.nvim_create_autocmd(event, { callback = callback, pattern = string.format("buffer=%s", buffer.handle) })
```
and since that is evaluated as a pattern in Neovim's source code that autocmd never gets removed after the buffer it's "attached" to is gone.

I'll update to use the in-built autocmd creation stuff in neogit for this PR real quick. Then I'll open another PR here that will return `true` in the callbacks if the buffer they're matching for is gone (which deletes the autocmds when true is returned from the callback), and then go cry to Neovim upstream maybe with a PR.

treatybreaker:
That should do it.
I'm creating all the options in a separate table first so I can conditionally create the winleave autocmd (as we only want it affecting floating windows) then calling `Buffer.create` against that table.

lua/neogit/lib/popup.lua
```diff
@@ -575,6 +575,21 @@ function M:show()
       }
     end,
   }
+
+  if buf_options.kind == "floating" then
```
CKolkey, line 579: I'd just keep the constructor as is, and instead of conditionally adding the autocmd, put the condition in the callback function. That'll keep the constructor consistent with the other callers.

CKolkey (changes requested):

lua/neogit/lib/popup.lua
treatybreaker: Done in the latest force push.

CKolkey (approved):

PR #597, opened 2023-07-15
fix: update repo name to new one in README.md

bloznelis:


ten3roberts:
Thank you. Seems like I missed that spot 😅

Issue #581, opened 2023-07-14
Provide a CONTRIBUTING.md file


miguelcrespo:
Hi, I was checking out the code of this plugin since I use it a lot and I wanted to understand the code and I notice there's no `CONTRIBUTING.md`, it might be beneficial to include one to give a brief introduction of how to contribute to this project, some sections I have in mind:

- Introduction to the project and architecture
- Code standards
- How to run the project locally and start developing

ten3roberts:
That is a good idea and would further help Neogit becoming an active project.

I'll also add for myself to remember that the `todo.txt` could be removed as it is mow far from relevant

PR #599, opened 2023-07-15
feat: `CONTRIBUTING.md`

ten3roberts:
Resolves: #581

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,11 +31,25 @@ Opening a view is typically done through a *popup* which allows you to configure
 
 These reside inside [`./lua/neogit/popups/`], and are split into `init.lua` and `actions.lua` for the setup and keybindings, and the git commands to execute, likewise intended to get an overview of the options and keybindings for the popup in `init.lua` without concerning yourself with the git commands and parsing in `actions.lua`.
 
+## Getting Started
+
+If you are using [Lazy.nvim](`https://github.com/folke/lazy.nvim`) you can configure it to prefer sourcing plugins from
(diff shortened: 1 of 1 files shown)


CKolkey:
Great start. I'd be happy to help expand on the architecture section, unless you want to. The only thing I'd add is more insistence on test coverage.. Thats already showing up as a weak spot with fixes breaking things...

ten3roberts:
Yes, I'll add that section.

What is the general concensus in markdown? Should proper nouns such as Neogit and Magit be put in *Neogit* or `Neogit`, or just plain?

CKolkey:
Hmm. I don't have a strong feeling on it. I like the backticks, as it makes it stand out nicely, but I'm pretty open to either.

CONTRIBUTING.md
```diff
@@ -0,0 +1,91 @@
...
+[`Plenary`](https://github.com/nvim-lua/plenary.nvim) for logging.
+
+Export the environment variables `NEOGIT_LOG_CONSOLE="sync"` to enable logging, and `NEOGIT_LOG_LEVEL="debug"` for more
+verbose logging.
+
+```lua
+local logger = require("neogit.logger")
+
+logger.fmt_info("This is a log message: %d", 2)
+logger.fmt_debug("This is a verbose log message: %q", status)
```
CKolkey, line 60: it's just `info` and `debug`, `trace`, `warn`- no prefix

ten3roberts:
I think this should be a good enough starting point for the contributing guide, and over time we'll refine and add more sections to it when we find parts to be unintuitive.

PR #601, opened 2023-07-15
feat: auto load popups and auto populate `help` popup

ten3roberts:
This allows less locations to edit when adding a new popup


ten3roberts:
Figured when working on #599 that I'd make it such that the `Help` popup always reflects the popups available in the status buffer, and that they all work the same way.

This way it is easier to contribute new ones, as there were previously 3 different files oyu had to change to add one, now it is only one.

This is almost ready, just have to fix the lints.

CKolkey:
Yeah, consolidating that is a great move.

ten3roberts:
Yes, the help buffer is a lot more useful now and is guaranteed to work the same (rather than error out as a popup required an argument when starting, but was supplied `nil` in the help buffer). A common example of this is the stash help which requires special feeding

lua/neogit/popups/echo/actions.lua
CKolkey, line 1: Is this just for testing?

lua/neogit/popups/init.lua
```diff
@@ -0,0 +1,109 @@
+local M = {}
+
+---@param name string
+---@param f nil|fun(create: fun(...)): any
+--- Creates a curried function which will open the popup with the given name when called
+function M.open(name, f)
+  f = f or function(f)
```
CKolkey, line 7: Can you explain how this works and why? 😅

ten3roberts:
It all works on my side, and the tests.

Though it is quite a large change as it touches the status buffer, so let me know if you find anything

lua/neogit/popups/init.lua
```diff
@@ -0,0 +1,109 @@
...
+function M.mappings_table()
+  local config = require("neogit.config")
+  local async = require("plenary.async")
+  return {
+    {
+      "HelpPopup",
+      "Help",
+      M.open("help", function(f)
+        f {
+          use_magit_keybindings = config.values.use_magit_keybindings,
```
CKolkey, line 36: I'm a little torn on this pattern of reaching into the status buffer for stuff. On the one hand, its a little cleaner than needing to pass the line into the constructor, but on the other hand it makes this function dependent on external state instead of passing that state in as an argument.

Not saying to change this, just voicing my thoughts

lua/neogit/status.lua
```diff
@@ -611,6 +611,8 @@ local function get_selected_commits()
   return items
 end
 
+M.get_selectet_commits = get_selected_commits
```
CKolkey, line 614: Can we just expose this at the function definition instead of here? It makes the goto definition work better that way

CKolkey:
Generally I really like this - hopefully I'll make some time later to look in more detail. Left some comments, I can definitely tell you're more at home with functional style stuff than I am - I dig that. Nice to learn new stuff

lua/neogit/popups/init.lua
ten3roberts: Yes, I am as well.

I usually go about coding with a domain driven approach. Some code lives in a certain domain/space, and can enter a smaller contained domain. Code usually becomes messier when a code enters a larger domain than which is was called from. Think of a graph which has to remain acyclic, like a tree.

The status buffer would need a bit of splitting like you did for the repository.

ten3roberts: Perhaps splitting it as well into an `init/view` and `actions` and then using actions in other places.

lua/neogit/popups/echo/actions.lua
ten3roberts: Yes, it is for a unit test inside `M.test`

lua/neogit/popups/init.lua
ten3roberts: Yes absolutely.

It is a function which given a popup name will return a function (which takes 0 arguments) which will when called load the popup `neogit.popups/{name}` module, and then pass its `create` function to the second argument `f`.

`f` can then call the `create` function with any arguments it likes, or whenever it wants.

I originally had it be plain argument, and `create` was called inside the `open` returned function instead, but this allows async and sleeping, as well as aborting the popup using the supplied function.

Essentially, this function will create a keymapping ready function, which will supply `f` with the means to open the correct popup.

the `f = f or function(create) create() end` will simply call the `create` function directly with no arguments if no second argument is supplied. Most popups don't need any extra arguments, but some do, such as stashing.

The normal flow is 

`open` loads popup module
uses `f` to help calling the `popup.create` method.

In most cases, if the popup.create requires no special loading or arguments,

`popup.open("my_awesome_popup") -> function() which opens popup` is enough

This is usually called *lifting* and creating a function which "fills in" some arguments and returns a function, eg `function(a,b) -> function(b)` is called currying (partial application)

lua/neogit/popups/init.lua
CKolkey: Splitting the status buffer up is a longer-term goal of mine, but it clearly will take some doing. It also pre-dates the buffer lib, so doesn't really take advantage of the library that all the popups/buffers do. _Eventually_ I'd like the status buffer to not have a special place like it currently does, and just be one more view (albeit the "main" view), but given the size/complexity of it, thats not a simple undertaking. So, small bits here and there.

It may be a good middle-step to use the `init/view`/`actions` pattern - I'm not against that.

CKolkey:
Everything looks great, except hitting `?` pops up a notice saying there's no help popup :P 

All the other popups seems to work though :)

ten3roberts:
That is strange. I am not getting that error, though it makes the popup close itself, likely a "race" between closing the previous popup and opening a new popup with the same name

ten3roberts:
This behavior can be seen on master as well

CKolkey:
It can? The help popup comes up fine for me.

(also, this is why we need better test coverage)

ten3roberts:
Yes. What happens if you do `?` and then a second later `?`

For me that closes the help popup.

There are definitely a lot of the older parts of the neogit that need to be fixed and are quite buggy and magic.

ten3roberts:
How about we merge this for now, and then fix any outlying errors afterwards?

The contributing PR depends on this one, and having the contributing guide merged sooner than later would be preferred

CKolkey:
Lemme just verify the help popup is working first

CKolkey:
Ok, the problem was on my end: I had `Toggle` bound to both `h` and `l` (to work more like a file tree), and the help popup didn't like that. Unfortunately the error saying "popup not found" wasn't the most helpful, and I discovered this by adding `error(value)` to `popups/init.lua`.

So, everything looks good, but maybe we can improve the error handling?

lua/neogit/popups/echo/init.lua
CKolkey, line 1: Could you add a comment to the top of this file that it's for testing only?

lua/neogit/popups/echo/actions.lua
CKolkey, line 1: Same, add a note that this is for testing only

ten3roberts:
Oh, great. I'll fix that then.

What did `error(value)` show, or in other words how does it fail to do `pcall(require, "mod")`?

lua/neogit/popups/init.lua
```diff
@@ -0,0 +1,109 @@
...
+    { "CommitPopup", "Commit", M.open("commit") },
+    { "LogPopup", "Log", M.open("log") },
+    {
+      "CherryPickPopup",
+      "Cherry Pick",
+      {
+        "nv",
+        M.open(
+          "cherry_pick",
+          async.void(function(f)
```
CKolkey, line 62: Can you verify that this needs to be `async.void`? I'm not sure that it's needed, especially considering the revert popup isn't, and... they're real similar.

CKolkey (changes requested):

CKolkey:
> Oh, great. I'll fix that then.
> 
> What did `error(value)` show, or in other words how does it fail to do `pcall(require, "mod")`?

If I add this to the `setup()`, the following error occurs
```lua
    mappings = {
      status = {
        ["l"] = "Toggle",
        ["h"] = "Toggle",
      },
    },
```

```
E5108: Error executing lua /Users/cameron/code/neogit/lua/neogit/popups/init.lua:20: vim/shared.lua:436: 
The reverse lookup found an existing value for "Toggle" while processing key "<tab>"
stack traceback:
        [C]: in function 'error'
        /Users/cameron/code/neogit/lua/neogit/popups/init.lua:20: in function 'mapping'
        .../cameron/code/neogit/lua/neogit/lib/mappings_manager.lua:22: in function 'invoke'

        [string ":lua"]:1: in main chunk
```

lua/neogit/popups/init.lua
ten3roberts: I don't think it needs to either, it was how it was at present.

The difference is that async.void may run on the next microtick, though not exactly sure. It should be immediate unless it needs that microframe delay for getting the visual selection. Will remove and tset

ten3roberts:
Oh that's interesting, so the require fails because setup code is run when the module is loaded, and that fails.

I thought that require would only fail if not found, but I suppose it failing if anything errors makes sense, as lua allows generating modules entirely programmatically when loading instead of from a file (if you are crazy enough to do that and throw debuggabability out of the window)

lua/neogit/popups/init.lua
CKolkey: Yea, pretty sure I added that because I had no idea how async stuff worked when I did.

ten3roberts: I tested it and it works without.

There is an error on master when doing `AA` on a visual selection. More testing (again) would catch it

```sh
Error executing vim.schedule lua callback: ...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:18: The coroutine failed with this message: ...rs/teiroberts/dev/nvim/neogit/lua/neogit/lib/git/cli.lua:749: invalid value (table) at index 8 in table for 'concat'
stack traceback:
	[C]: in function 'error'
	...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:18: in function 'callback_or_next'
	...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:45: in function <...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:44>

```

CKolkey (approved):
Great work :D

The error popup is a nice touch

gegoune:
Seems like one can no longer just do `vim.keymap.set('n', '<leader>gc', neogit.popups.commit.create)` - how can users invoke specific popups now?

ten3roberts:
```lua
vim.keymap.set('n', '<leader>gc', neogit.popups.open("commit"))
```

Would be enough if you do not want to pass extra data to the popup create function.

PR #604, opened 2023-07-15
fix: do not directly modify the mapping config, use a copied value

### LINKED ISSUE #561: All of the status commands are mapped as a normal mode commands
### Description

tbl_add_reverse_lookup modifies the input therefore all status command names are mapped as normal mode commands.

https://github.com/NeogitOrg/neogit/blob/4ed8336fc54a66102617f82d01a68fc3b0e94594/lua/neogit/popups/help/actions.lua#L6

```
:nmap GoTo
n  GoToPreviousHunkHeader *@<Cmd>lua require 'neogit.lib.mappings_manager'.invoke(2, 48)<CR>
n  GoToNextHunkHeader *@<Cmd>lua require 'neogit.lib.mappings_manager'.invoke(2, 32)<CR>
n  GoToFile    *@<Cmd>lua require 'neogit.lib.mappings_manager'.invoke(2, 13)<CR>
```

### Neovim version

NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3

Gentoo Linux

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. `:lua require('neogit').open()`
3. `:nmap`

### Expected behavior

Only "key" is mapped, not "command name".

### Actual behavior

Both "key" and "command name" are mapped.

### Minimal config

```Lua
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
```

treatybreaker:
Prior to this commit the reverse lookup done was directly modifying the `neogit.config.values.mappings.status` table which is not desired leaving status command names mapped (see `nmap Goto` for instance). We really want to modify a *copied* value to use later within the code to avoid mapping those commands. That is what this PR does.

See #561 for more details.

Closes #561

### DIFF
--- a/lua/neogit/popups/help/actions.lua
+++ b/lua/neogit/popups/help/actions.lua
@@ -3,7 +3,10 @@ local M = {}
 local util = require("neogit.lib.util")
 local NONE = function() end
 
-local status_mappings = vim.tbl_add_reverse_lookup(require("neogit.config").values.mappings.status)
+-- Using deep extend this way creates a copy of the mapping values
+local status_mappings = vim.tbl_add_reverse_lookup(
+  vim.tbl_deep_extend("force", require("neogit.config").values.mappings.status, {})
+)
 
 local function present(commands)
   local presenter = util.map(commands, function(command)


CKolkey:
This is a really clever way to do this, nice. There is a utility function, in `lua/neogit/lib/util.lua`, build_reverse_lookup - does that work here?

treatybreaker:
> There is a utility function, in lua/neogit/lib/util.lua, build_reverse_lookup - does that work here?

When I use `build_reverse_lookup` it doesn't seem to do the same things as the `vim.tbl...` stuff. Getting an error where it's unable to compare the keys on line 17 of that file here:
https://github.com/NeogitOrg/neogit/blob/33016ad20b92b7eb83d27880b4abfbce53b467ad/lua/neogit/popups/help/actions.lua#L16

I could *probably* work around it, but the vim functions are built-in and probably should replace the util code as they become more available in each version.

CKolkey:
Cool! Didn't know that. Alright, don't worry about it then.

PR #605, opened 2023-07-15
fix: use error code from `git rev-parse --is-inside-work-tree` to check if inside git directory

rish987:
In my case `/bin/bash` was spitting a warning message, which caused the string equality check to fail. It should be more robust to check for an exit code of 0 instead.


CKolkey:
Agreed! Thanks

PR #606, opened 2023-07-15
fix: ensure all created autocmds are part of the Neogit augroup

treatybreaker:
Prior to this PR using `vim.api.nvim_get_autocmds({ group = "Neogit" })` would only return the initial `setup` autocmds that Neogit created. It did *not* return the autocmds created for the buffers that Neogit created.

This PR enables users to get all the autocmds that Neogit creates. It does this by exporting the augroup in `autocmds.lua` and then importing it in `buffer.lua` so they share the same augroup.

Something to note, we may want to create a different `augroup` for the Neogit buffers in `buffer.lua` instead of using the same `augroup` in `autocmds.lua`, though it may make it a bit more difficult to search for.

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -123,4 +123,5 @@ return {
   close = status.close,
   setup = setup,
   complete = complete,
+  autocmd_group = vim.api.nvim_create_augroup("Neogit", { clear = false }),
 }
--- a/lua/neogit/autocmds.lua
+++ b/lua/neogit/autocmds.lua
@@ -1,7 +1,7 @@
 local M = {}
 
 local api = vim.api
-local group = api.nvim_create_augroup("Neogit", { clear = true })
+M.group = api.nvim_create_augroup("Neogit", { clear = true })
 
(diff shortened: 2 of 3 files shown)


CKolkey:
The only thing I'd say is to move the `group` into the setup, since I kinda have short-term plans to remove that `autocmds` file. You can check that out here https://github.com/CKolkey/neogit/pull/44 if you're curious. Otherwise this makes perfect sense :)

treatybreaker:
Do you mean within the `setup` function in `autocmds.lua` or elsewhere? Like the `setup` function in the primary file?

The only other convenient place I see to set the `augroup` is in the config table.

CKolkey:
Sorry, I mean the setup function for the plugin: https://github.com/NeogitOrg/neogit/blob/master/lua/neogit.lua

Then export it from the module, and consume it where needed. Or something like that.

treatybreaker:
@CKolkey done. Take a look and let me know what you think.

CKolkey:
I think it looks great :D

PR #614, opened 2023-07-15
File watcher wip

CKolkey:
Major changes:
### Watcher
- Adds a filesystem watcher, which keeps an eye on the `.git/` folder. Any changes will refresh the internal state for the repo. The major advantage of this is that during things like a cherry pick, merge, or rebase, etc, the status buffer will always be correct. The filesystem watcher is debounced to prevent over-eager refreshing.

### Repo
- Adds `#dispatch_refresh` to refresh the repo state in an async context. 
- A semaphore is used to prevent both action-triggered refreshing and watcher-triggered refreshing from happening concurrently.
- Add a table for invalidating diffs. Mostly used when staging/unstaging files.
- `index_stat` added to capture timestamp of `.git/index` and not refresh repo state if the timestamp hasn't changed. Possibly dumb.

### Status
- `status.refresh()` no longer takes a table for which things to refresh. We can just refresh everything every time as it's always done in an async context.
- Remove the refresh lock stuff, since the repo is what will always trigger a refresh, and _that_ has a lock instead.
- add `M.update()` calls to some of the status buffer actions to trigger a state refresh.
- `M.get_status()` doesn't seem to be used anywhere.. Though now I'm wondering if it's not part of our "public api" :\

### Other
- Remove the responsibility of popup actions to call `a.util.scheduler()` and `status.refresh()`. The action's function is wrapped, and the scheduler/status functions are always invoked. The biggest disadvantage of this is you don't see the name of the function that ran, but in practice that wasn't used.


lua/neogit/lib/git.lua
ten3roberts, line 1: :'<,'>sort :smile:

lua/neogit/watcher.lua
```diff
@@ -0,0 +1,85 @@
...
+    end)
+  )
+
+  logger.debug("[WATCHER] Watching git dir: " .. gitdir)
+
+  local w = assert(uv.new_fs_event())
+
+  w:start(
+    gitdir,
+    {},
```
ten3roberts, line 54: Should this be recursive?

I don't think this will catch changes in `.git/index/file` etc, though I may be wrong

http://docs.libuv.org/en/v1.x/guide/filesystem.html#file-change-events

https://github.com/luvit/luv/blob/master/docs.md#uvfs_event_startfs_event-path-flags-callback

tests/init.lua
```diff
@@ -24,5 +24,5 @@ end
 
 require("plenary.test_harness").test_directory("tests/specs", {
   minimal_init = "tests/minimal_init.lua",
-  sequential = true,
+  -- sequential = true,
```
ten3roberts, line 27: Sequential makes the output of tests not interleave. Any reason to change this?

lua/neogit.lua
```diff
@@ -39,8 +40,7 @@ local setup = function(opts)
   hl.setup()
   signs.setup()
   state.setup()
-
-  require("neogit.autocmds").setup()
+  watcher.setup()
```
ten3roberts, line 43: I don't think the watcher should start inside config.setup, but rather when you open the status buffer.

How does this handle `Neogit cwd="other"` or when multiple different repos are used.

Additionally, I don't see any cleanup, as with file watchers, it is very common to create `EMFILE: Too many open files`, especially if running multiple watchers. The number of plugins that I've encountered this exact error for when they are using fs watching is quite high, and I've also run into this error with syncthing as well, though it catches it and handles it by reducing the watching.

lua/neogit/lib/git/stash.lua
```diff
@@ -105,7 +105,7 @@ function M.drop(stash)
 end
 
 function M.list()
-  return cli.stash.args("list").call():trim().stdout
+  return cli.stash.args("list").call_sync():trim().stdout
```
ten3roberts, line 108: I'm sorry but I do not think that solving this by blocking operations is the right call. It will cause the editor to stutter and get small hangups when doing things, especially if you leave the status buffer open.

The `can't yield across C boundary means that the function was called from a callback such as `vim.fn.defer` or similar and not in the callstack of any `a.void`

ten3roberts (changes requested):

lua/neogit/watcher.lua
CKolkey: Yeah, it's recursive, no worries about that

tests/init.lua
CKolkey: No, don't think so. I'll put it back.

lua/neogit/watcher.lua
CKolkey: Huh, maybe I'm mistaken. Not passing the recursive flag.

```
                - `flags`: `table`
                  - `watch_entry`: `boolean` or `nil` (default: `false`)
                  - `stat`: `boolean` or `nil` (default: `false`)
                  - `recursive`: `boolean` or `nil` (default: `false`)
```

lua/neogit.lua
CKolkey: > I don't think the watcher should start inside config.setup, but rather when you open the status buffer.
Reasonable. It's to trigger a refresh of the `status` buffer, so, the init method of the `status` module seems like a good spot.

> How does this handle `Neogit cwd="other"` or when multiple different repos are used.
It.. doesn't. In fact, I think my previous refactor of extracting `status.repo` into `git.repo` and treating it like a singleton broke that. I'll try to work out a way to support this.


> Additionally, I don't see any cleanup, as with file watchers, it is very common to create `EMFILE: Too many open files`, especially if running multiple watchers. The number of plugins that I've encountered this exact error for when they are using fs watching is quite high, and I've also run into this error with syncthing as well, though it catches it and handles it by reducing the watching.

Fair. Since this isn't recursive, it's just watching the top-level of `./.git/` for changes, so there are only a handful of files watched. Still, it would be easy to use `:stop()` when the status buffer closes.

lua/neogit/lib/git/stash.lua
CKolkey: Alright. What do you suggest to prevent the `can't yield across C boundary` issue? For what its worth, there's no discernable stutter/lag of the UI for me.

lua/neogit/lib/popup/builder.lua
```diff
@@ -267,10 +269,29 @@ function M:action(keys, description, callback)
...
+      local refresh = function()
+        git.repo:dispatch_refresh {
+          callback = require("neogit.status").dispatch_refresh,
+          source = "action",
+        }
+      end
+
+      a.run(cb, refresh)
+    end)
+  end
```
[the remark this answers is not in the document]
CKolkey: I updated this - didn't see a reason to `void()` wrap the outer function when using `run` inside it. Maybe thats wrong, but `run()` will run the `cb` function in a coroutine, which is what we care about, not this outer function thats just creating some closures.

ofseed:
Could this PR solve the problem of existing buffers not being refreshed after switching branches using neogit?

CKolkey:
> Could this PR solve the problem of existing buffers not being refreshed after switching branches using neogit?

Hm. Maybe? I actually already merged a version of this a few days ago, so you can tell me if it works how you expect or not :)

ofseed:
I tried switching branches or cherry-picking, but I still need to execute `:e` to refresh existing buffers.

What makes me most curious is that gitsigns will refresh the existing buffer signs after performing these operations, which may mean that gitsigns also has a relatively complete "file watcher"? I don't quite understand the difficulty of this matter. The "better file change detection" on the Neovim roadmap seems to have been around for many years.

CKolkey:
Could you make a screen recording and open an issue? I cherry pick/change branches all the time and the status buffer is always right.

Also, <c-r> will refresh the buffer too :)

ofseed:
Of course.


https://github.com/NeogitOrg/neogit/assets/61115159/707b6da2-9e13-49fe-90fb-f7a296874701

CKolkey:
> Of course.
> 
>  2023-09-01.19-13-38.mp4

Ohh! You mean the _buffer_. Thats not going to be managed by neogit, but you can add an autocmd on "FocusGained", "BufEnter", "CursorHold" events that runs `silent! checktime %` to have vim check if the file has been updated.

ofseed:
Thank you! That's very useful.

Some git integrations like fugitive, could reload the buffer after these commands are executed. If you are not interested in such a feature, I might implement it when I have free time.

CKolkey:
You're welcome! If you want to, thats alright, but I'm hesitant to start messing with users open buffers. I kinda want to draw a line that starts at the status buffer, and say that everything from the status buffer onwards is ours, but everything before that is not.

PR #616, opened 2023-07-15
Test overhauls

treatybreaker:
Here's what's been done.

1. Large overhaul to the test directory structure to make it more obvious where tests and other tooling should go.
2. Some of the modules that were previously local to `git_harness.lua` have been exported
3. A `util.lua` file has been created with some basic stuff in it currently, comparing lists and exporting the project directory
4. A basic test for testing `neogit.lib.git.branch`
5. Reorganization and renaming of specs to mock the primary Neogit code organization
6. Use bare repositories to handle git clones and git repositories

PR #618, opened 2023-07-15
fix: show the correctly bound key for the top Status Hint

### LINKED ISSUE #564: Modify the hint according to user keymappings
I want to keep 's' to use leap.nvim in the neogit buffer, so i have made `a = Stage`  (mnemonic add) 

I want to modify the hint at the top of the neogit buffer to say `st[a]ge` instead. Maybe also add a few other helpful hints. I think this part can be manual, but the `?` popup should also be adjusted to include new mappings (it still says `s Stage` after the above mapping)

treatybreaker:
Prior to this PR the top hint the status buffer was hardcoded to show certain keys.

This PR checks the keymap and inserts the correct key for the top hints, unfortunately this does impact the way the keys look in the hint:
image

This PR also adds a function `get_reversed_status_map` to `config.lua` and uses that for the reversed mapping lookup.

Fixes #564 (partially)

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -152,7 +152,22 @@ local function draw_buffer()
 
   local output = LineBuffer.new()
   if not config.values.disable_hint then
-    output:append("Hint: [<tab>] toggle diff | [s]tage | [u]nstage | [x] discard | [c]ommit | [?] more help")
+    local reversed_status_map = config.get_reversed_status_maps()
+
+    local function hint_label(map_name, hint)
+      return "[" .. reversed_status_map[map_name] .. "] " .. hint
+    end
+
+    local hints = {
+      hint_label("Toggle", "toggle diff"),
+      hint_label("Stage", "stage"),
+      hint_label("Unstage", "unstage"),
+      hint_label("Discard", "discard"),
+      hint_label("CommitPopup", "commit"),
+      hint_label("HelpPopup", "help"),
+    }
+
+    output:append("Hint: " .. table.concat(hints, " | "))
     output:append("")
   end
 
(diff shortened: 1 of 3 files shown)


lua/neogit/status.lua
```diff
@@ -152,7 +152,22 @@ local function draw_buffer()
...
+      return "[" .. reversed_status_map[map_name] .. "] " .. hint
+    end
+
+    local hints = {
+      hint_label("Toggle", "toggle diff"),
+      hint_label("Stage", "stage"),
+      hint_label("Unstage", "unstage"),
+      hint_label("Discard", "discard"),
+      hint_label("CommitPopup", "commit"),
+      hint_label("HelpPopup", "HelpPopup"),
```
CKolkey, line 167: Probably should be:
```lua
hint_label("HelpPopup", "help")
```

CKolkey (changes requested):
Small tweak

lua/neogit/status.lua
treatybreaker: Oops, will fix. Thanks for catching that :sweat_smile:

treatybreaker: Done as part of the latest force push.

CKolkey:
Looks great - could you add a spec? Either extract to a function and test directly, or just load the status buffer and read the line, and expect it's what you think.

treatybreaker:
@CKolkey Would it be ok for me to create a separate PR for the spec after the status of #616 is determined (either rejected or merged)? Otherwise I get to play the merge conflict game, which I'm not overly fond of, as well as updating whatever spec I write now to utilize some of the improved functions in #616.

CKolkey:
Haha, understood. I like the approach in that PR a lot, so we'll get @ten3roberts to weigh in and (most likely) move forward with it.

PR #623, opened 2023-07-16
fix: parsing of remotes without `.git` suffix

ten3roberts:


ten3roberts:
I will also need to infer https/git, as git is now assumed for the new remote

ten3roberts:
This makes sure that changing a remote works when current remote end with `.git` and when it doesn't.

CKolkey:
Test? :)

ten3roberts:
Yes I'll add that, along with adding to the contributing guide on how to much user input

ten3roberts:
I've now added testing to the remotes, and low and behold found some more small issues related to the argument passing to the cli builder, and remote parsing. All that is fixed now :smile:

lua/neogit/lib/git/remote.lua
```diff
@@ -24,9 +24,10 @@ local function cleanup_push_variables(remote, new_name)
 end
 
 function M.add(name, url, args)
+  print("Adding remote", name, url, vim.inspect(args))
```
CKolkey, line 27: I don't think this is strictly needed (no need to make it too noisy) but if you think so, lets do a notification.

lua/neogit/popups/remote/actions.lua
```diff
@@ -9,20 +9,31 @@ local status = require("neogit.status")
...
+  local host, _, remote = origin:match("([^:/]+)[:/]([^/]+)/(.+)")
+
+  remote = remote and remote:gsub("%.git$", "")
+
+  local msg
+  if host and remote then
+    msg = string.format("%s:%s/%s.git", host, name, remote)
+  end
+
+  local remote_url = input.get_user_input("Remote url: ", msg)
```
CKolkey, line 30: This is ok if `msg` is nil?

lua/neogit/popups/remote/init.lua
```diff
@@ -8,7 +8,7 @@ function M.create()
   local p = popup
     .builder()
     :name("NeogitRemotePopup")
-    :switch("f", "f", "Fetch after add", { enabled = true, cli_prefix = "-" })
```
CKolkey, line 11: `false` would be the default (`enabled` can be removed entirely)... but to mirror Magit, I made this `true`. Any reason for the change? I rather like the behaviour as-is.

tests/util/git_harness.lua
```diff
@@ -55,13 +56,24 @@ function M.in_prepared_repo(cb)
...
+        error(err)
+      end
+    end)
   end
 end
 
+---@param cmd string[]
+---@return string[]
+local function exec(cmd)
+  local result = vim.system(cmd, { text = true }):wait()
```
CKolkey, line 71: It'll be cool to start using `vim.system` everywhere. I suppose we should put it somewhere that we require nightly, yea?

lua/neogit/popups/remote/actions.lua
ten3roberts: Yes, nil just means that no default value is provided. This occurs if there is no remote *origin*ally.

lua/neogit/popups/remote/init.lua
ten3roberts: Yes, when testing the remotes don't exist, which means fetching fails, and the remote is never added.

What would be the best way other than feedkeys("-f') to disable it in testing?

tests/util/git_harness.lua
ten3roberts: Yes, most plugin I've seen require nightly or at least a recent enough version of neovim. I think I've only seen one or two plugins that do fallbacks (heirline with `nvim_get_hl` is one of them).

This is only for testing and I figured that if you are developing neovim you are using nightly, as you'd want to use apis which you know won't be replaced in the next version, like `nvim_buf_get_option` etc.

lua/neogit/popups/remote/init.lua
CKolkey: Not sure I see the issue with feeding `-f` to disable it. I definitely would prefer that to changing the default behaviour. 

Remember we can start adding unit tests for just the actions to test these functions with more isolation. These popup specs, in my mind, definitely fall into the "integration" category, so doing more than one action in them feels fine.

tests/util/git_harness.lua
CKolkey: Oh, I'm 100% in favor of nightly - I even accidentally used that new `nvim_get_hl` without realizing it was new. Oh well. We still label this project as work-in-progress, so I'm happy to lean into that for now.

lua/neogit/popups/remote/init.lua
ten3roberts: Yes unit testing for that by splitting into the `actions.lua` is a good thing.

I purposely did two git remotes additions in the integration test to also ensure that the popup closes and resets correctly.

There is a balance to be found. Larger tests make it easier to find bugs with residual states, especially using async as you could have a semaphore never unlocking. Smaller tests miss out on that, bug make it harder to tell what *exactly* failed by looking at the test failures.

lua/neogit/lib/git/remote.lua
ten3roberts: Thought I cleared those out. Couldn't reliably get the logging for the tests to show up in the right order when all were run. Will need to expand the testing `init.lua` to support running just 1 test

CKolkey: _agreed_, thats a pain point for me too.

ten3roberts: `NEOGIT_LOG_CONSOLE=sync` works somewhat better, as it uses `nvim_writeln` most of the time rather than a `vim.schedule(write)`, but may defer a write to the next frame if in a fast event.

The other issue is a way to run a single test easier.

`:PlenaryBustedFile %` does not work for me and ends up breaking my editor by making the test output lines interleave with the Neovim ui itself, causing a "staggering" for each printed terminal line. Don't know if you've had the same experience.

The `recipes.json` is a custom plugin I am using and is a crude way to solve it, though does not work with the new `tests/init.lua` which runs all tests anyway.

I think I'll wait for treaty breaker's minimal init and see where that goes, and then add the testing shim ontop of that. Having the same `init.lua` during testing *and* issue reproduction is golden.

tests/util/git_harness.lua
treatybreaker: Wait wait wait! I was just browsing, but just now noticed this discussion.

It's ok to use `nightly` functionality? We are only interested in `nightly` support?

If so, this makes life *much* easier.

CKolkey: Hahaha, I mean, it wasn't an intentional thing - I sorta goofed but then nobody seemed to complain. I'd still, for the most part, like to avoid things that aren't in a stable release yet, as anything in nightly is subject to change without notice. So, it's kind of a case-by-case thing. I'd definitely be open to considering using new API's.

treatybreaker: Damn, I was getting super hyped about going completely to nightly functions. Raining on my parade 😠.

I know that in nightly a *ton* of stuff is deprecated or just straight up removed. So it makes sense to stick to stable for most things on that note.

CKolkey: Hahaha, yeah, sorry about that ⛈️  Once it hits stable then aggressive refactoring are most welcome ;)

lua/neogit/lib/git/remote.lua
```diff
@@ -75,4 +75,8 @@ function M.list()
   return cli.remote.call_sync():trim().stdout
 end
 
+function M.get_url(name)
+  return cli.remote.get_url(name).call():trim().stdout
```
CKolkey, line 79: Shouldn't this return the string, not a table? so `stdout[1]`?

CKolkey (approved):

PR #625, opened 2023-07-16
test(feat): add tests for reversed config map function

treatybreaker:
This adds the tests promised for #618


treatybreaker:
I have found several bugs with how mappings are set and managed... **pain**, *withering*, ***pain***.

The real solution is to basically redo how mappings are managed otherwise duplicate mappings make it virtually impossible to tell when we have the correct map. I'm going to begin doing a potential breaking change refactor for maps, hopefully it isn't breaking, but I'm not too confident.

CKolkey:
I've kinda known in the back of my head that this would happen at some point. Just make sure the new system is tested 😇

treatybreaker:
Closing this and will open a new PR for this once (if) #629 gets merged

PR #629, opened 2023-07-17
refactor!: improve mapping setup

treatybreaker:
***This is a breaking change to the mapping setup.***

This makes it easier to define which command goes to a key. Prior to this you'd have to set a key to empty like so: `["s"] = ""` then set another key to change the binding.

Now the binding is set to `Command -> Key` instead of `Key -> Command`. This also fixes some bugs in which a user sets two bindings to empty to remove them which currently causes neogit to crash.

Something contentious I chose in this PR, the status mappings may *only* map to a single key whereas the finder may map to any number of keys. I believe the finder has a small enough footprint where duplicate mappings are unlikely to occur, but the status maps are large enough where duplicates are likely. Furthermore, duplicate keys for status is really unnecessary.

I have added validation logic to check `M.values` in `config.lua` and ensure the config is correct. This will allow this breaking change to go through without causing scary, difficult to identify errors if a user was utilizing the previous mapping options.

### What has been done bullet points:
- Change mapping to be `Command -> Key` instead of `Key -> Command`
  - This makes it easy to get which command it bound to what key(s) and fixes a crash when a user tries to disable two bindings by setting their commands to empty.
- Allow Status mappings to only map to ***one*** key
- Comprehensive testing of `neogit/config.lua`
- Added a comprehensive validation function for `neogit/config.lua`
  - That will make this breaking change less painful for folks who were setting keybinds as     the errors emitted by the validation function make this all quite obvious
- Removed the `use_magit_keybindings_option`
- Updated the README to reflect the mapping changes

This makes the PR #625 possible.


treatybreaker:
@CKolkey @ten3roberts 

Take a look at what I have here currently. I have written basically 100% test coverage for the Neogit config and a 100% validation of it via a validation function.

I did that so if this breaking change is accepted it should be *much* easier to debug for folks who were using the previous mapping setup.

I still need to write tests to validate the mappings are correctly bound. Will do tomorrow.

CKolkey:
I'll be a little slow to review things as family is in town, but I don't have a major objection. 

As a note, I'd like to entirely remove the 'use magit keybindings' option. For one, we hardly utilise it. For two, its meaningless as DOOM emacs will have different mappings from normal emacs.. and users can customise the bindings however they like anyhow. If this is going to be a breaking change, lets make the most of it.

Another thing - the log view and reflog view can open the cherry pick and revert popups - that should follow the users bindings.

Finally, making this easy to debug for users is a top goal, so I'm glad thats something you already considered :)

tests/specs/neogit/config_spec.lua
ten3roberts, line 1: This testing is really good, though I am ever so slightly concerned with the *number* of them, as they all run in serial and may induce much more noise in the test run.

I personally think that a single (or a handful) of tests which test a fully invalid config, and checks all the errors, rather than one test for one error.

CKolkey: IMO lua is fast enough to not worry about the number of tests, at least for now. If we have a test suite thats so slow that its a problem, I'll consider that a good position to be in :)

ten3roberts: Oh perfect then.

Thanks a lot for kind efforts of creating the *very* comprehensive tests.

Our goal is to test more and more of neogit to help catch bugs before they go out in the world, and this is a leap in that direction and sets up a good example for others.

👍

treatybreaker:
@CKolkey @ten3roberts 

Resolved the merge conflicts.

I'm opening this up for review, reference the PR description for the modifications done. Let me know what you'd like changed or needs to be done.

lua/neogit/config.lua
```diff
@@ -1,157 +1,493 @@
...
-      local ok, value = pcall(require, key)
-      return ok and value or false
+function M.get_default_values()
+  return {
+    disable_hint = false,
+    disable_context_highlighting = false,
+    disable_signs = false,
+    disable_commit_confirmation = false,
+    disable_builtin_notifications = false,
+    use_telescope = false,
```
CKolkey, line 14: `use_telescope` - This line got removed earlier today

lua/neogit/config.lua
```diff
@@ -1,157 +1,493 @@
...
+      return false
+    end
+    return true
+  end
+
+  -- More complex validation functions go below
+  local function validate_kind(val, name)
+    if
+      validate_type(val, name, "string")
+      and not vim.tbl_contains({ "split", "vsplit", "tab", "floating" }, val)
```
CKolkey, line 211: This can also be "auto"

lua/neogit/config.lua
```diff
@@ -1,157 +1,493 @@
...
+      if config.disable_insert_on_commit ~= "auto" then
+        err(err_msg)
+      end
+    elseif type(config.disable_insert_on_commit) ~= "boolean" then
+      err(err_msg)
+    end
+  end
+
+  local function validate_integrations()
+    local valid_integrations = { "diffview" }
```
CKolkey, line 271: "telescope" is a valid integration

lua/neogit/config.lua
```diff
@@ -1,157 +1,493 @@
...
+    end
+
+    if validate_type(config.mappings.finder, "mappings.finder", "table") then
+      for command, key_binds in pairs(config.mappings.finder) do
+        validate_finder_map(command, key_binds)
+      end
+    end
+
+    -- Validate mappings.status
+    local valid_status_commands = {
```
CKolkey, line 366: Probably would be a good idea to derive this list somehow so adding new commands/popups doesn't require a stop here

lua/neogit/config.lua
```diff
@@ -1,157 +1,493 @@
...
+    end
+  end
+
+  if validate_type(config, "base config", "table") then
+    validate_type(config.disable_hint, "disable_hint", "boolean")
+    validate_type(config.disable_context_highlighting, "disable_context_highlighting", "boolean")
+    validate_type(config.disable_signs, "disable_signs", "boolean")
+    validate_type(config.disable_commit_confirmation, "disable_commit_confirmation", "boolean")
+    validate_type(config.disable_builtin_notifications, "disable_builtin_notifications", "boolean")
+    validate_type(config.use_telescope, "use_telescope", "boolean")
```
CKolkey, line 432: `use_telescope` - This got removed today

CKolkey (changes requested):

lua/neogit/popups/help/init.lua
```diff
@@ -4,7 +4,7 @@ local actions = require("neogit.popups.help.actions")
 local M = {}
 
 -- TODO: Better alignment for labels, keys
-function M.create(env)
+function M.create()
```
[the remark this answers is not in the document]
treatybreaker: Oops! Remnant from a different branch. Fixed!

lua/neogit/config.lua
treatybreaker: Removed from my branch now.

lua/neogit/config.lua
treatybreaker: Fixed.

lua/neogit/config.lua
treatybreaker: Fixed.

lua/neogit/config.lua
treatybreaker: Done, also done for finder commands.

lua/neogit/config.lua
treatybreaker: Removed it from my branch, including the tests.

treatybreaker:
Actually before this gets merged @CKolkey or @ten3roberts do we have a issue for tracking breaking changes? If not it might be wise to make a pinned issue with them so if folks start getting errors they can easily see what changed.

CKolkey:
We've been hitting the breaking changes pretty hard lately, so.. yeah, let's get that set up. I'll review this tomorrow at some point.

treatybreaker:
Hey @CKolkey, just checking to see where this is at. Is there anything extra you'd like me to do for this PR?

lua/neogit/config.lua
```diff
@@ -1,157 +1,457 @@
...
+      },
+    },
+  }
+end
+
+M.values = M.get_default_values()
+
+---Validates the config
+---@return boolean false if the config is invalid
+---@return string all error messages emitted during validation
```
ten3roberts, line 158: You previously mentioned that it was hard to parse and programmatically use the output of the validate function in tests, especially if there are multiple errors.

We could strike *three* birds with one stone by making this return a *list* of errors. The current method, by returning a boolean for success, and a pretty human readable concatenated message for all the errors leads to some ambiguities. Given the function signature, it is entirely possible to return both `true` and errors at the same time.  Looking at the function from the outside. there would be no way to tell it could not do that, as it is entirely possible given the definition space, in which the output would be very malformed.

By returning a single list of strings, the success status of the validation becomes entirely self-evident by an empty list and it becomes impossible to end up in the trinary "on *and* off" state.

The header message could be added at the call site in setup as it is not really part of the validation. This way, the validation function becomes data oriented withing a simple well-defined space. There would be no ambiguities, and parsing and checking the validation message in the tests would be straight forward, one list element per error. This way, you could also validate that the *correct* error, and the right amount of them are emitted if you want to be pedantic about the tests.

This style of *constraining using definition* something that I've learned from use of Rust and Haskell and has worked wonders for large interconnected systems, just sharing my advice.

ten3roberts (changes requested):
The validation of the config is a great additon to the config, and makes error spotting by users much easier.

Left some comment on some ways to deal with the validation specifics; let me know what you think.

lua/neogit/config.lua
```diff
@@ -1,157 +1,457 @@
...
+      finder = {
+        ["Select"] = { "<cr>" },
+        ["Close"] = { "<c-c>", "<esc>" },
+        ["Next"] = { "<c-n>", "<down>" },
+        ["Previous"] = { "<c-p>", "<up>" },
+        ["MultiselectToggleNext"] = { "<tab>" },
+        ["MultiselectTogglePrevious"] = { "<s-tab>" },
+        ["NOP"] = { "<c-j>" },
+      },
+      status = {
```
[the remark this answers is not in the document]
treatybreaker: Setting to false doesn't work for the reverse lookup unless we do some preparsing to remove the key from the table entirely.

The primary issue comes when they have multiple mappings set to a command, in the reverse lookups we'll still fail to get the reversed map, for instance setup like so:

```lua
neogit.setup({
    mappings = {
        status = {
            ["s"] = "Status",
            ["b"] = "Status"
        }
    }
})
```

That will error, and so will setting them to false. We could check if a command has multiple maps and reject it in the validation, but we still don't have a way of disabling the mapping unless we do some preprocessing to remove keys that are set to false prior to getting the reverse map.

Furthermore, a user no longer has to disable default maps with the `command -> key` setup, by setting their new mapping they necessarily disable the default map.

I can do the preprocessing and whatnot if you'd prefer to keep it `key -> command`, but I think there's less extra code and processing done with `command -> key`. Up to you, just let me know.

lua/neogit/config.lua
```diff
@@ -1,157 +1,457 @@
...
+        end
+      end
+    end
+
+    validate_signs_table("hunk", config.signs.hunk)
+    validate_signs_table("item", config.signs.item)
+    validate_signs_table("section", config.signs.section)
+  end
+
+  local function validate_disable_insert_on_commit()
```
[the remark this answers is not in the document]
treatybreaker: What would be the preferred scope? These are intended for the config mostly, not sure what else this code would cover unless you're intending on doing validation before each sign 

LMK, happy to move it.

lua/neogit/config.lua
treatybreaker: No you're right, when I'm in scripting languages I have a habit of kludging things together. "Proper" programming languages tend to lend themselves to making this type of mistake more difficult. Rust especially makes it obvious on domain whereas Lua, at least for me, not so much.

Will be fixed here shortly. The only thing I'm thinking of doing is returning a key value store of them with the key containing the config value that is bad, say `signs.hunk` with the error message being its value. Might make it a bit quicker to test and validate.

lua/neogit/config.lua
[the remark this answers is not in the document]
ten3roberts: I see. Given my understadning, the reverse lookup is mainly for user displaying purposes, and not functionality related?

Just throwing things at the wall to see what sticks and what doesn't

https://github.com/NeogitOrg/neogit/pull/659

Would that one solve the reverse lookup missing functionality?

lua/neogit/config.lua
ten3roberts: Oooh :eyes: 

Agreed, Rust is phenomenal at this.

```rust 
fn validate(config: Config) -> BTreeMap<ConfigKey, String>
```
 as a Rustacean would see it, love it!

Lua being lua though we'd need to remember to use `vim.tbl_count` as `#value` only considers integer keys

treatybreaker: > as #value only considers integer keys

I didn't know that... 👀. I have a bunch of stuff then that's uhhh, having a not so good time lmao.

lua/neogit/config.lua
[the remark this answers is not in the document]
treatybreaker: It solves the reverse lookup presentation, but not the binding issues. We'd still have to do some preprocessing :/.

lua/neogit/config.lua
ten3roberts: Oh no, seems like I poked the sleeping beast.

```vim
lua assert(#{a = 5}, 0)
```

This passes, lua truly is strange, and the list handling is *odd*

lua/neogit/config.lua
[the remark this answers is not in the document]
ten3roberts: Yes unfortunately, We'd need a `filter` or just an `if` in that for loop as well.

How does the unbinding work with your changes?

lua/neogit/config.lua
treatybreaker: That's cursed. Proof that we should be using elisp instead of lua. Anyhow, currently working on getting a `key -> { error_messages }` return going.

treatybreaker: Done @ten3roberts. Take a look and let me know on this specific change. I am still going to make modifications relevant to https://github.com/NeogitOrg/neogit/pull/629#discussion_r1271098487.

Of particular interest is how I'm doing the formatting now within the neogit setup function. You may want me to change that be a separate function that takes in those config errors and then outputs the string for error presentation. Just let me know.

lua/neogit/config.lua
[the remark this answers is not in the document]
treatybreaker: Well you're directly setting the keybinds for the maps like so:

```lua
["Status"] = "s"
```

To unbind `Status` entirely you'd just do:

```lua
["Status"] = ""
```

And since the commands are the key in my version there's not really a need to do a reversed map as far as I know, avoiding the whole duplicate key stuff we have right now.

lua/neogit/config.lua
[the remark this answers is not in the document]
treatybreaker: Moving this discussion to your lower comment -> https://github.com/NeogitOrg/neogit/pull/629#discussion_r1271098487

lua/neogit/config.lua
```diff
@@ -1,157 +1,457 @@
...
+    validate_type(config.auto_refresh, "auto_refresh", "boolean")
+    validate_type(config.sort_branches, "sort_branches", "string")
+    validate_type(config.console_timeout, "console_timeout", "number")
+    validate_kind(config.kind)
+    validate_type(config.auto_show_console, "auto_show_console", "boolean")
+    if validate_type(config.status, "status", "table") then
+      validate_type(config.status.recent_commit_count, "status.recent_commit_count", "number")
+    end
+    validate_signs()
+    validate_disable_insert_on_commit()
```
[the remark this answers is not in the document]
treatybreaker: Take a look at [this commit](https://github.com/NeogitOrg/neogit/pull/629/commits/e6660a2c0077291e87503cfe4835b9f403845b31), this should do what you're looking for.

lua/neogit/config.lua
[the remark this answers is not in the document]
CKolkey: So, while I do see the issues with having `["s"] = "Stage"`, I do think that doing the "expected" thing has a lot of value, which in this case is to stick to the current mapping schema. Also, for the longest time, I've had both `h` and `l` bound to `Toggle`, to treat folds like a file-tree, going "up" a level and "down" a level. 

I'm not very concerned about users unbinding keys.. Honestly, I think of neogit as an app within an app, so, while it's "in" nvim, overloading common things like "v" is fair game. 

I think the most compelling reason to adopt this change is to use the *correct* bindings in all locations. The log view, commit view, and reflog view can all just use `config.mapping.cherry_pick` and get the right thing.

So, I'm in favor (mostly) for that reason.

lua/neogit.lua
```diff
@@ -31,16 +31,39 @@ local setup = function(opts)
...
+    error(
+      string.format(
+        "\n%s\n%s\n%s\n%s",
+        header,
+        table.concat(header_message, "\n"),
+        header_sep,
+        table.concat(config_errs_message, "\n")
+      ),
+      vim.log.levels.ERROR
+    )
```
CKolkey, line 60: Following the `module.setup()` pattern, can this get extracted to `config.setup()`? I don't think the validation logic needs to live outside the module it applies to.

lua/neogit/status.lua
```diff
@@ -152,10 +152,11 @@ local function draw_buffer()
 
   local output = LineBuffer.new()
   if not config.values.disable_hint then
-    local reversed_status_map = config.get_reversed_status_maps()
+    local status_map = config.values.mappings.status
 
     local function hint_label(map_name, hint)
-      return "[" .. reversed_status_map[map_name] .. "] " .. hint
+      local key = status_map[map_name]
+      return "[" .. key .. "] " .. hint
```
CKolkey, line 159:
```suggestion
      return string.format("[%s] %s", status_map[map_name], hint)
```

lua/neogit/popups/help/actions.lua
```diff
@@ -4,7 +4,7 @@ local util = require("neogit.lib.util")
 local NONE = function() end
 
 -- Using deep extend this way creates a copy of the mapping values
```
CKolkey, line 6: I don't think this comment applies anymore

tests/specs/neogit/config_spec.lua
```diff
@@ -0,0 +1,544 @@
+local config = require("neogit.config")
+
+describe("Neogit config", function()
+  before_each(function()
+    config.values = config.get_default_values()
+  end)
```
CKolkey, line 6: Can you add a newline between `6` & `7`?

CKolkey (changes requested):

lua/neogit.lua
treatybreaker: Yeah, that's fair. I'll move it :slightly_smiling_face:

lua/neogit/popups/help/actions.lua
treatybreaker: Fair, will remove.

tests/specs/neogit/config_spec.lua
treatybreaker: Stylua is removing the newline as a heads up. Do you want me to add a stylua ignore?

lua/neogit/config.lua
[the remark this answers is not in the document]
treatybreaker: If y'all don't think the mapping changes here are worth it, then what I can do is extract all the config validation and tests to a separate PR and we can look at that alone there.

In terms of the mappings, if we don't want to go with my changes I guess we can do this:

If a user sets a default mapping to false then we remove the key entirely from the table. Secondly, if they have the same command mapped to multiple keys we can do a.) fail validation or b.) rework some of the internals to better handle these multiple mappings (such as not using reversed tables and instead searching for all commands(?)).

tests/specs/neogit/config_spec.lua
CKolkey: Eh, nah. I don't always agree with stylua, but it's not usually worth it to fight. Not _this_ time.

treatybreaker:
Ok, after playing around a bit with #659 I think that might be a superior way to go. Though for the hint mappings maybe we'd show at minimum 2 keys instead of just one.

@CKolkey and @ten3roberts if y'all are ok with this I'd like to promote @ten3roberts solution as it isn't a breaking change and I can move all of these config validation and tests into a separate PR, dropping my mapping refactor. I think the solution presented in #659 is just straight up superior.

lua/neogit/config.lua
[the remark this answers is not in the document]
ten3roberts: Yes I am not sure either, which is why I preemptively opened a PR (#659) that reworks the internals for multi mappings which makes the hints and everything work. I know @CKolkey uses two mappings for one action, and I think others may be as well, such as `=` for staging. If possible, it would thus be good to keep it supported, removing functionality is always a bit iffy, especially if there is no workaround.

I have absolutely no objections against mucking about in the internals to make it work though.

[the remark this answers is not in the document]
treatybreaker: I think you're patch over there at #659 is what we should use instead of this as it preserves functionality and addresses all of the bugs known in other issues. I already have a branch based on it for the config validation.

treatybreaker:
Closing this in favor of #659, I have already rebased all the config validation and tests for the config to that branch. That's the major changes in this PR moved.

PR #631, opened 2023-07-17
Add annotations for the Neogit config

treatybreaker:
This is being set to a draft until the status of #629 is determined.

This adds annotations to the Neogit config such that it's easy to see values and their types from within an editor running the lua language server.


ten3roberts:
Great, I am a big fan of strong typing and thinks it helps to think in types and what they allow

PR #632, opened 2023-07-18
Provide full installation example for Lazy distro

Gleland:
Hello! Thanks for making this awesome plugin. I added a short change to make the `Lazy` install a bit more copy-paste. Thanks for your time!


CKolkey:
I think you can also just pass `opts = {}` instead of a function to config.

vheon:
> I think you can also just pass `opts = {}` instead of a function to config.

when you just want to run the setup one can simply `config = true`

Gleland:
>when you just want to run the setup one can simply config = true

even better! Thanks for the feedback, I've just updated it.

ten3roberts (approved):

ten3roberts:
Thanks, looks great. Always good to mention the plugin manager that a lot of people, including me use

PR #636, opened 2023-07-18
Breaking: Change default for RevertPopup from "_" (doom emacs) to "v"

CKolkey:
This kinda kills two birds with one stone. Users get confused when selections don't work, because they are using REGULAR visual mode instead of LINEWISE visual. So, bind `Revert` to `v` (which is closer to magit: https://magit.vc/manual/2.90.0/magit/Reverting.html ).


ten3roberts (approved):
Nice.

Makes me think of the `-v` flag for grep and other unix tools that *invert* or *revert* makes it consistent with this

CKolkey:
I hadn't considered that, haha - makes a lot more sense now.

PR #638, opened 2023-07-18
fixup: test setup for macos

CKolkey:
@treatybreaker I had some issues with the test setup running locally on my macbook - made some tweaks. Everything seems to be in good order here, but let me know if these changes broke something for you or @ten3roberts

### DIFF
--- a/tests/util/util.lua
+++ b/tests/util/util.lua
@@ -27,22 +27,25 @@ end
 
 M.neogit_test_base_dir = "/tmp/neogit-testing/"
 
+local function is_macos()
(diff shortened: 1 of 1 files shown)


treatybreaker:
@CKolkey Works just fine for me over here on Arch Linux. Seems it works in the CI which is Ubuntu as well. I imagine it works just fine for any *nix OS or Mac now.

CKolkey:
@wrightjjw, wanna try running `make test` on this branch and tell me if it works? ;)

CKolkey:
I always though it was styled "Arch Linux btw"...

PR #640, opened 2023-07-18
fix: properly access the current git branch via an accessor

### LINKED ISSUE #639: Opening pull popup fails unless status was opened prior
### Description

Continuation of discussion started in https://github.com/NeogitOrg/neogit/issues/637#issuecomment-1640751018.

Users were previously able to call all popups without having opened status buffer before, it has changed in one of the recent commits.

I am loving this minimal reproducer btw!

### Neovim version

NVIM v0.10.0-dev-682+gd0d132fbd-Homebrew
Build type: Release
LuaJIT 2.1.0-beta3

macOS

### Steps to reproduce

1. nvim --clean -u minimal.lua

(I have added offender there)

### Expected behavior

Show pull popup without any errors.

### Actual behavior

```
E5113: Error while calling lua chunk: /tmp/nvim_issue/neogit//lua/neogit/popups/pull/init.lua:26: attempt to
 concatenate local 'current' (a nil value)
stack traceback:
        /tmp/nvim_issue/neogit//lua/neogit/popups/pull/init.lua:26: in function 'c'
        /tmp/nvim_issue/neogit//lua/neogit/popups/init.lua:8: in function 'f'
        /tmp/nvim_issue/neogit//lua/neogit/popups/init.lua:17: in function </tmp/nvim_issue/neogit//lua/neog
it/popups/init.lua:11>
        /Users/gegoune/repos/dots/nvim/minimal.init.lua:13: in main chunk
```

### Minimal config

```Lua
for name, url in pairs {
  neogit = 'https://github.com/NeogitOrg/neogit',
  plenary = 'https://github.com/nvim-lua/plenary.nvim',
  telescope = 'https://github.com/nvim-telescope/telescope.nvim',
} do
  local install_path = vim.fn.fnamemodify('/tmp/nvim_issue/' .. name, ':p')
  if vim.fn.isdirectory(install_path) == 0 then vim.fn.system { 'git', 'clone', '--depth=1', url, install_path } end
end

require('neogit').setup()
print 'Minimal Config Ready!'
require('neogit.popups').open 'pull'()
```

treatybreaker:
This makes popups that were being called by `require("neogit.popups").call("POPUP_HERE")()` work correctly prior to the Neogit status being opened.

This PR will cause an early return when neither a head branch or a current branch is detected to avoid any errors.

Fixes #639

### DIFF
--- a/lua/neogit/popups/branch/init.lua
+++ b/lua/neogit/popups/branch/init.lua
@@ -6,7 +6,7 @@ local actions = require("neogit.popups.branch.actions")
 local config_actions = require("neogit.popups.branch_config.actions")
 
 function M.create()
-  local current_branch = git.repo.head.branch or ""
+  local current_branch = git.branch.current()
   local show_config = current_branch ~= "" and current_branch ~= "(detached)"
   local pull_rebase_entry = git.config.get("pull.rebase")
   local pull_rebase = pull_rebase_entry:is_set() and pull_rebase_entry.value or "false"
--- a/lua/neogit/popups/branch_config/init.lua
+++ b/lua/neogit/popups/branch_config/init.lua
@@ -5,7 +5,7 @@ local git = require("neogit.lib.git")
 local actions = require("neogit.popups.branch_config.actions")
 
 function M.create(branch)
-  branch = branch or git.repo.head.branch
+  branch = branch or git.branch.current()
   local g_pull_rebase = git.config.get_global("pull.rebase")
   local pull_rebase_entry = git.config.get("pull.rebase")
   local pull_rebase = pull_rebase_entry:is_set() and pull_rebase_entry.value or "false"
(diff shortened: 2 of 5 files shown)


CKolkey:
`git.repo.branch.head or git.branch.current()` is a bit repetitive, since the former is called within the latter. Just call `git.branch.current()` and lets make a note to avoid accessing repo state directly, and instead use accessors in the correct git lib

CKolkey:
I also think the early return is wrong, unless it works with a a detached head state

treatybreaker:
I took that as a precaution, but I'll remove it. As far as I can tell it's basically impossible to reach the early return.

treatybreaker:
@CKolkey Should be good now.

treatybreaker:
> git.repo.branch.head or git.branch.current() is a bit repetitive, since the former is called within the latter. Just call git.branch.current()

Yeah, I'd have noticed that if I wasn't in a rush. I committed that, saw that it worked, and instantly went and did the laundry. Probably should just slow down and validate a bit :P.

CKolkey:
We need self washing pants. Thats the real problem here.

PR #641, opened 2023-07-19
Perf: Speed-up buffer updates

CKolkey:
Speed-up redrawing by using buffers for lines/highlights/signs/extmarks and only flush buffers once. Test this by opening the `LogView` and toggling open and closed a commit (via `<tab>`). On master... slow. On this branch, faster.

before:
Screenshot 2023-07-19 at 09 44 35

after:
Screenshot 2023-07-19 at 10 03 50

I'm not sure if I can get the `add_highlight` call to be any faster, but thats the next thing to try.

### DIFF
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -25,6 +25,10 @@ function Buffer:new(handle)
     border = nil,
     mmanager = mappings_manager.new(handle),
     kind = nil, -- how the buffer was opened. For more information look at the create function
+    line_buffer = {},
+    hl_buffer = {},
+    sign_buffer = {},
+    ext_buffer = {},
   }
 
   this.ui = Ui.new(this)
@@ -81,6 +85,46 @@ end
 
 function Buffer:set_lines(first, last, strict, lines)
   api.nvim_buf_set_lines(self.handle, first, last, strict, lines)
+end
+
+function Buffer:set_line_buffer(line)
+  table.insert(self.line_buffer, line)
+end
+
+function Buffer:add_highlight_buffer(...)
+  table.insert(self.hl_buffer, { ... })
+end
+
+function Buffer:place_sign_buffer(...)
+  table.insert(self.sign_buffer, { ... })
+end
+
+function Buffer:set_extmark_buffer(...)
+  table.insert(self.ext_buffer, { ... })
+end
+
+function Buffer:resize(length)
+  api.nvim_buf_set_lines(self.handle, length, -1, false, {})
+end
+
+function Buffer:flush_buffers()
+  api.nvim_buf_set_lines(self.handle, 0, -1, false, self.line_buffer)
+  self.line_buffer = {}
+
+  for _, sign in ipairs(self.sign_buffer) do
+    self:place_sign(unpack(sign))
+  end
+  self.sign_buffer = {}
+
+  for _, hl in ipairs(self.hl_buffer) do
+    self:add_highlight(unpack(hl))
+  end
+  self.hl_buffer = {}
+
+  for _, ext in ipairs(self.ext_buffer) do
+    self:set_extmark(unpack(ext))
+  end
+  self.ext_buffer = {}
 end
 
 function Buffer:set_text(first_line, last_line, first_col, last_col, lines)
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -25,6 +25,7 @@ function Buffer:new(handle)
     border = nil,
     mmanager = mappings_manager.new(handle),
     kind = nil, -- how the buffer was opened. For more information look at the create function
+    namespace = api.nvim_create_namespace("neogit-buffer-" .. handle),
     line_buffer = {},
     hl_buffer = {},
     sign_buffer = {},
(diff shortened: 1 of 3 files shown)


ten3roberts:
How are you producing the flamegraph 👀

lua/neogit/lib/buffer.lua
```diff
@@ -83,6 +88,48 @@ function Buffer:set_lines(first, last, strict, lines)
   api.nvim_buf_set_lines(self.handle, first, last, strict, lines)
 end
 
+function Buffer:set_line_buffer(line)
```
ten3roberts, line 91: I suppose this means *buffering*/*deferred*, and not related to a vim buffer/file?

If so, perhaps `buffered_set_line` would make the distiction clearer

lua/neogit/lib/ui/init.lua
```diff
@@ -220,73 +224,94 @@ function Ui:_render(first_line, first_col, parent, components, flags)
...
-          local padding_left = c:get_padding_left()
-          local text = padding_left .. c.value
-          self.buf:set_lines(curr_line - 1, curr_line, false, { text })
-          if highlight then
-            self.buf:add_highlight(curr_line - 1, c.position.col_start, c.position.col_end, highlight, 0)
-          end
-          if sign then
-            self.buf:place_sign(curr_line, sign, "hl")
+          if not flags.hidden then
+            self.buf:set_line_buffer(table.concat { c:get_padding_left(), c.value })
```
ten3roberts, line 263: How about just making the existing immediate buffered functions deprecated, and just using buffered writing.

I see no downside functionality wise you if are willing to take the time and make this PR slightly larger.

In C printf fprintf etc are *always* buffered, and I think that providing an unbuffered *and* buffered API is error prone as you can do a buffered write, and another function deeper does an immediate write later. The end result is that you get the immeadiate write followed by the buffered one, I.e `A,B` becomes `B,A` and would be very nasty to have and find in weird, hard to reproduce edge cases.

Removing the access to simultatenous use of buffers and non buffered solves this.

*Rust* makes it deliberately hard to perform immediate writes when you have a buffered writer, precisely because of the aforementioned re-ordering, see https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.get_mut

ten3roberts (changes requested):

lua/neogit/lib/buffer.lua
CKolkey: Yeah, the wording is weird. I'll change it.

lua/neogit/lib/ui/init.lua
CKolkey: Yea, I'm open to ideas here. This is mostly a POC I came up with last night. I was thinking maybe some debounced thing, where after 100ms of not getting new lines, auto-flushing the buffers. Not sure if thats getting too crazy though.

Not providing two different ways to write to the vim-buffer is a good call. So, keep the existing names, and just internally buffer the writes.. Ideally with auto flushing. What do you think about that?

CKolkey:
> How are you producing the flamegraph? 👀

So, in `init.lua`, at the bottom:

```lua
    if os.getenv("PROFILE") then
      require("plenary.profile").start("profile.log", { flame = true })
      vim.api.nvim_create_autocmd("VimLeavePre", { callback = require("plenary.profile").stop })
    end
```

so, run `PROFILE=1 nvim` and exercise the code path.

The, with https://github.com/jonhoo/inferno installed:
```fish
inferno-flamegraph profile.log > profile.svg && open profile.svg
```

The callstacks aren't perfect, but are definitely good enough to make decisions with.

lua/neogit/lib/ui/init.lua
ten3roberts: Not entirely sure about the auto-buffering. I consider that forgetting to flush is an implementation error, and auto flushing after 100ms could lead to the flushing happening at a bad time or when you can't write to a buffer.

I would just add `flush` at the bottom for each of the build-up and display functions

CKolkey: That's fair. I'll ditch that thought 😉

CKolkey: Been thinking about if this should be the default behaviour or not - I think having an explicit buffered API is better, so callers know they need to flush said buffer for anything to happen. Otherwise you see `Buffer:set_line` and it does nothing. Does that make sense?

lua/neogit/lib/buffer.lua
```diff
@@ -242,7 +289,7 @@ function Buffer:set_foldlevel(level)
 end
 
 function Buffer:replace_content_with(lines)
```
ten3roberts, line 291: How about making these *immediate* functions error if the buffer is not flushed?

That would be a implementation bug if not flushed, and will lead to text appearing despite the buffer being replaced. What do you think

ten3roberts (approved):

ten3roberts:
I suppose this may break something, but I don't consider `Buffer` part of the public api until we split out the UI library

lua/neogit/lib/buffer.lua
CKolkey: So something like this?

```lua
function Buffer:set_line(...)
  if #self.line_buffer > 0 then
    error()
  end

  ...
end
```

CKolkey:
Yeah, I wouldn't say `Buffer` is public.

PR #648, opened 2023-07-19
chore: remove doc/tags

ten3roberts:
This fixes the error when updating Neogit through an auto-indexing plugin manager such as lazy. See https://github.com/NeogitOrg/neogit/discussions/628#discussioncomment-6492754

For the users they may need to once more do the `git restore doc/tags` dance as `docs/tags` would be in the state of `modified by us, deleted by them` or `UD`


ten3roberts:
```vim
:Lazy sync
```

image

Caused by going from `master` to the `remove-tags` branch in plugin spec and updating (after removing local dev overrides)

ten3roberts:
This will be a breaking change, so I added the requested tracking issue for those

ten3roberts:
@CKolkey what do you think about this breaking change?

`auto-session` had to do a similar breaking change a few months ago

CKolkey:
I'm totally fine with this, but I'm not sure how this is a breaking change. What will users need to do?

CKolkey (approved):

ten3roberts:
The breaking commit is that users may need to go and explicitely restore the repo inside `~/.local/share/nvim/lazy/neogit` as lazy will fail to pull as `tags` may be changed locally due to the aforementioned auto-indexing

PR #655, opened 2023-07-21
Perf: use string.format for a minor performance gain

CKolkey:
String concatenation is expensive in LUA since all strings are interned, so they need to be rehashed every single time it's mutated. String.format + table.concat is a better way to do this.


ten3roberts (approved):
Yes. Big fan of formatting over concatenation in all languages myself :+1:

Issue #657, opened 2023-07-21
Disable "telescope integration not enabled" error message.


ten3roberts:
### Discussed in https://github.com/NeogitOrg/neogit/discussions/656

<div type='discussions-op-text'>

<sup>Originally posted by **liujoey** July 21, 2023</sup>
I'm using `fzf-lua` as my main fuzzy finder, I don't have telescope because it is too slow for my large project. And I use `stevearc/dressing.nvim` to set `fzf-lua` as a replacement of `vim.ui.select`, but neogit just doesn't like that idea, it constantly popup error messages telling me that telescope integration was disabled. Is there a configuration that I can disable this error message? Thank you.</div>

PR #658, opened 2023-07-21
fix: #657 don't show error message for optional integrations

ten3roberts:


lua/neogit/buffers/log_view/init.lua
```diff
@@ -122,7 +122,7 @@ function M:open()
           end
         end,
         ["d"] = function()
-          if not config.ensure_integration("diffview") then
```
CKolkey, line 125: instead of a boolean param, can we do an opts table, like `{ required = x }`? Not a huge fan of naked boolean params

ten3roberts: Yes. Not entirely sure about the existing name for the function either.

`ensure_`, at least the way I've encountered and used the name refers to a function which does a check and in some way *ensures* a value, such as generating one if it doesn't exist, ensures a column is *present* in a database etc.

This is essentially a query or check function, especially when adding a `{ required = true }`

Perhaps `query_integration(name) -> bool` and the have the caller do an error or fallback, with better context, such as `diffview integration needs to be enabled for "diff head"`

CKolkey: I like the idea of leaving it up to the caller. That removes the need for the `{ required = true }` param altogether, and... wait for it... we could call it... `check_integration(name)`. Because it's a check.. for the integration. :P

ten3roberts: Great. Often great solutions come from stepping back and looking at the larger picture, rather than adding yet another bool argument

lua/neogit/config.lua
```diff
@@ -59,12 +59,10 @@ M.values = {
     item = { ">", "v" },
     section = { ">", "v" },
   },
-  integrations = setmetatable({}, {
-    __index = function(_, key)
-      local ok, value = pcall(require, key)
-      return ok and value or false
-    end,
-  }),
+  integrations = {
```
ten3roberts, line 62: The previous auto integration relies on the metatable being left intact and would break if the table was replaced. Additionally, this makes the config data only and does not muddle printing or probing of the config with runtime determined fields.

This functionality is now *ensured* by the `check_integration` function

lua/neogit/buffers/log_view/init.lua
ten3roberts: Made the change to check and provided some more descriptive error messages that show a notification.

Just fixed a `call("string", true, false, true true)` situation where a big async function begun with an if statement, that either allocated a scratch buffer and did a thing, or did a thing directly. Just pried it apart into two functions, as the bools were never dynamic but always hardcoded anyway

ten3roberts:
What do you think of this @CKolkey

lua/neogit/config.lua
```diff
@@ -153,13 +151,20 @@ M.values = {
...
+  local enabled = M.values.integrations[name]
+
+  if enabled == nil or enabled == "auto" then
+    local success, _ = pcall(require, name)
+    logger.fmt_info("Found auto integration '%s = %s'", name, success)
+    return success
   end
 
-  return true
+  logger.fmt_info("Found explicit integration '%s' = %s", name, enabled)
```
CKolkey, line 166: I'd like to start prefixing these with `[LOCATION]` to make reading the log easier. So, `[CONFIG]` in this case

CKolkey (changes requested):
Minor change, then good to go :)

lua/neogit/config.lua
ten3roberts: Like this e09dd40?

CKolkey (approved):

lua/neogit/config.lua
CKolkey: Bang-on.

lua/neogit/buffers/log_view/init.lua
CKolkey: Yeahhh. Thats something I remember from.. I think Uncle Bob... early on in my career. If you are controlling a functions behaviour externally with boolean flags... congrats, you have two functions :P

Issue #619, opened 2023-07-16
Custom keys are not shown in the help popup, instead uses defaults


treatybreaker:
**Is your feature request related to a problem? Please describe.**
Currently the help popup in the Status Buffer does not show the actually bound keys for each action, it's hardcoded.

**Describe the solution you'd like**
Use the actual mappings from the `config.mappings.status` to set the correct keys in the help popup.

**Additional context**
Follow up for #564

treatybreaker:
This was resolved with the merge of #659

karmicdude:
@treatybreaker or @CKolkey
now this issue is back on the rise
I set custom mappings for `mappings.status` table and it works, but help shows wrong information with default mappings

miguno:
Same here. I modified the default keybinds (e.g., "u" should open the git pull popup instead of the default "unstage" behavior) and this works, i.e., "u" does open the pull popup. However, the hint does not reflect this configuration change: it still says "[u] unstage".

- Neovim 0.9.5
- "neogit": { "branch": "master", "commit": "760d6d74c328ce2a38b02677bf5161999b992da2" }

Issue #635, opened 2023-07-18
Got error after update


hiberabyss:
### Description

Have following error message after update  neogit plugin:

<img width="1438" alt="image" src="https://github.com/NeogitOrg/neogit/assets/2898670/d5a9d0b5-1e53-4485-91ab-5505576f56ae">

### Neovim version

NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.9.1/share/nvim"

Run :checkhealth for more info

### Operating system and version

macOS 13.4.1 (c) 

### Steps to reproduce

Open vim, execute `:Neogit`

### Actual behavior

Have the error message.

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing neogit and dependencies.")
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
```

CKolkey:
That config works fine for me :\

CKolkey:
Do you have any custom mappings set?

hiberabyss:
> Do you have any custom mappings set?

What kind of mapping may cause the problem?

treatybreaker:
I know this bug. Let me guesss @hiberabyss you're disabling two more default maps by doing this:

```lua
[KEY_HERE] = ""
```

#629 contains a fix for it if that's what you're doing.

hiberabyss:
Still have following error after updating:

[...]

PR #659, opened 2023-07-21
feat: show all the keys for a given command in help

ten3roberts:


treatybreaker:
I have a branch based off of this that has all the config validation goodies from #629 and tests. See [here](https://github.com/treatybreaker/neogit/tree/feat/config-validation).

ten3roberts:
> I have a branch based off of this that has all the config validation goodies from #629 and tests. See [here](https://github.com/treatybreaker/neogit/tree/feat/config-validation).

Yes thank you. I created this just to link in the discussion depending on which direction we go with that PR. If you think this solution is suitable feel free to rebase or cherry pick it over, if you don't think this would work feel free to take your own approach 😊

treatybreaker:
I mean this works *right now*. The only fault I see with it is that it doesn't show *all* hint keys (which I think we show at least two at the top).

The only other critique, if it even is that, is that only one mapping is shown per command so we end up seeing duplicate commands in the help menu with different keys. I don't think this should even be considered a problem imo.

My linked branch is rebased on top of this, already good to go (tests pass, validation functions are updated for the different (original) format over here).

treatybreaker:
Basically this is me saying "Open this PR, I'll close mine, and when this gets merged I'll open another PR based on this one with the config validations and tests." :wink:

ten3roberts:
Those are really good suggestions, I wasn't entirely sure about how to go about multiple keys per line, but i suppose:

```
s/= Stage
c/C-c Commit
`

rather than

```
s Stage
= Stage
c Commit
C-c Commit
```

and the same for the hint line

If you have any wagers about the `/` or other divider please do let me know. It could also be a space. The aligning is already handled by our UI library.

It is again one of those programming problems where the codebase uses both `f` and `f¯¹` and you need to decide which one you use as external input but you ultimately need to resolve both.

I love the validation and all the other goodies in #629 and am looking forward of having that merged as well

**Comment by treatybreaker** [comment]
No I think a slash is fine, something like that would be really good.

If you want to roll with this PR I'll close #629 and when this is merged I'll open a new PR with the validation and tests to cover whatever changes happen over here.

Far prefer the code changes here over #629.

**Comment by ten3roberts** [comment]
Oh great, then I'll fix those up a bit later when I have access to a computer 😃

File: lua/neogit/popups/help/actions.lua [hunk 1]
```diff
@@ -15,9 +15,13 @@ local function present(commands)
       fn = fn[2]
     end
 
-    return { name = name, key = status_mappings[cmd], fn = fn }
+    return util.map(status_mappings[cmd] or {}, function(key)
+      return { name = name, key = key, fn = fn }
+    end)
```
treatybreaker, line 20:
```suggestion
local keymap = status_mappings[cmd]
if keymap ~= nil then
  return { { name = name, key = table.concat(keymap, ","), fn = fn } }
end
```

That should show multiple keys for a single command instead of showing multiple commands with single keys.

ten3roberts: Yes you are correct.

It would create a table of `{ mapping, mapping, nil, mapping }`

... and of course lua being lua lists are defined as *continuous* integer keyed tables. Having a nil effecitely makes list operations only read up to the first nil, so the list is truncated. Much like null-terminated strings in C.

Have no idea who decided that a table is read `table[1], table[2], ...` until you encounter a nil, smh

ten3roberts: image

ten3roberts:
Fixes: #661, #635

lua/neogit/popups/help/actions.lua
CKolkey: My only bone to pick with using white space is that on the branch config popup there are multi key actions that are separated by whitespace. I think a comma is fine, just use a neutral color instead of the same color.

treatybreaker:
> Fixes: #661 #635

Also closes #619 as a heads up.

ten3roberts:
Hmm, managed to comment as a review, oops.

Realised that making the key be a concatenation of all the bound keys won't work as a keymap is created for that in the popup.

It will need a little extra support in the popup to handle multiple keys for a popup action, as the current state will create a keybinding for "s,=" and not 2 different ones

Will finish that up so that we can carry on with your changes @treatybreaker

ten3roberts:
image

Required mucking about in the popup code, but now we have an actual semantic differentiation between the keys.

ten3roberts:
Final things and bits are now done.

Unmapped actions get a special highlight and displayed, rather than being a string `<unmapped>` which actually creates a keymapping `< u n m a p p e d >` which definitely should not happen

image

lua/neogit/popups/help/actions.lua
```diff
@@ -7,19 +7,26 @@ local NONE = function() end
 local status_mappings = require("neogit.config").get_reversed_status_maps()
 
 local function present(commands)
-  local presenter = util.map(commands, function(command)
+  local presenter = util.filter_map(commands, function(command)
```
CKolkey, line 10: Does this still need to be `filter_map`? I don't see a way it returns `nil`.

CKolkey:
> @CKolkey let me know of there are any thoughts you have about this implementation. Otherwise, we'll merge this and move one with @treatybreaker's awesome PR

Seems like this works just fine!

CKolkey (approved):

treatybreaker:
Cool, it got merged 😄. I'll get my stuff ready to go for y'all this evening so you can take a look tomorrow morning. I'm nearly out of my hell week (which I should fully finish this evening).

PR #664, opened 2023-07-25
Add upstream name to "unpulled from" and "unmerged into" sections of status

CKolkey:
Screenshot 2023-07-25 at 11 14 04

per https://github.com/NeogitOrg/neogit/discussions/663, we could be showing the upstream name!


ten3roberts (approved):
Great. Always good to know where you are pushing and pulling to 😄

PR #667, opened 2023-07-25
Expand on Status Sections

CKolkey:
Splits unpulled/unmerged into two different sections for `@{upstream}` and `pushRemote`.
Top `nvim`, bottom `emacs`.

Screenshot 2023-07-26 at 19 59 27

Also splits the internal repo's unmerged/unpulled to differentiate between pushRemote and upstream. This is useful if they are not the same, so you can get more detailed information about the current repo. If upstream and pushRemote _are_ the same, then we'll only show the upstream sections on the status buffer.


lua/neogit/lib/hl.lua
ten3roberts, line 1: Perhaps we should document highlight groups for other plugin authors.

It is always so fun seeing how your project influences the rest of the ecosystem.

ten3roberts (approved):

lua/neogit/lib/hl.lua
CKolkey: Yup! I have a WIP branch doing exactly that.

PR #674, opened 2023-07-28
Add Config Validation

treatybreaker:
This adds config validation for Neogit and tests for the validation (and config).
This also adds a default function for the config, this is to make isolation for the config easier, especially in tests. See the new tests to see how that is being used.

See #629 for some prior discussion on these changes.

### DIFF
--- a/lua/neogit/config.lua
+++ b/lua/neogit/config.lua
@@ -34,7 +34,6 @@ function M.get_default_values()
     disable_insert_on_commit = true,
     use_per_project_settings = true,
     remember_settings = true,
-    use_magit_keybindings = false,
     auto_refresh = true,
     sort_branches = "-committerdate",
     kind = "tab",
--- a/lua/neogit/config.lua
+++ b/lua/neogit/config.lua
@@ -19,159 +19,444 @@ function M.get_reversed_status_maps()
   return result
 end
 
-M.values = {
-  disable_hint = false,
-  disable_context_highlighting = false,
-  disable_signs = false,
-  disable_commit_confirmation = false,
-  disable_builtin_notifications = false,
-  telescope_sorter = function()
-    return nil
-  end,
-  disable_insert_on_commit = true,
-  use_per_project_settings = true,
-  remember_settings = true,
-  use_magit_keybindings = false,
-  auto_refresh = true,
-  sort_branches = "-committerdate",
-  kind = "tab",
-  -- The time after which an output console is shown for slow running commands
-  console_timeout = 2000,
-  -- Automatically show console if a command takes more than console_timeout milliseconds
-  auto_show_console = true,
-  status = {
-    recent_commit_count = 10,
-  },
-  commit_editor = {
-    kind = "split",
-  },
-  commit_select_view = {
+---Returns the default Neogit configuration
+---@return table
+function M.get_default_values()
+  return {
+    disable_hint = false,
+    disable_context_highlighting = false,
+    disable_signs = false,
+    disable_commit_confirmation = false,
+    disable_builtin_notifications = false,
+    telescope_sorter = function()
+      return nil
+    end,
+    disable_insert_on_commit = true,
+    use_per_project_settings = true,
+    remember_settings = true,
+    use_magit_keybindings = false,
+    auto_refresh = true,
+    sort_branches = "-committerdate",
     kind = "tab",
-  },
-  commit_view = {
-    kind = "vsplit",
-  },
-  log_view = {
-    kind = "tab",
-  },
-  rebase_editor = {
-    kind = "split",
-  },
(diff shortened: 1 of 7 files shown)


plugin/neogit.lua
```diff
@@ -21,10 +21,6 @@ end, {
...
-api.nvim_create_user_command(
-  "NeogitResetState",
-  function()
-    require("neogit.lib.state")._reset()
-  end,
-  { nargs = "*", desc = "Reset any saved flags" }
-)
+api.nvim_create_user_command("NeogitResetState", function()
+  require("neogit.lib.state")._reset()
+end, { nargs = "*", desc = "Reset any saved flags" })
```
treatybreaker, line 26: This was formatted by Stylua :shrug:

CKolkey: Sometimes when it does a crap job, I'll stick an ignore comment above and just do myself 🤓

treatybreaker: So long as my formatter and linter are happy then I'm happy. I haven't run into Stylua making lua code too mangled, so I tend to just run with what it outputs. Let me know if you want me to stick an ignore line there and put that block back to the original.

lua/neogit/config.lua
```diff
@@ -19,159 +19,444 @@ function M.get_reversed_status_maps()
...
+    disable_signs = false,
+    disable_commit_confirmation = false,
+    disable_builtin_notifications = false,
+    telescope_sorter = function()
+      return nil
+    end,
+    disable_insert_on_commit = true,
+    use_per_project_settings = true,
+    remember_settings = true,
+    use_magit_keybindings = false,
```
CKolkey, line 37: This can be removed, yea?

plugin/neogit.lua
CKolkey: It's pretty alright, but lacks nuance that other linters have. I'd kill for allowing lining up equal signs.. alas.

lua/neogit/config.lua
treatybreaker: Yeah. Accidentally got included duri g a rebase.

Removing in a moment

treatybreaker: Oh shit, seems it got included when #659 was merged. Huh.

treatybreaker: Done

plugin/neogit.lua
treatybreaker: I hear that if you light candles in a circle around 3am and chant "Mike Pall, Mike Pall, align my code, I beg of you!" five times fast Mike Pall descends from Lua heaven to align the electrons by hand on your drive while screaming "Stylua!" to get your whitepsace looking pretty.

lua/neogit/popups/init.lua
```diff
@@ -25,16 +25,12 @@ end
 --- Returns an array useful for creating mappings for the available popups
 ---@return table<string, Mapping>
 function M.mappings_table()
-  local config = require("neogit.config")
-
   return {
     {
       "HelpPopup",
       "Help",
       M.open("help", function(f)
```
CKolkey, line 32: Do we still need to pass the second argument here, since it's just an empty table now?

plugin/neogit.lua
CKolkey: Thats a tier in leet-code I haven't gotten to yet XD

lua/neogit/popups/init.lua
CKolkey: Like just having `open("help")` without the function as the second arg

treatybreaker: No, it appears we don't need it. I'll push a commit removing the `env` arg from help as it is now unnecessary with the removal of the magit keybinding.

lua/neogit/popups/help/init.lua
```diff
@@ -26,9 +26,7 @@ function M.create(env)
     p = p:action(cmd.keys, cmd.name, cmd.fn)
   end
 
-  p = p:env(env):build()
-
-  p:show()
```
CKolkey, line 31: This is subtle (and something that predates me) but `:show()` _doesn't_ return the same `self` as `:build()`

`:show()` is a method on the `Buffer` class, whereas `:build()` is a method on the popup builder. So, this should be:

```lua
p = p:build()
p:show()
```

CKolkey (changes requested):

lua/neogit/popups/help/init.lua
treatybreaker: Copy that. Fixed in the latest latest force push.

That's interesting that we have a builder pattern that inherits some other non-builder methods.

CKolkey:
Hell yeah! Great work. Thanks :D

PR #676, opened 2023-07-29
feat: Spin out/off branch

### LINKED ISSUE #665: Spin off branch
Allow creating a new branch from the current branches unpushed/unmerged changes.

A common example is making commits to `main`, which are not pushed to `origin`, and *then* creating a branch with those commits and resetting `main` to `origin/main`. I've found myself reaching for this feature very often, as I often start just coding and forgetting to create a branch, or don't know what feature I will make yet.

Magit uses `b s` and `b S` 

https://magit.vc/manual/magit/Branch-Commands.html

Contributions are very welcome

morgsmccauley:
This PR adds the Spin out/off branch commands.

Spin off will; create a new branch based on the current branch, reset the current branch to its remote if it exists, and checkout the new branch. Spin out is almost the same, except that it does not checkout the new branch.

I haven't included the `FROM` functionality mentioned in the [Magit docs](https://magit.vc/manual/magit/Branch-Commands.html#index-b-s), which allows resetting to a source different from the current remote. Let me know if this should be included.

Resolves #665


tests/specs/neogit/operations_spec.lua
```diff
@@ -46,4 +46,24 @@ describe("branch popup", function()
       eq("branch-from-test", get_current_branch())
     end)
   )
+
+  it(
```
morgsmccauley, line 50: These tests aren't complete - what is the best way to setup up commits and verify which commits are present?

lua/neogit/popups/branch/actions.lua
```diff
@@ -22,6 +22,40 @@ local function parse_remote_branch_name(ref)
...
 
+M.spin_off_branch = operation("spin_off_branch", function()
+  if #git.repo.staged.items > 0 or #git.repo.unstaged.items > 0 then
+    notif.create("Staying on current branch as there are uncommitted changes.", vim.log.levels.INFO)
+    return
+  end
+
+  local name = git.branch.create()
+
+  local upstream = git.repo.upstream.ref
```
CKolkey, line 33: Prefer using the `git.branch.*` module here - in this case `git.branch.upstream_ref()` - instead of accessing repo state directly.

lua/neogit/popups/branch/actions.lua
```diff
@@ -22,6 +22,40 @@ local function parse_remote_branch_name(ref)
   return remote, branch_name
 end
 
+M.spin_off_branch = operation("spin_off_branch", function()
+  if #git.repo.staged.items > 0 or #git.repo.unstaged.items > 0 then
```
CKolkey, line 26: If we do a quick stash here, before the reset, can that just be restored once checking out the new branch?

CKolkey (changes requested):
Great start 😍

tests/specs/neogit/operations_spec.lua
CKolkey: That aspect of our test suite is still a bit of a work in progress, and @treatybreaker  might have some ideas here, but you can just use util.system to interact with the normal git binary, make a commit, and read the revision sha and/or the HEAD file. I can probably explain better when I'm not on my phone, if thats not enough to go off ☺️

treatybreaker: Yeah, best way to go for now is to liberally abuse `util.system`.

To set up commits what I recommend is "cleaning" the repository by stashing and clearing that stash like so: 

```lua
local util = require("tests.util.util")

util.system("git stash && git stash clear")
```

Then creating the commits you need like so:

```lua
local util = require("tests.util.util")

util.system("touch random-file && git commit -m 'Commit Message'")
```

This will guarantee that the most recent commit is "Commit Message".

The other thing to note is that the original commits and test repository state can be found by going into `tests/.repo` then changing the `.git.orig` file in there to `.git` and that will make the base test repository we use accessible via the git cli.

I need get around to doing a update to the test utilities so we have some desired state configuration for the repository setup entirely in Lua so this type of stuff is less of a PIA.

lua/neogit/popups/branch/actions.lua
CKolkey, line 26: You can extract 
```lua
#git.repo.staged.items > 0 or #git.repo.unstaged.items > 0
```
to `lua/neogit/lib/git/status.lua` as `M.is_dirty()` (or something like that)

lua/neogit/popups/branch/actions.lua
```diff
@@ -22,6 +22,40 @@ local function parse_remote_branch_name(ref)
...
+  end
+
+  local name = git.branch.create()
+
+  local upstream = git.repo.upstream.ref
+  if upstream then
+    git.cli.reset.hard.args(upstream).call_sync()
+  end
+
+  git.cli.checkout.branch(name).call_sync()
```
CKolkey, line 38: Before resetting the existing branch and checking out the new one, Magit adds a reflog entry for the action:

https://github.com/magit/magit/blob/main/lisp/magit-branch.el#L517
```elisp
(magit-call-git "update-ref" "-m" (format "reset: moving to %s" base) (concat "refs/heads/" current) base)
```

Which, roughly translates to lua as:
```lua
local base = "main"
local current = "feature-branch"

magit_call_git("git", "update-ref", "-m", string.format("reset: moving to %s", base), "refs/heads/" .. current, base)
```

lua/neogit/popups/branch/actions.lua
```diff
@@ -22,6 +22,40 @@ local function parse_remote_branch_name(ref)
   return remote, branch_name
 end
 
+M.spin_off_branch = operation("spin_off_branch", function()
+  if #git.repo.staged.items > 0 or #git.repo.unstaged.items > 0 then
+    notif.create("Staying on current branch as there are uncommitted changes.", vim.log.levels.INFO)
```
CKolkey, line 27: Looking at how Magit handles this (https://github.com/magit/magit/blob/main/lisp/magit-branch.el#L490-L492), this guard clause only applies if we are _not_ checking out the new branch, ie, for `spin_out` but not for `spin_off`

CKolkey (changes requested):

CKolkey:
Hope you don't mind - I saw some API's that would be useful to you here, so I implemented them: https://github.com/NeogitOrg/neogit/pull/677

You'll need to rebase, but I think that'll make your life a little easier ;)

lua/neogit/popups/branch/actions.lua
CKolkey: update: I added `git.log.update_ref(from, to)` for this

lua/neogit/popups/branch/actions.lua
CKolkey: You'll need to implement the input logic since I removed it from `git.branch.create()` - I'm trying to separate the interactivity layer from the library layer. Just steal the logic from here: https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/popups/branch/actions.lua#L69-L73

lua/neogit/popups/branch/actions.lua
morgsmccauley: Ah, you're right 👍🏼

I think I misunderstood this part entirely. If there are uncommitted changes, Magit doesn't short-circuit the operation, it [forces checkout](https://github.com/magit/magit/blob/main/lisp/magit-branch.el#L493) so the changes aren't lost. This behaviour is actually called out in the docs but didn't make sense to me until now:
> If there are any uncommitted changes, then behave exactly like `magit-branch-spinoff'.

Given the above, I'll:
- Remove this guard from `spin_off`
- Make `spin_out` behave like `spin_off` if there are uncommitted changes

lua/neogit/popups/branch/actions.lua
morgsmccauley: Another thing I misunderstood 😅

In Magit, it seems that the ref log is updated _instead_ of hard resetting when the new branch should be checked out, and is to ensure uncommitted changes aren't lost. This would be the default behaviour for `spin_off`, and when there are uncommitted changes during `spin_out`.

I'll update to align with the above.

lua/neogit/popups/branch/actions.lua
morgsmccauley: Good call 👍🏼

lua/neogit/popups/branch/actions.lua
morgsmccauley: Couldn't find `.upstream_ref()` but assume you meant `.upstream()` - updated :)

morgsmccauley: Added the input logic too. I'm wondering though, can I just re-use `M.create_branch()` or does that complicate things because it's a different operation?

```lua
M.create_branch = operation("create_branch", function()
  local name = input.get_user_input("branch > ")
  if not name or name == "" then
    return
  end

  name, _ = name:gsub("%s", "-")
  git.branch.create(name)
  status.refresh(true, "create_branch")
end)
```

morgsmccauley:
Thanks for the help @CKolkey @treatybreaker!

I've made a few more updates, I believe the implementation is more inline with what Magit has now.

I'll change back to draft for now as there's still the following to do:
- Choose which branch to reset to, i.e. `FROM`
- Update automated tests
- Manual testing

Please comment if I'm not on the right track! 🙏🏼

lua/neogit/status.lua
```diff
@@ -1238,6 +1238,10 @@ function M.create(kind, cwd)
   }
 end
 
+function M.is_dirty()
```
CKolkey, line 1241: This should go to `lib/git/status.lua` - this file is for status _buffer_ logic :)

CKolkey (changes requested):

CKolkey:
> Thanks for the help @CKolkey @treatybreaker!
> 
> I've made a few more updates, I've left a couple comments unresolved to ensure my reasoning is correct. I believe the implementation is more inline with what Magit has now.
> 
> I'll change back to draft for now as there's still the following to do:
> 
> * Choose which branch to reset to, i.e. `FROM`
> * Manual testing - haven't done so for the latest changes
> * Add/update automated tests
> 
> Please comment if I'm not on the right track! 🙏🏼

It's looking great! Reading emacs lisp is... not very straightforward. Something I've found is useful is to ask an LLM to translate the emacs-lisp to lua - it won't be valid, but it does a good job to sketch out the logic/control flow.

For example, here's `magit--branch-spinoff` in lua:
```lua
local function magit_branch_spinoff(branch, from, checkout)
    if magit_branch_p(branch) then
        error(string.format("Cannot spin off %s. It already exists", branch))
    end

    if not checkout and magit_anything_modified_p() then
        print("Staying on HEAD due to uncommitted changes")
        checkout = true
    end

    local current = magit_get_current_branch()
    if current then
        local tracked = magit_get_upstream_branch(current)
        local base

        if from then
            if not magit_rev_ancestor_p(from, current) then
                error(string.format("Cannot spin off %s. %s is not reachable from %s", branch, from, current))
            end
            if tracked and magit_rev_ancestor_p(from, tracked) then
                error(string.format("Cannot spin off %s. %s is ancestor of upstream %s", branch, from, tracked))
            end
        end

        if checkout then
            magit_call_git("checkout", "-b", branch, current)
        else
            magit_call_git("branch", branch, current)
        end

        local indirect_upstream_branch = magit_get_indirect_upstream_branch(current)
        if indirect_upstream_branch then
            magit_call_git("branch", "--set-upstream-to", indirect_upstream_branch, branch)
        end

        if tracked then
            base = from and from .. "^" or magit_git_string("merge-base", current, tracked)
            if not magit_rev_eq(base, current) then
                if checkout then
                    magit_call_git("update-ref", "-m", string.format("reset: moving to %s", base), "refs/heads/" .. current, base)
                else
                    magit_call_git("reset", "--hard", base)
                end
            end
        end
    else
        if checkout then
            magit_call_git("checkout", "-b", branch)
        else
            magit_call_git("branch", branch)
        end
    end

    magit_refresh()
end
```

It obviously doesn't work as-is, but it's easier (for me, at least) to read :)


In terms of what you have left, I think you can skip the `FROM` argument for now and just presume the `FROM` is your current branch. Some of the magit functions can be used in a library-like way (not from the UI), but thats not an express goal of ours, as this isn't public api.

All said, its looking really great :D

lua/neogit/popups/branch/actions.lua
CKolkey: No worries, I had misread it too (in the opposite direction). Sounds like you're on the right track now :D

lua/neogit/popups/branch/actions.lua
CKolkey: Ahh, I didn't realize it was instead of, but it looks like you're right. Nice work

lua/neogit/popups/branch/actions.lua
CKolkey: Definitely don't re-use it. While these functions are all co-located, they should be thought of in isolation. `DRY` as in `Definitely repeat yourself` :P

morgsmccauley: Make sense 👍🏼 😄

tests/specs/neogit/operations_spec.lua
ten3roberts: @morgsmccauley I just merged a PR which adds `exec` inside `git_harness.lua` for testing.

It allows you to run an arbitrary git command `{ "git", "log", "--one-line" }` etc as a list of arguments, and it returns a list of the lines.

I think that will greatly help

ten3roberts (changes requested):
Thank you for your work. I am looking forward to being able to forget creating a feature branch before committing to main. 

Left some small comments regarding the implementation

lua/neogit/popups/branch/actions.lua
```diff
@@ -22,6 +22,60 @@ local function parse_remote_branch_name(ref)
...
+    git.log.update_ref(current_branch_name, upstream)
+  end
+
+  status.refresh(true, "spin_off_branch")
+end)
+
+M.spin_out_branch = operation("spin_out_branch", function()
+  local checkout = false
+  if status.is_dirty() then
+    notif.create("Staying on HEAD due to uncommitted changes", vim.log.levels.INFO)
```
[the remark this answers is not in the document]
CKolkey: I think warn makes sense though. It didn't fail, it just took a different code path and is informing the user.

lua/neogit/popups/branch/actions.lua
```diff
@@ -22,6 +22,60 @@ local function parse_remote_branch_name(ref)
...
+    checkout = true
+  end
+
+  local name = input.get_user_input("branch > ")
+  if not name or name == "" then
+    return
+  end
+
+  name, _ = name:gsub("%s", "-")
```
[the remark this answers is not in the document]
CKolkey: Oh, thats something I added a while back. Git branches cant include a space, so instead of erroring, i opted to fix it for the user. IMO its super handy.

lua/neogit/popups/branch/actions.lua
```diff
@@ -22,6 +22,60 @@ local function parse_remote_branch_name(ref)
   return remote, branch_name
 end
 
+M.spin_off_branch = operation("spin_off_branch", function()
```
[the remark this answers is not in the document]
CKolkey: Thats how magit does it. Perfectly reasonable, id just say to extract it to a local function with separate exported functions. We do that pattern for the commit popup, pull, push, and probably others that I forgot.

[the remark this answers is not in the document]
morgsmccauley: Agree, I wanted to re-use but wasn't sure how best to do that given the `operation` wrappers. I'll extract to a single function 👍🏼

lua/neogit/popups/branch/actions.lua
```diff
@@ -22,6 +22,60 @@ local function parse_remote_branch_name(ref)
...
+  if checkout then
+    git.cli.checkout.branch(name).call_sync()
+  end
+
+  local upstream = git.branch.upstream()
+  if upstream then
+    if checkout then
+      git.log.update_ref(current_branch_name, upstream)
+    else
+      git.cli.reset.hard.args(upstream).call_sync()
```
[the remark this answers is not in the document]
morgsmccauley: The completion of this isn't required for the following code so it can definitely be async. I'll update!

morgsmccauley:
> Some of the magit functions can be used in a library-like way (not from the UI), but thats not an express goal of ours, as this isn't public api.

Ah, that makes more sense now.

Given that we can ignore `FROM` I believe it's just the tests remaining. I should hopefully have those finished in the next couple days :).

lua/neogit/popups/branch/actions.lua
```diff
@@ -22,6 +22,46 @@ local function parse_remote_branch_name(ref)
   return remote, branch_name
 end
 
+local function spin_off_branch(checkout)
+  if status.is_dirty() and not checkout then
```
CKolkey, line 26: This would be `git.status.is_dirty()`

lua/neogit/popups/branch/actions.lua
```diff
@@ -22,6 +22,46 @@ local function parse_remote_branch_name(ref)
...
+
+  status.refresh(true, "spin_out_branch")
+end
+
+M.spin_off_branch = operation("spin_off_branch", function()
+  return spin_off_branch(true)
+end)
+
+M.spin_out_branch = operation("spin_out_branch", function()
+  return spin_off_branch(false)
```
CKolkey, line 62: These don't need to return anything

tests/specs/neogit/operations_spec.lua
morgsmccauley: @ten3roberts it looks like `vim.system()` is only available in the release build of Neovim, so this will fail in the [test action](https://github.com/NeogitOrg/neogit/blob/master/.github/workflows/test.yml#L31C53-L31C53) which uses the stable build.

Is there another way to achieve this or should we target release in the action?

morgsmccauley: Actually I didn't need to use `exec`, some of the other existing functions were enough.

I've added tests for both spin out/off now :)

CKolkey:
I think https://github.com/NeogitOrg/neogit/pull/693 should fix the config issue with the spec. Rebase/merge master to grab it

ten3roberts (approved):
Looks great. The merged solution looks fantastic

CKolkey (approved):

CKolkey:
Fantastic work :D Thanks for your contribution

morgsmccauley:
Thank you for all the help @CKolkey @ten3roberts @treatybreaker!!

PR #679, opened 2023-07-29
fix: always open new status buffer for `open` command

### LINKED ISSUE #668: Switch git directory of the neogit tab on the fly
I try to switch repositories on the fly with `:Neogit cwd=~/myfirstrepo` and after that with `:Neogit cwd=~/mysecondrepo`, but after the second command the tab where neogit is open does not switch to that directory.

I wan't the neogit tab to be pointing to "~/mysecondrepo" after the second command.

can't think of any alternative right now.

**Additional context**
-

treatybreaker:
This is done to ensure the `cwd` is updated for the status buffer.

1. This ensures that Neogit uses the updated cwd when passed the `cwd` flag
2. This ensures that Neogit *always* gets a cwd via the `open` command

Closes #668

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -46,6 +46,10 @@ local open = function(opts)
     opts.cwd = vim.fn.expand(opts.cwd)
   end
 
+  if not opts.cwd then
+    opts.cwd = vim.fn.getcwd()
+  end
+
   if not did_setup then
     notification.create("Neogit has not been setup!", vim.log.levels.ERROR)
     logger.error("Neogit not setup!")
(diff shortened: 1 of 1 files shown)


CKolkey:
Does this restore the cwd when the status buffer closes?

treatybreaker:
> Does this restore the `CWD` when the status buffer closes?

No, but it should be unnecessary to set the `CWD` again. Opening the status buffer necessarily requires opening a new window. Setting the `CWD` affects the entirety of Neovim, true, but switching back to the original window/tab, either by closing the window/tab or just switching to it, will set the `CWD` back to that original window.

See this:

https://github.com/NeogitOrg/neogit/assets/58627896/33138d2b-5ea9-4b2f-84e6-67139e253a2b

Notice that the `CWD` is different between the two windows when checked. Since that's the case, it is unnecessary to set the `CWD` back and, in fact, harmful as a refresh of the status buffer would go back to the previous `CWD` which is undesired.

CKolkey:
Cool, good demo. I didn't realize that would be reset automatically

Issue #680, opened 2023-07-31
Missing Vimdocs


grassdne:
Hey! Cool plugin! I noticed that your vim [documentation is horribly lacking](../blob/master/doc/neogit.txt) and wanted to let you all know that this is incredibly lame!

Just making sure you're aware 🙃

CKolkey:
Perhaps you don't mean it, but your comment comes off as rude, and I don't appreciate it. 

That said, yes, the docs need work. Feel free to open a PR if you'd like to provide anything of value to this conversation, or you can wait for my one year old daughter to bang them out. She's almost done, just needs to learn how to talk first.

PR #683, opened 2023-07-31
Intricate hand-crafted documentation for the basic Lua API made with love and appreciation

grassdne:
- Maybe in the future one would prefer to have docs autogenerated with [the README](https://github.com/kdheepak/panvimdoc), source code documentation, custom scripts, etc for easier maintenance.
- I don't love the name "popups" but it's not that significant. 
- #680: I'm truly sorry for my words here yesterday.  I did not mean to sound cruel.

### DIFF
--- a/doc/neogit.txt
+++ b/doc/neogit.txt
@@ -31,6 +31,8 @@ CONTENTS                                        *neogit-contents*
     2. Commands                      |neogit-commands|
     3. Mappings                      |neogit-mappings|
     4. Highlights                    |neogit-highlights|
+    5. API                           |neogit-api|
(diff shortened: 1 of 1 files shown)


CKolkey:
Hey! Thanks a ton for following up on this - no hard feelings. And, seriously, top notch work on the docs. I genuinely appreciate your contribution :)

Now, to your points:
1. That could definitely be cool. The readme is... also in need of a bit of love. It would be nice to hit two birds with one stone.
2. Haha, well, we kinda inherited that from magit: https://magit.vc/manual/magit-popup.html
3. Apology accepted :)

PR #703, opened 2023-08-04
Remove todo section

ten3roberts:
This file is from before our time and is explicitly stated as not maintained.

The features suggested are mostly stale or implemented. There are some we may consider resurrecting or keeping. They should instead be converted to proper GitHub issues where people can take part of the planning instead.


CKolkey (approved):
Yeah, no reason to keep it around

ten3roberts:
Good.

Still going through old open issues to find if there are things which we have accidentally implemented without knowing about the issue. I've done it many times and accidentally fixed (and also broken 😝) things.

Will resolve the all these things and other leftover bits and pieces bit by bit

PR #706, opened 2023-08-05
fix: if signs are not disabled, avoid overrided by user settings

ofseed:
`signcolumn` may be set to `no` and there is no way to set it to yes only in Neogit.

So Neogit may set its own sign value when the user does not disable it.


CKolkey:
Good thinking :)

PR #709, opened 2023-08-06
fix: open commit view of unmerged/unpulled section

pastalian:
Oversight in commit 7241375be7b0ee5cf490221b9552f1c4ea871d22


CKolkey:
Thanks :)

PR #711, opened 2023-08-06
Add documentation for commit popup

### LINKED ISSUE #713: Error when attempting to use instant fixup
### Description

When I try to use instant fixup, it errors out in cases when it wouldn't previously. On the latest commit, I can't even select from the commit selection window so this error isn't reachable. I can reproduce this on commit 1d740143f02b3cee39a426172c5747b22015194f, because I can't select a commit to fixup on the latest version I can't figure out if this has been fixed.

<details>

<summary>Erorr Message:</summary>

Error executing vim.schedule lua callback: ...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:18: The coroutine failed with this message: ....local/share/nvim/lazy/neogit/lua/neogit
/lib/git/cli.lua:538: bad argument #1 to 'ipairs' (table expected, got string)
stack traceback:
        [C]: in function 'error'
        ...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:18: in function 'callback_or_next'
        ...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:45: in function <...share/nvim/lazy/plenary.nvim/lua/plenary/async/async.lua:44>

</details>

### Neovim version

NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info

Linux 6.4.7-arch1-3

### Steps to reproduce

1. Make a change in a git repo
2. Open neogit
3. commit -> instant fixup
4. (cannot select commit on latest version of neogit, continue if you can)
5. select commit to target for fixup
6. error occurs

### Expected behavior

It should produce an instant fixup without error

### Actual behavior

The above error

### Minimal config

```Lua
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
```

CKolkey:
- Adds documentation for Commit popup
- Changes CommitSelectBuffer to always return a table. If the user cancelled, this table is empty. Callers have been adjusted accordingly. The reason for this was to simplify calling code. There's no reason to check for nil, just handle 0 or more commits.
- Adds commit popup to log view and reflog view (normal and visual mode)
- Fixes a bug in the mappings manager: Currently, if the same key is defined for two different modes (normal, visual), the first one is clobbered by the second one as the are all stored in a single 1-dimentional table by the key. This is fixed by using a 2-dimentional table instead, where the first layer is the mode prefix.
- Also removed the third boolean argument for visual mappings (dictating if the selection should be cleared). We always said that it should be, and I don't forsee a reason why it shouldn't be.
- Added `git.log.is_ancestor(commit_a, commit_b)` function
- Added `git.status.anything_staged()` and `git.status.anything_unstaged()` functions
- Added `commit_popup -> augment` action
- Expanded on the logic in the `commit_special` function:
  - If nothing is staged, we check if anything is unstaged. If there is, offer to add `--all` flag. Otherwise bail and warn user.
  - If rebasing (instant fixup/squash), check if base commit is ancestor to HEAD. Offer resolution if not.
  - Always pass `--no-edit` unless otherwise specified
  - Confirm modifications for instant squash/fixup when already pushed.
- Pass env into commit popup which can contain a commit hash to be used with `commit_special` actions.

Also fixes #713


ten3roberts:
I've used [Mini.Doc](https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-doc.md) for auto generating help files from source code.

I know that there is another solution floating around, but can't currently find it.

Maybe it is something we can use to keep our docs up to date, let me know what you think about that. It would also tie well with the lsp type annotations that have been added to for example the config.lua

CKolkey:
I did consider something like that, and I have no objections to it, and... have no real reason why I didn't use it. 😂

Writing it out by hand is kinda relaxing? I'll look into setting that up.

PR #719, opened 2023-08-08
Allow to set kind=replace in the config

### LINKED ISSUE #718: Please restore kind=replace config option
### Description

Hi!

It seems that after #674 `kind=replace` config is disallowed, I hope that's just a bug. I suggest to restore it. As that's only in validation logic, it's trivial to fix it, but I can make a PR if you'd like.

### Neovim version

latest

linux

### Steps to reproduce

set kind=replace

### Expected behavior

_No response_

### Actual behavior

validation error

### Minimal config

```Lua
require("neogit").setup{
  kind = "replace",
}
```

festeh:
Closes #718

### DIFF
--- a/lua/neogit/config.lua
+++ b/lua/neogit/config.lua
@@ -294,12 +294,12 @@ function M.validate_config()
   local function validate_kind(val, name)
     if
       validate_type(val, name, "string")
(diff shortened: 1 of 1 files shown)


CKolkey:
Thanks!

PR #734, opened 2023-08-10
Checkout commit in log buffer and under the cursor

ten3roberts:
Currently deciding between `bL` and using `enter` in the log view.

The former is more discoverable, while the latter is perhaps more intuitive

Magit uses `bb` for both branch and commit, and will use the commit under the cursor if not specified, otherwise it prompts for a branch.

Perhaps a hybrid approach of both enter in log buffer, and a short circuit with `bb` when the cursor is on a commit in the `Recent Commits` section instead of opening telescope with branches


ten3roberts:
Yeah, thought so as well

ten3roberts:
No worries, now it is workiing even better, and the list shows the hovered commit. I managed to get it such that the LogView can open the Branch view, whereupon it prepopulates the commit you are on, which in creates this really consistent flow of popups and views.

So `Ll` select the commit you want and `bb` and you have it checked out.

Discovered that the status buffer `Head: (detached) ...` does not show the commit hash, will fix that as well

ten3roberts:
Fixes: #734

lua/neogit/buffers/log_view/init.lua
```diff
@@ -140,9 +141,24 @@ function M:open()
           local dv = require("neogit.integrations.diffview")
           dv.open("log", stack[#stack].options.oid)
         end,
+        ["b"] = require("neogit.popups").open("branch", function(p)
```
CKolkey, line 144: Oh, shoot, does this work?? Very cool. I guess it would be good to require popups as a local, and use the same open function pattern for the cherry pick and revert popups.

Oh, and add branch to visual mode too (you were probably already doing that though :p)

lua/neogit/lib/git/cli.lua
```diff
@@ -181,6 +181,11 @@ local configurations = {
           return tbl.args(branch)
         end
       end,
+      commit = function(tbl)
```
ten3roberts, line 184: Frankly I find it very difficult when doing these things as I always have to refer to this file and *search* to see if the thing I want to do is implemented, only to find out it sortof is but under a too specific name.

Both `commit` and `branch` just add a normal argument.

It would be far less cognitive load to just `{ "git", "checkout", commit }` as I know what I *want*, and don't necessarily need a semantic wrapper around each of those arguments which has the wrong name given the context.

Again, once `vim.system` lands I consider switching fully to doing raw git commands and wrapping them in higher level concepts like `lib.git.branch.checkout(commit)` upon which you also get lsp completion and help.

ten3roberts: The current state is just this magic black box where is *sortof* does the right thing, but then inserts things at unexpected places which are different than the order you specified them, and that is without considering the implementation bugs we likely still have in this as it is rather complex.

Not bashing it or anything negative, about it, just that it goes against my first few intuitions I have when approaching programming problems.

CKolkey: Thats fair, I've got no special affection for this abstraction. Once the `vim.system` api lands in stable, I'm all for it :)

CKolkey: I think the real bummer here is that the libgit2 bindings for lua are written in lua 5.4 😞

lua/neogit/buffers/commit_select_view/init.lua
```diff
@@ -98,9 +96,14 @@ function M:open(action)
         end
       end,
     },
-    after = function()
-      if commit_at_cursor then
-        vim.fn.search(commit_at_cursor.oid)
+    after = function(buffer, win)
+      if item and item.commit then
```
CKolkey, line 100: Not opposed to this change, I'd just recommend setting the number of commits to something fairly high (couple thousand) and see if/how it impacts the buffer load time. Maybe its negligible, just wanna check.

The find_component function doesn't have any kind of early return once its found something iirc, so it's going to loop over every single component even if it finds it on the first go.

Same goes for the log buffer.

lua/neogit/lib/buffer.lua
```diff
@@ -469,7 +470,9 @@ function Buffer.create(config)
   end
 
   if config.after then
-    buffer:call(config.after)
+    buffer:call(function()
+      config.after(buffer, win)
```
CKolkey, line 474: Just curious, whats the use case for passing the window into this?

lua/neogit/buffers/commit_select_view/init.lua
ten3roberts: Perhaps we should just short circuit in the find_component then, and add an option for max_count in it

lua/neogit/lib/buffer.lua
ten3roberts: The buffer may not always be shown, depends on the value of `open` which is sometimes user configurable, and other autocommands. Additionally, autocommands may open other windows

This just makes sure we only operate on the window we intent, and not scroll e.g a `notify-nvim` popup that happened to appear

README.md
ten3roberts, line 1: Not exactly sure about what happened here.

I think Obsidian did something to the whitespace, fixing

ten3roberts: image

Looks like a `\r` got removed

image

Anyhow, started expanding on what you can do with neogit in the readme, to make it more discoverable, as frankly even I find new amazing features I didn't know we had, like the captain in Wall-E finding out they have a jogging track 😄

lua/neogit/lib/buffer.lua
CKolkey: Ahhh, ok, thats a good call. I've had that issue with the notify popups. Never thought too hard about it, but this is definitely better.

lua/neogit/status.lua
```diff
@@ -177,13 +177,24 @@ local function draw_buffer()
...
+  )
+  else
   output:append(
     string.format(
       "Head:     %s %s",
       git.repo.head.branch,
       git.repo.head.commit_message or "(no commits)"
     )
   )
+  end
```
CKolkey, line 197: I don't mind ignoring stylua here, but can you indent the inner blocks?

CKolkey (changes requested):

lua/neogit/status.lua
ten3roberts: Fixed, didn't see that there was a stylua ignore there above

lua/neogit/lib/ui/init.lua
```diff
@@ -65,6 +70,14 @@ end
 
 function Ui._find_component(components, f, options)
   for _, c in ipairs(components) do
+    if options.limit then
+      if options.limit == 0 then
+        return nil
+      end
+
+      options.limit = options.limit - 1
+    end
```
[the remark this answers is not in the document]
ten3roberts: Yes, some kind of reverse lookup would be needed

I've reverted the change for now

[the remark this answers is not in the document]
CKolkey: Thanks. Yeah, a reverse lookup could be a better way to do it. But we (at some point) should take this from `O(n)` to `O(1)` (ideally)

CKolkey (approved):

README.md
CKolkey: Loved that movie.

PR #739, opened 2023-08-11
fix: fixes repo not found when special chars in cwd

### LINKED ISSUE #736: Neogit does not recognize git repo when Neovim's CWD contains special characters
### Description

When invoking Neogit when Neovim's CWD contains a directory where some folder has special characters which need to be escaped, it does not recognize that it's in a git repo and asks whether I want to initialize a new one.

I noticed this while working in folders which have square brackets in the name. This is common in Nuxt projects to indicate dynamic routes, for example.

It *only* happens when that folder is in the CWD, for some reason. Opening a file inside such a folder from an nvim instance running in another CWD, does not produce this error.

### Neovim version

NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3

macOS 13.5

### Steps to reproduce

Assuming `minimal.lua` from below is in the current directory.

```sh
mkdir repo
cd repo
git init
mkdir "[bad_name]"
touch "[bad_name]/file.txt"
```

The following will produce the error when trying to invoke Neogit:

```sh
cd "[bad_name]"
nvim -nu ../../minimal.lua file.txt
```

The following will NOT, however:

```sh
nvim -nu ../minimal.lua "[bad_name]/file.txt"
```

### Expected behavior

Neogit sees the repo and shows the git status page.

### Actual behavior

Neogit does not recognize the repo and asks whether I want to initialize one.

### Minimal config

```Lua
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
```

ranmaru22:
This fixes the error where Neogit thinks you're not inside a git repo when invoking it while nvim's CWD it set to any path with special characters in it.

Quoting the interpolated string in `string.format` does the trick.

Fixes #736

### DIFF
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -480,7 +480,7 @@ local git_is_repository_sync = function(cwd)
   if not cwd then
     vim.fn.system("git rev-parse --is-inside-work-tree")
   else
-    vim.fn.system(string.format("git -C \"%s\" rev-parse --is-inside-work-tree", cwd))
+    vim.fn.system(string.format("git -C %q rev-parse --is-inside-work-tree", cwd))
   end
 
   return vim.v.shell_error == 0
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -480,7 +480,7 @@ local git_is_repository_sync = function(cwd)
   if not cwd then
     vim.fn.system("git rev-parse --is-inside-work-tree")
   else
-    vim.fn.system(string.format("git -C %s rev-parse --is-inside-work-tree", cwd))
+    vim.fn.system(string.format("git -C \"%s\" rev-parse --is-inside-work-tree", cwd))
   end
 
   return vim.v.shell_error == 0


ten3roberts:
I think you could use `%q` rather than `\"%s\"` to avoid double escaping and quotes inside the value

Thanks for the PR

ranmaru22:
> I think you could use `%q` rather than `\"%s\"` to avoid double escaping and quotes inside the value

Thanks, that looks better and works too. I changed it.

ten3roberts:
Great.

There would still be an issue when the path contains `$SOMETHING` as the shell will interpret it an an environment variable, but I am not sure if that is even allowed as a filepath.

Anyhow, that is out of scope for now, and we may consider using `vim.system` with a non-shell invocation instead once that is in stable or sooner.

@CKolkey and I are thinking of replacing the whole jobstart and process api with it to make commands such as there easier.

PR #747, opened 2023-08-14
Add a special timeout length for popup buffers

CKolkey:
For users who might have set it very low and then experience issues triggering flags/options


IndianBoy42:
I just tested by manually using `:setlocal timeoutlen=1000` and it seems to override it globally. The docs for `timeoutlen` call it a global option and `setlocal` says 
> If the option does not have a local value the global value is set

PR #759, opened 2023-08-17
fix: set `signcolumns` to 'auto' when disable_signs is false to avoid unnecessary statuscolumn

milanglacier:
When setting `signcolumns` to `yes`, the buffers with filetype `NeogitCommitMessage` and `NeogitCommitMessage` will also have 1 column in the left margin indicating a signcolumn. However, we shouldn't have signcolumns for those buffers. We only need to have signcolumns for the `NeogitStatus` buffer. Setting the option to `auto` will automatically handle the situation.


CKolkey:
Thanks for the contribution :)

Issue #753, opened 2023-08-15
Show current local and remote commit hashes in NeogitStatus


mambusskruj:
**Is your feature request related to a problem? Please describe.**
Yes, it's nice to show commit message header for local and remote sources, but in addition it would be nice to also include commit hash, so user will know if commits are identical.

**Describe the solution you'd like**
Show current local and remote commit hashes in NeogitStatus

CKolkey:
Could be cool to have as an option, sure

mambusskruj:
Yeah, because with commit amend it can starts to be confusing.

CKolkey:
Makes sense. Though, with an ammended HEAD, you'll see both the "unpulled" and "unpushed" sections appear. 

Regardless, its a good idea.

ten3roberts:
I started of this in that it shows the commit ref if you are in a detached HEAD state.

Do you think that the same format could be used for the normal HEAD states? Or would you consider changing it to something else?

mambusskruj:
Thank you! I think its time to close the issue.

PR #763, opened 2023-08-17
fix: show commit abbrev in status buffer

ten3roberts:


ten3roberts:
image

Decided to show the abbrev after the branch name first, like I did for detached, though realizing that placing the abbrev first makes both abbrevs and start of branch names line up.

ten3roberts:
Fixes #753

ten3roberts:
This PR has also subtly started moving to the new suggested higher level `lib.git` wrappers, with the end intent of using those instead of `git.cli`.

When vim.system lands I consider starting to swap out the cli calls in the wrappers, such as `lib/git/rev_parse.lua` to just `process.call({"git", "rev-parse", "--short", oid).stdout[1]` internally

lua/neogit/lib/hl.lua
```diff
@@ -155,6 +155,9 @@ function M.setup()
     NeogitCommandCodeNormal = { link = "String" },
     NeogitCommandCodeError = { link = "Error" },
     NeogitBranch = { fg = palette.orange, bold = true },
+    NeogitHeadOid = { link = "Comment" },
```
CKolkey, line 158: Can these just link to `NeogitObjectId` instead of `Comment`? I wanna make this easier on theme-authors.

syntax/NeogitStatus.vim
```diff
@@ -49,13 +54,14 @@ for section in b:sections
   execute 'syn region Neogit' . id . 'Region start=/^' . section . '\ze.*/ end=/./ contains=Neogit' . id
 endfor
 
-syn region NeogitHeadRegion         start=/^Head: \zs/        end=/$/ contains=NeogitBranch
```
CKolkey, line 52: Any reason for the spacing change?

CKolkey (changes requested):

CKolkey:
Good stuff! I think the last part of this (not that it needs to be in this PR - optional sidequest) is that because they are now OIDs, they should be valid to pass into any of the popups like every other OID.

lua/neogit/status.lua
```diff
@@ -180,15 +180,16 @@ local function draw_buffer()
   if git.repo.head.branch == "(detached)" then
     output:append(
       string.format(
-        "Head:     (detached) at %s %s",
```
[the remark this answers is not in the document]
ten3roberts: Yes.

The teo reasons I decided to try this way is so that the commit hashes line up with Merge and attached Head states,

and the second is that vim didn't quite seem to want to match or escape the parenthesis inside the `syn region` 😅

syntax/NeogitStatus.vim
ten3roberts: I originally indented it to fit `NeogitHeadDetachedRegion` before renaming to `NeogitDetachedRegion` as it was a little long.

Anyhow, it is resolved now.

I always get the same kind of feeling of *oh* when I encounterd old vimscript and need to remember to escape my strings properly as they aren't quoted. The good ol' backslash because you wanted a space in an option, or a combintation of colons because vim does not support lists for options.

ten3roberts:
image

image

Thank you for your help, should have fixed everything up by now.

lua/neogit/lib/git/status.lua
```diff
@@ -130,13 +127,15 @@ local function update_status(state)
 
   state.cwd = cwd
   state.head = head
-  state.upstream = upstream
+  state.upstream.ref = upstream.ref
```
CKolkey, line 130: I don't think this change is right. `state.upstream` needs to be set to the local `upstream` table. I think you can exercise this part by creating a new branch locally (it has no upstream yet)

syntax/NeogitStatus.vim
CKolkey: Yeah... I don't miss vimscript. These patterns especially, for syntax regions/matches. I've done a lot of spelunking in the gitrebase syntax file for inspiration, but oh man, lua/extmarks is a much easier API

lua/neogit/lib/git/status.lua
ten3roberts: Right, I suspected such might be the case.

The reason with setting this as a table is that it will race with the `update_branch_information` async function and 50% of the time causes the abbrev to be overwritten by the new table

ten3roberts: I.e

image

These two will compete over who can set it last.

CKolkey: Hm. Yeah, thats an issue. But the problem I've seen when just trying to update specific parts of it is that when creating a new branch, the `upstream` info persists to this new branch, and you wind up with neogit thinking the upstream is `origin/master` when thats plainly wrong.

There's probably some middle way that solves both issues. Selective nullification?

CKolkey:
Hope you don't mind - I was messing with the syntax stuff and got it to a (I think) nice place. Works when the head is detached or on branch, and I figured pushing a commit was easier than trying to explain it all. If you don't like the change, just reset it or tweak it :)

ten3roberts:
> Hope you don't mind - I was messing with the syntax stuff and got it to a (I think) nice place. Works when the head is detached or on branch, and I figured pushing a commit was easier than trying to explain it all. If you don't like the change, just reset it or tweak it :)

No that's great. I much prefer that than pushing to master and then having to merge back in with conflicts :smiley:

CKolkey:
> > Hope you don't mind - I was messing with the syntax stuff and got it to a (I think) nice place. Works when the head is detached or on branch, and I figured pushing a commit was easier than trying to explain it all. If you don't like the change, just reset it or tweak it :)
> 
> No that's great. I much prefer that than pushing to master and then having to merge back in with conflicts 😃

Haha, yeah, agreed

CKolkey (approved):
I added some early return logic for when HEAD is detached - everything else looks good, merge at will :D

PR #771, opened 2023-08-20
set nolist for popups

### LINKED ISSUE #770: Set `nolist` in popup windows to avoid visual noise
Popup windows don't set `nolist` and therefore can contain visual noise. See the `-` at the bottom right after "Edit", "Augment", and "Instant fixup"
image

Set `nolist` for all popup windows.

Have an autocommand for all the popup filetypes that set `nolist`

**Additional context**
Tested with current version: 1c8a67839dd1e7cc8ae3366cbe1a02ae315f1650

After setting `nolist` the popup looks like this:
image

CKolkey:
Fixes #770

### DIFF
--- a/lua/neogit/lib/popup/init.lua
+++ b/lua/neogit/lib/popup/init.lua
@@ -564,7 +564,7 @@ function M:show()
     kind = config.values.popup.kind,
     mappings = mappings,
     after = function(buf, _)
-      vim.cmd([[setlocal nocursorline]])
+      vim.cmd([[setlocal nocursorline nolist]])
 
       if self.state.env.highlight then
         for i = 1, #self.state.env.highlight, 1 do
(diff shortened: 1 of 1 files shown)


lua/neogit/lib/popup/init.lua
```diff
@@ -564,7 +564,7 @@ function M:show()
     kind = config.values.popup.kind,
     mappings = mappings,
     after = function(buf, _)
-      vim.cmd([[setlocal nocursorline]])
+      vim.cmd([[setlocal nocursorline nolist]])
```
gegoune, line 567: Why not 
```suggestion
      vim.opt_local.list = false
```
?

CKolkey: Yea, I was thinking something like that would be nice. Decided to go with this since we have the window ID as the second function arg 🤷🏼 Which is.... more... specific?

```lua
      vim.api.nvim_set_option_value("cursorline", false, { win = win })
      vim.api.nvim_set_option_value("list", false, { win = win })
```

gegoune: double win/win even! :) Thanks for that change!

PR #785, opened 2023-08-26
feat: Add basic fzf-lua integration

jarviliam:
Adds the basic popup integration for fzf-lua. I mapped a majority of the telescope options to the fzf-lua equivalents as to not add more custom options.
Also if both telescope and fzf-lua integrations are enabled, the telescope integration takes precedence.

Let me know if I need to fix anything or if I'm missing anything.


CKolkey:
Looks good! Thanks for the contribution :)

PR #786, opened 2023-08-26
fix: Do not append duplicate `max-count` on nvim versions below 0.10

jarviliam:
Fixes a bug when on nvim less than 0.10 where all log commands have a duplicate `--max-count` appended to them. This is because the `vim.tbl_contains` function used on 0.10 allows for a function/predicate match while lower versions do not.


CKolkey:
Thanks - IIRC there are some other 0.10 API calls in the codebase, but not really "on purpose".

Issue #784, opened 2023-08-26
Log, Show Signature should only highlight the hash until full commit is displayed


theherk:
**Is your feature request related to a problem? Please describe.**
I'm not quite sure this is a bug having looked through the code, but it is a usability issue and doesn't function as well as the feature in Magit, in my view. When generating the log with `=S` "Show Signatures" enabled, the log shows the signature blocks in the table view. Not really helpful, and very cluttered. All you probably need to know in this view is signed / unsigned and good sig / bad sig.

**Describe the solution you'd like**
This argument should instead pass to the underlying commit fetch for displaying the commit. So the only difference shown on the log page should be that the commit hash is colored, like green instead of grey, indicating a good signature. Then if you actually select the commit `<cr>` to display, you should see the signature block.

**Additional context**
Here are some screenshots in Magit showing what I mean.

<img width="1536" alt="Screenshot 2023-08-26 at 14 52 52" src="https://github.com/NeogitOrg/neogit/assets/3605078/17a42da0-568b-425c-8cb2-d30936531e36">

<img width="1536" alt="Screenshot 2023-08-26 at 14 53 26" src="https://github.com/NeogitOrg/neogit/assets/3605078/f44f6a53-7868-4ae1-9f91-ff45933cd768">

jarviliam:
I can adrees this fix this if its open. Currently neogit passes the --show-signature argument to git log. Magit however appears to check if the argument is passed, remove the argument and substitute the log format with "%G?" to show the result of the commit signature. Then we would need to add a highlight to the good signatures and bad ones.

I think magits approach to remove the show-signature line is correct for the log list view (have to check how they show it in the details page though)

CKolkey:
Wide open ;) 

So, if you wanna go about it that way, here's where to look:
https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/popups/log/actions.lua

Note `popup:get_internal_arguments()` - thats used for the graph rendering. 
https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/popups/log/init.lua#L37

Commit lines and graph only lines: https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/buffers/log_view/ui.lua

You could pretty easily extend the component to show a different color hash:
https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/buffers/common.lua#L91-L192

[...]

PR #787, opened 2023-08-27
feat: Add highlighting to commit based on signature flag

jarviliam:
Modifies the log view buffer to highlight the commit based on the status of the commit signature.
I tried mapping the highlight colors from magit but feel free to adjust them.

Still need to add the signature output on the commit details buffer but that one is a bit more messy as I need to record if `--show-signature` was passed in the parent BufferWindow and pass it down to the CommitDetailView to be used in `git show`. I'll work on that in another PR 🤔 .

Solves half of #784

### DIFF
--- a/lua/neogit/buffers/common.lua
+++ b/lua/neogit/buffers/common.lua
@@ -88,6 +88,25 @@ local function build_graph(graph)
   end
 end
 
+-- - '%G?': show "G" for a good (valid) signature,
+--   "B" for a bad signature,
+--   "U" for a good signature with unknown validity,
+--   "X" for a good signature that has expired,
+--   "Y" for a good signature made by an expired key,
+--   "R" for a good signature made by a revoked key,
+--   "E" if the signature cannot be checked (e.g. missing key)
+--   and "N" for no signature
+local highlight_for_signature = {
+  G = "NeogitGraphGreen",
+  B = "NeogitGraphRed",
+  U = "NeogitGraphBoldBlue",
+  X = "NeogitGraphOrange",
+  Y = "NeogitGraphOrange",
+  R = "NeogitGraphRed",
+  E = "NeogitGraphBlue",
+  N = "NeogitGraphGray",
+}
+
 M.CommitEntry = Component.new(function(commit, args)
   local ref = {}
   if commit.ref_name ~= "" then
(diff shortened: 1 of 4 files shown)


lua/neogit/buffers/common.lua
```diff
@@ -88,6 +88,25 @@ local function build_graph(graph)
...
+local highlight_for_signature = {
+  G = "String", -- Green
+  B = "Error", -- Red
+  U = "Operator", -- Cyan
+  X = "SpecialChar", -- Orange
+  Y = "SpecialChar", -- Orange
+  R = "Error", -- Red
+  E = "Special", -- Blue
+  N = "Comment", -- Default
+}
```
CKolkey, line 108: Check out `hl.lua` - we have highlight groups for each color. They're only used for the `--color` flag when viewing the log graph, but no reason they can't be used here too.

CKolkey (changes requested):

CKolkey:
Great stuff! Small change to the highlight groups, but otherwise looks good to me.

jarviliam:
@CKolkey Modified to use the graph colors. Thanks for letting me know!

lua/neogit/lib/git/log.lua
```diff
@@ -236,6 +237,7 @@ local format = table.concat({
   "%ce", -- Committer Email
   "%ae", -- Author Email
   "%b", -- Body
+  "%G?",
```
CKolkey, line 240: Can you add a comment here for what this one is?

lua/neogit/lib/hl.lua
```diff
@@ -115,6 +115,7 @@ function M.setup()
     NeogitGraphBlue = { fg = palette.blue },
     NeogitGraphPurple = { fg = palette.purple },
     NeogitGraphGray = { fg = palette.grey },
+    NeogitGraphOrange = { fg = palette.orange },
```
CKolkey, line 118: Can you add this to the documentation section for highlight groups? (`doc/neogit.txt:265`)

CKolkey (changes requested):
Two small things then this is good to go!

lua/neogit/lib/hl.lua
jarviliam: Added along with `NeogitGraphGray` because it was missing. The docs also state `NeogitGraphBlack` but I think it was removed in a previous commit?

CKolkey: Yea, I got my wires crossed between writing the highlighting logic and documenting it. Thanks for catching that :)

CKolkey:
Thanks for the contribution!

PR #790, opened 2023-08-28
feat: Show signature details in commit detail view

jarviliam:
Modifies log commit detail view to show the signature block when the `--show-signature` flag is passed to the log buffer. When the log commit detail view buffer is opened it will run `git verify-commit` against the hash and record the output.

It might be a bit messy but we have to record if the --show-signature flag was set, then remove the --show-signature flag from the log graph view command and the actual git log command.
Then when we open a commit log from the log list page we need to check if the --show-signature flag was set from the parent git log command and run `git verify-commit` to output the signature.

I tried just parsing the signature lines in the intial `git log` command and store it in the actual commit struct but its difficult to tell where a signature block ends with just string parsing alone.

Second half of #784

### DIFF
--- a/lua/neogit/buffers/commit_view/init.lua
+++ b/lua/neogit/buffers/commit_view/init.lua
@@ -15,6 +15,7 @@ local M = {
 -- @class CommitViewBuffer
 -- @field is_open whether the buffer is currently shown
 -- @field commit_info CommitInfo
+-- @field commit_signature table|nil
 -- @field commit_overview CommitOverview
 -- @field buffer Buffer
 -- @see CommitInfo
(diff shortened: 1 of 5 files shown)


jarviliam:
@CKolkey This PR is a bit more messy than the last one but if you have any tips/suggestions on better ways to approach my implementation/lua usage please let me know! 👍

CKolkey:
Hmm. Ok, I thought about this today: Since we're getting the information anyways, don't worry about conditionally displaying it. Just show the signature line if the commit is signed, and don't show it if it's not. That way commits shown from the reflog buffer or status buffer can benefit from the info, because otherwise it wouldn't be shown there. Hows that sound to you?

jarviliam:
Yeah thats fine with me 👍 . The only consequence I can think of is that the `--show-signature` flag basically becomes unused. Should we remove it or just leave it in for now?

CKolkey:
That flag can still control how the log view is rendered, so I think its find to leave.

Another thing i noticed is that Magit will silently drop it when fetching more than 256 commits, for performance reasons. Could you check if we should do the same?

jarviliam:
Added 👍 . My mistake I misunderstood how magit handles the `%G?`, I thought it was used regardless of the `--show-signature` argument but it seems like its not.

CKolkey:
No worries, I just happened to find it yesterday by coincidence: https://github.com/magit/magit/blob/main/lisp/magit-log.el#L175-L183

CKolkey:
Code wise everything looks reasonable! Anything else left, or can I merge this?

jarviliam:
Nope, I think thats it. Might look into making the signature block a little more prettier in the future but should be good for now.

Thanks !

PR #797, opened 2023-08-30
feat: Show tag name and distance from head on status page

jarviliam:
Similar to Magit's status page this commit adds the most recent tag name to the status popup with the distance (commits) between the head and tag

Unfortunately lua's pattern matching isn't as complete as I had hoped for so there might be some cases where it fails. When it fails it will just set the tag and distance to nil making it unrendered.

### DIFF
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -96,6 +96,13 @@ local configurations = {
     },
   },
 
+  describe = config {
+    flags = {
+      long = "--long",
+      tags = "--tags",
+    },
+  },
+
   diff = config {
     flags = {
       cached = "--cached",
--- a/lua/neogit/lib/git/repository.lua
+++ b/lua/neogit/lib/git/repository.lua
@@ -19,7 +19,11 @@ local function empty_state()
     git_root     = root,
     head         = {
       branch = nil,
-      commit_message = nil
+      commit_message = nil,
+      tag = {
+        name = nil,
+        distance = nil,
+      },
     },
     upstream     = {
       branch         = nil,
--- a/lua/neogit/lib/git/status.lua
+++ b/lua/neogit/lib/git/status.lua
@@ -15,6 +15,11 @@ local function update_file(file, mode, name)
   return setmetatable({ mode = mode, name = name, has_diff = has_diff, diff = diff }, mt or {})
 end
 
+-- Generic pattern for matching tag ref and distance from rev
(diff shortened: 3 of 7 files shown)


CKolkey:
Nice! I've been wanting to add this but... life, you know. My only note is that this doesn't support when there are multiple tags. Have you considered using `git tag --points-at HEAD` to get the output instead?

lua/neogit/lib/git/status.lua
```diff
@@ -15,6 +15,11 @@ local function update_file(file, mode, name)
   return setmetatable({ mode = mode, name = name, has_diff = has_diff, diff = diff }, mt or {})
 end
 
+-- Generic pattern for matching tag ref and distance from rev
+-- Unfortunately lua's pattern matching isn't that complete so
+-- some cases may be dropped.
+local tag_pattern = "(.-)%-([0-9]+)%-g[0-9a-z]+$"
```
CKolkey, line 21:
```suggestion
local tag_pattern = "(.-)%-([0-9]+)%-g%x+$"
```
`%x` matches hexidecimal chars :)
https://www.lua.org/pil/20.2.html

lua/neogit/lib/git/repository.lua
```diff
@@ -21,6 +21,10 @@ local function empty_state()
       branch = nil,
       commit_message = nil
     },
+    tag = {
+      name = nil,
+      distance = nil,
+    },
```
CKolkey, line 27: Since it's always using `HEAD`, I think this could be pulled into the `head` table, ie:

```lua
{ head = { tag = { name = nil, distance = nil } } }
```

lua/neogit/lib/git/cli.lua
```diff
@@ -95,6 +95,12 @@ local configurations = {
       end,
     },
   },
```
CKolkey, line 97: This is just me being fussy (sorry), can you add an empty line between sections? It gets pretty tough to read otherwise.

CKolkey (changes requested):

CKolkey:
Minor changes. We can also defer multiple tag support for another PR. Syntax highlighting isn't a maybe, IMO, but can also be implemented in a new PR.

jarviliam:
Adjusted 👍 . I'll look into porting the multiple tag support in another PR.

CKolkey (approved):

CKolkey:
Nice! Thanks for this

PR #799, opened 2023-08-30
Fix pulling remotes, when there is a slash in the branch name.

CKolkey:


ten3roberts:
I suppose you discovered it with `feat/tag-distance-status` 😆 

Great job ❤️

PR #806, opened 2023-09-01
fix: Do not continue remote action when input is empty

jarviliam:
This commit fixes a small bug where if the user escapes out of the input action the return value is an empty string and the action attempts to still add the remote.


CKolkey:
Nice! Thanks

PR #808, opened 2023-09-02
feat: Add Tag popup

### LINKED ISSUE #783: Tagging
Tagging doesn't seem to be implemented yet.

I'd enjoy a tagging interface just like [Magit's tagging](https://magit.vc/manual/magit/Tagging.html).

For the moment I just use the CLI or LazyGit.

**Additional context**
It could be that this is already implemented, and I'm blind but I don't see it, nor conversations about it, and I didn't see it in the `umbrella` list.

jarviliam:
This commit implements the magit tag popup.

Completed:
- Create tag
- Delete tag

Needs improvement:
- Prune

Haven't started:
- Create release

The prune implementation works but I need to call `ls-remote` asynchronously as it takes  a bit of time.
I also need to figure out how to show the user the list of local/remote tags to be pruned by selecting the y option 🤔 .

Closes #783

### DIFF
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -123,6 +123,14 @@ local configurations = {
     },
   },
 
+  tag = config {
+    flags = {
+      n = "-n",
+      list = "--list",
+      delete = "--delete",
+    },
+  },
+
   rebase = config {
     flags = {
       interactive = "-i",
@@ -457,12 +465,21 @@ local configurations = {
   },
 
   ["ls-remote"] = config {
+    flags = {
+      tags = "--tags",
+    },
     aliases = {
       remote = function(tbl)
         return function(remote)
           return tbl.args(remote)
         end
       end,
+    },
+  },
+
+  ["for-each-ref"] = config {
+    options = {
+      format = "--format",
     },
   },
 
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -597,6 +597,7 @@ local mt_builder = {
         return tbl
       end
     end
+
     if action == "arg_list" then
       return function(args)
         for _, v in ipairs(args) do
--- a/lua/neogit/popups/tag/actions.lua
+++ b/lua/neogit/popups/tag/actions.lua
@@ -1,19 +1,24 @@
 local M = {}
+
 local git = require("neogit.lib.git")
 local client = require("neogit.client")
 local utils = require("neogit.lib.util")
 local FuzzyFinderBuffer = require("neogit.buffers.fuzzy_finder")
 local input = require("neogit.lib.input")
 local notification = require("neogit.lib.notification")
-local a = require("plenary.async")
 
 function M.create_tag(popup)
   local tag_input = input.get_user_input("Tag name: ")
-  local options = git.refs.get_revisions()
-  local selected_branch = FuzzyFinderBuffer.new(options):open_async()
+  if not tag_input or tag_input == "" then
+    return
+  end
+  tag_input, _ = tag_input:gsub("%s", "-")
+
+  local selected_branch = FuzzyFinderBuffer.new(git.refs.list()):open_async()
   if not selected_branch then
     return
   end
+
   local args = popup:get_arguments()
   if vim.tbl_count(args) > 0 and not vim.tbl_contains(args, "--annotate") then
     table.insert(args, "--annotate")
@@ -38,25 +43,12 @@ function M.create_release(_) end
 --- git tag -d TAGS
 ---@param _ table
 function M.delete(_)
-  local options = git.tag.list()
-  local tags = {}
-  if options ~= nil then
-    local selected_tags = FuzzyFinderBuffer.new(options):open_async { allow_multi = true }
-    if #selected_tags == 0 then
-      return
-    end
-    tags = selected_tags
-  else
-    local tag_input = input.get_user_input("Tag name:")
-    if not tag_input or tag_input == "" then
-      return
-    end
-    table.insert(tags, tag_input)
+  local tags = FuzzyFinderBuffer.new(git.tag.list()):open_async { allow_multi = true }
+  if #tags == 0 then
+    return
   end
 
-  local result = git.tag.delete(tags)
-  a.util.scheduler()
-  if result then
+  if git.tag.delete(tags) then
     notification.info("Deleted tags: " .. table.concat(tags, ","))
   end
 end
@@ -64,37 +56,44 @@ end
 --- Prunes differing tags from local and remote
 ---@param _ table
 function M.prune(_)
-  local remotes = git.remote.list()
-  local selected_remote = FuzzyFinderBuffer.new(remotes)
-    :open_async { prompt_prefix = "Prune tags using remote" }
+  local selected_remote = FuzzyFinderBuffer.new(git.remote.list()):open_async {
+    prompt_prefix = " Prune tags using remote > "
+  }
+
   if not selected_remote or selected_remote == "" then
     return
   end
+
   local tags = git.tag.list()
-  if tags == nil then
+  if #tags == 0 then
     return
   end
+
   local r_out = git.tag.list_remote(selected_remote)
   local remote_tags = {}
+
   -- Tags that exist locally put
   for _, line in ipairs(r_out) do
     if not line:match("%^{}$") then
       table.insert(remote_tags, line:sub(52))
     end
   end
+
   local l_tags = utils.set_difference(tags, remote_tags)
   local r_tags = utils.set_difference(remote_tags, tags)
 
   if #l_tags == 0 and #r_tags == 0 then
-    a.util.scheduler()
     notification.info("Same tags exist locally and remotely")
     return
   end
 
   local choices = { "&delete all", "&review each", "&abort" }
+
   if #l_tags > 0 then
-    local choice =
-      input.get_choice(#l_tags .. " tags can be removed locally", { values = choices, default = #choices })
+      local choice = input.get_choice(
+        #l_tags .. " tags can be removed locally",
+        { values = choices, default = #choices }
+      )
     if choice == "d" then
       l_tags = {}
     elseif choice == "r" then
@@ -105,11 +104,13 @@ function M.prune(_)
       return
     end
   end
+
   if #r_tags > 0 then
     local choice = input.get_choice(
       #r_tags .. " tags can be removed from remote",
       { values = choices, default = #choices }
     )
+
     if choice == "d" then
       r_tags = {}
     elseif choice == "r" then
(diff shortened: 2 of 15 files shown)


CKolkey:
Do you want feedback as you go, or later?

jarviliam:
Feedback as I go if possible.
I don't think create release can be done due to lua pattern matching limitations.

CKolkey:
What do you need it to do that it cant? I find them to be pretty versatile.

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,161 @@
...
+
+--- Gets revisions
+---@return table
+local function get_revisions()
+  local revisions = git.cli["for-each-ref"].format('"%(refname:short)"').call():trim().stdout
+  for i, str in ipairs(revisions) do
+    revisions[i] = string.sub(str, 2, -2)
+  end
+  return revisions
+end
```
CKolkey, line 15: I think this could get extracted to `git.refs.list()`

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,161 @@
...
+
+--- Outputs a list of tags locally
+---@return table|nil List of tags.
+local function list_tags()
+  local revisions = git.cli.tag.args("-l").call():trim().stdout
+  if #revisions == 0 then
+    return nil
+  end
+  return revisions
+end
```
CKolkey, line 25: This could move to `git.tags.list()`, and the `-l` arg could get added to the cli builder under `flags = { list = "--list" }`

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,161 @@
...
+
+function M.create_tag(popup)
+  local value = vim.fn.input { prompt = "Tag name: " }
```
CKolkey, line 28: This can use `neogit.lib.input`: https://github.com/NeogitOrg/neogit/blob/08d23c7ae1e7c3c66802990d5aa16269bdd996a7/lua/neogit/lib/input.lua#L67-L69

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,161 @@
...
+  local options = get_revisions()
+  local selected_branch = FuzzyFinderBuffer.new(options):open_async()
+  if not selected_branch then
+    return
+  end
+  local args = popup:get_arguments()
+  if vim.tbl_count(args) > 0 and not vim.tbl_contains(args, "--annotate") then
+    table.insert(args, "--annotate")
+  end
+  client.wrap(git.cli.tag.arg_list({ value, selected_branch }).apply_flags(args), {
```
CKolkey, line 41: Just to understand, why use both the new `apply_flags` and `arg_list`?

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,161 @@
...
+  local tags = {}
+
+  if options ~= nil then
+    local selected_tags = FuzzyFinderBuffer.new(options):open_async { allow_multi = true }
+    if #selected_tags == 0 then
+      return
+    end
+    tags = selected_tags
+  else
+    local tag_input = vim.fn.input { prompt = "Tag name:" }
```
CKolkey, line 72: Same, use built-in input lib.

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,161 @@
...
+    if not tag_input or tag_input == "" then
+      return
+    end
+    table.insert(tags, tag_input)
+  end
+
+  local result = git.cli.tag.arg_list({ "-d", unpack(tags) }).call():trim()
```
CKolkey, line 79: Add `--delete` flag to cli builder. Could also create `git.tag.delete(tags)` function that returns true/false on success/failure.

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,161 @@
...
+    notif.create("Deleted tags: " .. table.concat(tags, ","))
+  end
+end
+
+--- Prunes differing tags from local and remote
+---@param _ table
+function M.prune(_)
+  local a = require("plenary.async")
+  local notif = require("neogit.lib.notification")
+  local remotes = git.cli.remote.call():trim().stdout
```
CKolkey, line 91: Prefer `git.remote.list()` unless you have a good reason not to.

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,161 @@
...
+  local selected_remote = FuzzyFinderBuffer.new(remotes)
+    :open_async { prompt_prefix = "Prune tags using remote" }
+  if not selected_remote or selected_remote == "" then
+    return
+  end
+  local tags = list_tags()
+  if tags == nil then
+    return
+  end
+  local r_out = git.cli["ls-remote"].tags.args(selected_remote).call():trim().stdout
```
CKolkey, line 101: `git.tag.list_remote(remote)` ?

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,161 @@
...
+  end
+
+  if #l_tags > 0 then
+    local prompt = "Delete " .. #l_tags .. " tags locally? (y or n)"
+    if #l_tags == 1 then
+      prompt = "Delete " .. l_tags[1] .. " locally? (y or n)"
+    else
+      -- need to show user tags
+    end
+    local confirm = vim.fn.input { prompt = prompt }
```
CKolkey, line 125: There's a get confirmation function in the input lib you can use here

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,161 @@
...
+    end
+  end
+  if #r_tags > 0 then
+    local prompt = "Delete " .. #r_tags .. " tags from remote? (y or n)"
+    if #r_tags == 1 then
+      prompt = "Delete " .. r_tags[1] .. " from remote? (y or n)"
+    else
+      -- need to show user tags
+    end
+    local confirm = vim.fn.input { prompt = prompt }
```
CKolkey, line 138: same, use input lib to get confirmation

lua/neogit/popups/tag/init.lua
```diff
@@ -0,0 +1,29 @@
...
+    .builder()
+    :name("NeogitTagPopup")
+    :arg_heading("Arguments")
+    :switch("f", "force", "Force")
+    :switch("a", "annotate", "Annotate")
+    :switch("s", "sign", "Sign")
+    :switch("u", "local-user", "Sign as", { enabled = false })
+    :new_action_group("Create")
+    :action("t", "tag", actions.create_tag)
+    :action("r", "release", noop)
```
CKolkey, line 19: You don't need the `noop` function, if the last argument is nil then the action show up as disabled on the popup.

CKolkey (changes requested):
Really really phenomenal work. Most of the comments are just about organising, no real issues with anything substantial. LMK if you have any questions :)

lua/neogit/popups/tag/actions.lua
jarviliam: Is there someway to do that in the config tables in the `neogit/lib/git/cli.lua` 🤔

CKolkey: Hmm... nah, don't think so. It'll match `log.list()` and `remote.list()` anyways, so I think its a good fit in a lib.

CKolkey:
I fixed up some notification related things that I cleaned up recently.

lua/neogit/popups/tag/actions.lua
jarviliam: Is there another way to apply flags dynamically? I need to do `git tag {list of tags} {list of user flags}`

CKolkey: If we compare `arg_list` and `apply_flags`:
```lua
    if action == "arg_list" then
      return function(args)
        for _, v in ipairs(args) do
          table.insert(tbl[k_state].arguments, v)
        end
        return tbl
      end
    end

    if action == "apply_flags" then
      return function(flags)
        for _, v in ipairs(flags) do
          table.insert(tbl[k_state].options, v)
        end
        return tbl
      end
    end
```

And then where they are turned into a command:
```lua
  local function to_process(verbose, suppress_error, ignore_code)
    local cmd = {}

    for _, o in ipairs(state.options) do
      table.insert(cmd, o)
    end

    for _, arg in ipairs(state.arguments) do
      if arg ~= "" then
        table.insert(cmd, arg)
      end
    end

    ...
  end
```

Options just get inserted before arguments, so if you do like:
```lua
git.cli.tag.arg_list(util.merge(args, { value, selected_branch }))
```

That should, unless I'm mistaken, produce the same result.

jarviliam:
Sorry for the delay, I believe I've fixed the issues. Only problem I'm running into now before I think this is good enough to be merged is when doing the tag prune function and there is more than one tag to be pruned locally/remotely, I need to show the user a list of tags to be pruned and then ask for confirmation. Know any good ways to show a list to users or should I just write it out to stdout?

CKolkey:
It looks like how magit goes about this is:
- Select `p` action
- Ask user to select remote (prompt: `Prune tags using remote`)
- Ask for yes/no confirmation for each tag to be pruned.

I think we could improve on that slightly:
```lua
if #tags_to_prune > 1 then
  -- Present options to user with prompt like "{count} tags can be removed locally":
  -- - remove all
  -- - review each
  -- - abort
elseif #tags_to_prune == 1 then
  -- Just ask to remove this tag, the same as "review each" above, for the one
else
  -- Print something like "no tags found"
end
```

lua/neogit/popups/tag/actions.lua
```diff
@@ -0,0 +1,141 @@
...
+    if #selected_tags == 0 then
+      return
+    end
+    tags = selected_tags
+  else
+    local tag_input = input.get_user_input("Tag name:")
+    if not tag_input or tag_input == "" then
+      return
+    end
+    table.insert(tags, tag_input)
```
CKolkey, line 58: I think we can remove this else clause. Two reasons: 
- If `git.tag.list()` doesn't find it.. is there anything to delete?
- The fuzzy finder buffer works as an input when no entries match the prompt. I just added that yesterday. So a user can enter whatever they like, and if there's a prompt but no matches, we use the prompt input.

lua/neogit/lib/git/cli.lua
```diff
@@ -649,6 +666,15 @@ local mt_builder = {
...
 
+    if action == "apply_flags" then
+      return function(flags)
+        for _, v in ipairs(flags) do
+          table.insert(tbl[k_state].options, v)
+        end
+        return tbl
+      end
+    end
+
```
CKolkey, line 677: This can be removed, yea?

lua/neogit/popups/tag/init.lua
```diff
@@ -0,0 +1,27 @@
...
+
+function M.create()
+  local p = popup
+    .builder()
+    :name("NeogitTagPopup")
+    :arg_heading("Arguments")
+    :switch("f", "force", "Force")
+    :switch("a", "annotate", "Annotate")
+    :switch("s", "sign", "Sign")
+    :switch("u", "local-user", "Sign as", { enabled = false })
```
CKolkey, line 14: As this takes user input, it's an `option` not a `switch`. Switch is just for boolean flags.
```suggestion
    :option("u", "local-user", "", "Sign as", { key_prefix = "-" })
```

lua/neogit/popups/tag/init.lua
```diff
@@ -0,0 +1,27 @@
...
+    :new_action_group("Create")
```
CKolkey, line 15: The first action group is implicit, so you don't need to make a new one. You can just use:
```suggestion
    :group_heading("Create")
```

CKolkey (changes requested):

lua/neogit/popups/tag/actions.lua
jarviliam: Ah that makes sense!

CKolkey:
Looks awesome dude! Really appreciate the contribution.

PR #812, opened 2023-09-06
feat: Add support for pushing to Gerrit

dapc11:
As discussed in #811, this is an attempt to add support for being able to push to Gerrit based repositories.

Also found a little quirk, perhaps should be it's own PR. Anyway, it is a fail safe for when the parsing of Git config fails for some reason.


dapc11:
No problem @CKolkey, this is at least good for visibility in case someone else stumble upon the same struggles as I've had. Perhaps I'll come back with a more generic PR, to introduce the interface you're talking about.
Thanks for the help and for a great plugin in the meantime.

PR #821, opened 2023-09-12
Add config option to disable/enable line numbers in the status buffer

mrcjkb:
See https://github.com/NeogitOrg/neogit/discussions/820#discussioncomment-6982271


CKolkey:
Great! Thanks for the PR :)

Issue #824, opened 2023-09-14
Switches that default to enabled do not get remembered as disabled (remember_settings)


star-szr:
### Description

I am using the default `remember_settings` = `true`.

There are some switches like `--graph` and `--decorate` from the log popup that default to enabled. If you execute the popup action or close the popup, these always reset back to enabled (they don't persist the disabled state).

Note that this isn't limited to the log popup, the remote popup also has `-f`, there may be others, I didn't try to find them all since this looks like a wider issue.

### Neovim version

NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3

### Operating system and version

macOS 13.5.2

### Steps to reproduce

1. `nvim -nu minimal.lu`
2. `:Neogit`
3. `l` (open log popup)
4. `-g` (disable graph display)
5. `q` (close log popup)
6. `l` (reopen log popup)

Compare the behaviour to toggling color display and note that color display persists correctly.

### Expected behavior

I would expect that even though the switch defaults to enabled, if I disable the switch it would remember that.

### Actual behavior

The disabled state does not persist. It always switches back to enabled.

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "NeogitOrg/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing neogit and dependencies.")
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
```

CKolkey:
I had a thought on how to deal with this. The whole `enabled = true` flag that some switches have should be passed through the state implementation (`lib.state.get(<name>, <default>)`) as the default option, that way if the user has changed it we just respect that.

PR #828, opened 2023-09-17
test: default switch persistence with remember_settings

star-szr:
References: #824

Feedback very welcome, I will try to take a look at a fix soon but wanted to start with a failing test case.

### DIFF
--- a/lua/neogit/popups/log/actions.lua
+++ b/lua/neogit/popups/log/actions.lua
@@ -7,6 +7,8 @@ local LogViewBuffer = require("neogit.buffers.log_view")
 local ReflogViewBuffer = require("neogit.buffers.reflog_view")
 local FuzzyFinderBuffer = require("neogit.buffers.fuzzy_finder")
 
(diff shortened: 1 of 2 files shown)


CKolkey:
Cool thinking. The _intended_ behaviour is that by setting `enabled = true` in the popup builder, that provides a _default_ value being passed into `lib.state.get()` as the second arg. Putting in some print() calls shows me that this is working.

However, at some point in the pipe after that state is read, the `switch.enabled` value overrules the read state.

Removing `enabled = true` from the builder fixes things... but leaves us without a way to set the default value to "on".

CKolkey:
Well... this was pretty simple, actually. The logic in `lib.state.get()` would do:
`return value or default`. So... if the value is `false` and the default is `true`, then it's always going to use the default.

I'll make a quick patch, then let's see about merging this PR :)

CKolkey:
No notes, this looks great to me. We sorely need test coverage, so, I really appreciate stuff like this :)

Issue #830, opened 2023-09-18
Log always shows ref names (decorate switch doesn't do anything)


star-szr:
### Description

The log buffer always shows ref names even if you turn off the decorate switch from the log popup.

I think this is because [git as of 2.13.0](https://lwn.net/Articles/722364/) defaults to `--decorate=auto`:

> The default behaviour of "git log" in an interactive session has been changed to enable "--decorate".

There are multiple ways to solve this, but I also wonder do we need to support old git versions? Is there a minimum git version requirement? 2.13.0 was released May 10, 2017.

### Neovim version

NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3

### Operating system and version

macOS 13.5.2

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. `:Neogit`
3. `l`
4. Toggle the Show refnames/decorate switch off, if it's not already off
5. `l`

### Expected behavior

I would expect with the Show refnames/decorate switch off that the decorations would not show in the log.

### Actual behavior

Ref names are always shown in the log buffer.

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "NeogitOrg/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing neogit and dependencies.")
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
```

CKolkey:
Oh.. oops. Yeah. Thanks for pointing that out :D

PR #832, opened 2023-09-18
test: coverage for log decorate switch

star-szr:
References: #830 

Also touches up a couple minor things I noticed with the previous test I wrote. Let me know if you have any other suggestions!


CKolkey:
This is great! I'll happily accept any-and-all tests for the various popups :D

PR #839, opened 2023-09-23
test: coverage for log limiting

star-szr:
Also updated to reset switches before each test and close the log popup after each test.

The "respects decorate switch" test assertion has been updated since graph is on by default, and the switches are now being reset before each test.

I should also mention I tried using `tests.mocks.input` instead of the `<C-u>` and such, but it didn’t seem to work, from what I could tell it was not clearing the existing value (default of 256 in this case).

### DIFF
--- a/tests/specs/neogit/popups/log_spec.lua
+++ b/tests/specs/neogit/popups/log_spec.lua
@@ -6,6 +6,7 @@ local util = require("tests.util.util")
 local in_prepared_repo = harness.in_prepared_repo
 
 local status = require("neogit.status")
+local state = require("neogit.lib.state")
 
 local function act(normal_cmd)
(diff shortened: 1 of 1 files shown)


CKolkey:
Yea, `tests.mocks.input` is also a bit imperfect because it's prone to race conditions. When tests are run in parallel, they try to set the module values at the same time and... yeah, you get it, race condition 😓 

Seems like, when possible, `<c-u>` is a better approach. However, I think it's also because the option value input doesn't use the `input` module: https://github.com/NeogitOrg/neogit/blob/f0bd7c736cc242258f0fdce0029f3b6331fe952b/lua/neogit/lib/popup/init.lua#L307-L311

So... That has something to do with it too. 

Anyways, the PR is _much_ appreciated :D

PR #842, opened 2023-09-24
docs: improve minimal init

### LINKED ISSUE #621: Improve minimal init
See discussion [here](https://github.com/NeogitOrg/neogit/pull/616/#discussion_r1264728918)

**Requirements**
Provide an improved minimal init that does a few things:
1. Should not use a shared cache
2. Should not pollute global filesystem
3. Each run should be reasonably separated from previous minimal init usages such that the state is not impacted by previous runs
4. Minimal init should not utilize package managers to install plugins, use built-in functionality to keep the minimal init simple
5. Minimal init *should* persist dependencies (such as plenary) between runs to keep repeated invocations quick *without* changing the behavior of subsequent runs

**Some Solutions**
See [this](https://github.com/NeogitOrg/neogit/pull/616/#discussion_r1264728918) discussion for some ideas tossed around. Currently the comparison should be between [this minimal init](https://github.com/NeogitOrg/neogit/pull/616/#discussion_r1264730696) and [@ten3roberts minimal init](https://github.com/NeogitOrg/neogit/pull/616/#discussion_r1264732449).

If the Neovim flags `-n --clean` do not completely separate each neovim run from another then @ten3roberts minimal init is preferred as it better cleans up between runs.

**Additional context**
Investigate if the Neovim flags `-n --clean` completely separate each Neovim run. Does not share a cache, state, data, config between runs. Only thing that should be shared between runs is the cached downloaded plugin dependencies like plenary.

I'll be doing this investigation when I have a convenient moment to do so.

treatybreaker:
I guess I wasn't gone for long 😉.

### What does this PR do?

This PR significantly overhauls the minimal init in the `ISSUE_TEMPLATE/bug_report.yml` file. It's mostly a heavily
modified version of [@ten3roberts minimal init](https://github.com/NeogitOrg/neogit/pull/616/#discussion_r1264732449).

Per the requirements of #621, the new minimal init does the following:

- [x] Does not use a shared cache
- [x] Does not pollute global filesystem, instead uses a `.min` directory to save files in the current directory
- [x] Separates runs by cleaning up the `xdg` dir when vim exits
- [x] Does not use package managers to install plugins, instead we shell out to git
- [x] Persists dependencies between runs to speed up repeated invocations without modifying those dependencies
  - By default the minimal init downloads:
    - [Neogit](https://github.com/NeogitOrg/neogi)
    - [Telescope](https://github.com/nvim-telescope/telescope.nvim)
    - [Plenary](https://github.com/nvim-lua/plenary.nvim)
    - [Diffview](https://github.com/sindrets/diffview.nvim)

### Notes
- The minimal init does **_not_** setup any of the plugins. That's on the user to do as they may not be reporting a bug with telescope, diffview, etc.

- The new minimal init voids `packpath` and appends the path of each downloaded plugin to the `runtimepath`.

- One thing that we may want to change in this PR is that now the plugins are not saved into a temp directory and instead exist within the same directory as where the minimal init was ran from. If you want me to do a `mktemp` invocation and use a temp directory I can certainly make that happen.

### Fin.

If you like what you see in the minimal init I am thinking about copying some of it and shoving it into `tests/init.lua`
to better handle installing plugins and managing RTP.

Let me know what you think, I'm open to any changes or even an outright rejection of this PR if it is undesired 😃.

Closes #621

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -53,39 +53,87 @@ body:
       description: "Minimal(!) configuration necessary to reproduce the issue. Save this as `minimal.lua`. If _absolutely_ necessary, add plugins and config options from your `init.lua` at the indicated lines."
       render: Lua
       value: |
-        vim.cmd([[set runtimepath=$VIMRUNTIME]])
-        vim.cmd([[set packpath=/tmp/nvim/site]])
-        local package_root = "/tmp/nvim/site/pack"
-        local install_path = package_root .. "/packer/start/packer.nvim"
-        local function load_plugins()
-          require("packer").startup({
-            {
-              "wbthomason/packer.nvim",
-              {
-                "NeogitOrg/neogit",
-                requires = {
-                  { "nvim-lua/plenary.nvim" },
-                  { "sindrets/diffview.nvim" },
-                },
-                config = function()
-                  print("loaded neogit")
-                  require("neogit").setup()
-                end,
-              },
-            },
-            config = {
-              package_root = package_root,
-              compile_path = install_path .. "/plugin/packer_compiled.lua",
-              display = { non_interactive = true },
-            },
+        -- NOTE: See the end of this file if you are reporting an issue, etc. Ignore all the "scary" functions up top, those are
+        -- used for setup and other operations.
+        local M = {}
+
+        local base_root_path = vim.fn.fnamemodify(debug.getinfo(1, "S").source:sub(2), ":p:h") .. "/.min"
+        function M.root(path)
+          return base_root_path .. "/" .. (path or "")
+        end
+
+        function M.load_plugin(plugin_name, plugin_url)
+          local package_root = M.root("plugins/")
+          local install_destination = package_root .. plugin_name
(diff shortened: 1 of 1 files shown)


CKolkey:
_So_ extra to have type annotations hahaha. Not a complaint. I'll try to give this a spin later, and maybe @ten3roberts has thoughts on it too?

treatybreaker:
> _So_ extra to have type annotations hahaha. Not a complaint. I'll try to give this a spin later, and maybe @ten3roberts has thoughts on it too?

It is a *bit* extra, we may want to drop the annotations as it increases the size of the minimal init a fair amount. May be better to *only* have annotations on the `setup` function so a user gets `luals` validation in the one place they may have to modify and cut down on the cruft.

CKolkey:
Haha, I think annotations imply that people should interact with it in some way, which... maybe we don't want to encourage, since this is a "minimal" reproduction template and... now that we're on it... annotations are not "minimal", per-sey. 

I think your idea to cut it down makes a lot of sense.

treatybreaker:
@CKolkey take another look. Should be a bit more concise now on the latest force push.

CKolkey:
This is great.

PR #850, opened 2023-09-26
Bugfix: Error when opening popup

CKolkey:
When opening the branch popup with the cursor over the header, an error would display.

Screenshot 2023-09-26 at 22 30 58

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -804,7 +804,7 @@ function M.get_selection()
 
       local entire_section = section.first == first_line and first_line == last_line
 
-      for _, item in pairs(section.items) do
+      for _, item in pairs(section.items or {}) do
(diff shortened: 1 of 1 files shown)


CKolkey:
@ten3roberts There's probably a more elegant fix here. Ideas?

ten3roberts:
What we could do is always force sections to contain an empty list of items, so that we are consistent.

This is the age old conundrum of `nil/null` or an empty array, same as in javascript and many other dynamically typed languages.

ten3roberts:
This way, your won't have to remember to nil check your list of items, as an empty list conveys the same meaning already, but does not fail but rather loops nothing

ten3roberts:
Opened #851 for the alternate solution 😊

CKolkey:
Yeah, not the biggest fan of nil checks everywhere. Much nicer to just implement the interface as an empty list :)

PR #851, opened 2023-09-27
fix: nil item list when in header section

ten3roberts:
Fixes the error when opening the branch popup in the header section


CKolkey:
Much better! Thanks

PR #852, opened 2023-09-27
test: more tests for log limiting

star-szr:
Trace line evolution not included, possibly buggy.


CKolkey:
If this is something you're keen on doing more of, you have my _full_ support and blessing. Go nuts :D

The big thing for me is that there is some manner of regression testing, which you're accomplishing. For all the features I've added, I just kinda... bench test it... until it's working. But then if some internals get refactored (git config variables, for example), some way to verify that all the consumers of that API don't break is incredibly valuable.

PR #858, opened 2023-09-29
WIP: Fixes finding the right section

### LINKED ISSUE #857: Tab toggle diff toggles previous/above region
### Description

When toggling diffs with tab, the first diff in a file closes as expected, but moving to the next region instead opens the one above. Additionally, moving down several regions, tab will toggle the region two regions up. This happen only if regions above are closed.

https://github.com/NeogitOrg/neogit/assets/38856990/60a34e1d-3197-41ad-89c1-f45a029dcb83

### Neovim version

NVIM v0.9.2
Build type: Release
LuaJIT 2.1.1692716794

ubuntu 22.04

### Steps to reproduce

A file `file.txt` with changes several places.
1. `nvim -nu minimal.lua file.txt`.
2. Open neogit and attempt to toggle diff a region below an already closed region.

### Expected behavior

The region where the cursor is at is toggled.

### Actual behavior

Regions above where the cursor is at is toggled.

### Minimal config

```Lua
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "NeogitOrg/neogit",
        commit = "8727a3222a93d710a7000b750892cc25e2712937", -- Not working
        -- commit = "a48333352092237a968b9f7302f4265666a3ce03", -- Working
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
```

CKolkey:
Fixes #857 

The real changes here are:
```lua
if h.diff_from <= last_line and h.diff_to >= first_line then
```
to
```lua
if h.first <= last_line and h.last >= first_line then
```

That fixes interacting with the correct hunk.
Then:

```lua
if partial then
  from = h.diff_from + math.max(first_line - h.diff_from, 0)
  to = math.min(last_line, h.diff_to)
else
``` 
to
```lua
if partial then
  local cursor_offset = first_line - h.first
  local length = last_line - first_line

  from = h.diff_from + cursor_offset
  to = from + length
else
```

Fixes visual selection's of hunks being staged/unstaged with folded sections above them.

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -692,33 +692,28 @@ end
 ---@param first_line number
 ---@param last_line number
 ---@param partial boolean
----@return SelectedHunk[],string[]
+---@return SelectedHunk[]
 function M.get_item_hunks(item, first_line, last_line, partial)
   if item.folded or item.hunks == nil then
-    return {}, {}
+    return {}
   end
 
   local hunks = {}
-  local lines = {}
 
   for _, h in ipairs(item.hunks) do
-    -- Transform to be relative to the current item/file
-    local first_line = first_line - item.first
-    local last_line = last_line - item.first
-
-    if h.diff_from <= last_line and h.diff_to >= first_line then
-      -- Relative to the hunk
+    if h.first <= last_line and h.last >= first_line then
       local from, to
+
       if partial then
-        from = h.diff_from + math.max(first_line - h.diff_from, 0)
-        to = math.min(last_line, h.diff_to)
+        local length = last_line - first_line
+        from = h.diff_from + math.max((first_line - item.first) - h.diff_from, 0)
+        to = from + length
       else
         from = h.diff_from + 1
         to = h.diff_to
       end
 
       local hunk_lines = {}
-
       for i = from, to do
         table.insert(hunk_lines, item.diff.lines[i])
       end
@@ -734,14 +729,10 @@ function M.get_item_hunks(item, first_line, last_line, partial)
       setmetatable(o, o)
 
       table.insert(hunks, o)
-
-      for i = from, to do
-        table.insert(lines, item.diff.lines[i + h.diff_from])
-      end
     end
   end
 
-  return hunks, lines
+  return hunks
 end
 
 ---@param selection Selection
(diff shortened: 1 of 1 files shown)


CKolkey:
@ten3roberts i think the issue lies in the "if partial" truthy block. I can get it to work with one folded hunk, but not two..

CKolkey:
@ten3roberts I'm going to go ahead and merge this for now since it could do real harm to people - discarding a hunk with a fold will discard the file instead.

If I'm overlooking something, then by all means fix it better :)

ten3roberts:
Oh yes good call

Will look into all this now and see what I can do as well 😊

lua/neogit/status.lua
```diff
@@ -692,33 +692,28 @@ end
...
 
   local hunks = {}
-  local lines = {}
 
   for _, h in ipairs(item.hunks) do
-    -- Transform to be relative to the current item/file
-    local first_line = first_line - item.first
-    local last_line = last_line - item.first
-
-    if h.diff_from <= last_line and h.diff_to >= first_line then
```
ten3roberts, line 709: Right, I think I know what the problem is.

This calculation treated the cursor position from the items linenr in the statusbuffer as the line number in the diff.

This is correct if each diff is laid out after one another like with `git diff`, but when folding the numbers no longer match up.

ten3roberts: Using `.first` which from what I've understood mean `rendered line number of this item in the status buffer` is correct here

lua/neogit/status.lua
```diff
@@ -692,33 +692,28 @@ end
...
-      -- Relative to the hunk
+    if h.first <= last_line and h.last >= first_line then
       local from, to
+
       if partial then
-        from = h.diff_from + math.max(first_line - h.diff_from, 0)
-        to = math.min(last_line, h.diff_to)
+        local length = last_line - first_line
+        from = h.diff_from + math.max((first_line - item.first) - h.diff_from, 0)
+        to = from + length
```
ten3roberts: Will see if this will make it overrun a chunk

Issue #856, opened 2023-09-29
Exception when closing neogit dialog


zbindenren:
### Description

When closing the neogit dialog with `:q` I always get the following error:

```
Error detected while processing BufWipeout Autocommands for "<buffer=5>":                                                                                                                                                                                                                     
E5108: Error executing lua ...nvim/site/pack/packer/start/neogit/lua/neogit/status.lua:640: Vim:E186: No previous directory                                                                                                                                                                   
stack traceback:                                                                                                                                                                                                                                                                              
        [C]: in function 'lcd'                                                                                                                                                                                                                                                                
        ...nvim/site/pack/packer/start/neogit/lua/neogit/status.lua:640: in function 'close'                                                                                                                                                                                                  
        [string ":lua"]:1: in main chunk           
```

### Neovim version

NVIM v0.9.2
Build type: Release
LuaJIT 2.1.1694285958

### Operating system and version

Linux  6.1.55-1-lts #1 SMP PREEMPT_DYNAMIC Sat, 23 Sep 2023 16:57:15 +0000 x86_64 GNU/Linux

### Steps to reproduce

This is my neogit related config.

```lua
local ngit_status_ok, ngit = pcall(require, "neogit")
if not ngit_status_ok then
	vim.notify("neogit plugin not found!")
	return
end

ngit.setup({
	disable_commit_confirmation = true,
	integrations = {
		diffview = true,
	},
})

local group = vim.api.nvim_create_augroup('CloseNeoginOnPush', { clear = true })
vim.api.nvim_create_autocmd('User', {
  pattern = 'NeogitPushComplete',
  group = group,
	callback = ngit.close,
})
```

### Expected behavior

No error.

### Actual behavior

Error as mentioned above.

### Minimal config

```Lua
local ngit_status_ok, ngit = pcall(require, "neogit")
if not ngit_status_ok then
	vim.notify("neogit plugin not found!")
	return
end

ngit.setup({
	disable_commit_confirmation = true,
	integrations = {
		diffview = true,
	},
})

local group = vim.api.nvim_create_augroup('CloseNeoginOnPush', { clear = true })
vim.api.nvim_create_autocmd('User', {
  pattern = 'NeogitPushComplete',
  group = group,
	callback = ngit.close,
})
```

CKolkey:
How are you opening the neogit buffer?

dhananjaylatkar:
Similar error when closing Neogit buffer with `q`

```
[...]

PR #860, opened 2023-09-29
fix: #856 directory stack may be empty when window is closed

ten3roberts:


ten3roberts:
The issue was that `cd -` may not always be able to resolved if the working directory has never truly changed. Decided to keep the beviour of overriding it so that the status buffer does not suddenly get a different cwd through some other plugin kicking in, like `project.nvim`

CKolkey (approved):
_Nice_.

PR #861, opened 2023-09-29
Add ignore popup

ten3roberts:
- [x] Top level gitignore
- [x] Subdirectory
- [x] Private gitignore
- ~~[ ] Ignore extension using glob~~

### DIFF
--- a/lua/neogit/popups/ignore/actions.lua
+++ b/lua/neogit/popups/ignore/actions.lua
@@ -1,57 +1,37 @@
 local M = {}
+local Path = require("plenary.path")
 
----@param path string
+---@param path
 ---@param rules string[]
 local function add_rules(path, rules)
-  local async = require("plenary.async")
-  local notification = require("neogit.lib.notification")
-
-  local err, fd = async.uv.fs_open(path, "a+", 438)
-  if err then
-    return notification.error(string.format("Failed to read ignore file %q\n\n%s", path, err))
-  end
-
-  async.uv.fs_write(fd, table.concat(rules, "\n") .. "\n")
-
-  local err = async.uv.fs_close(fd)
-  if err then
-    return notification.error(string.format("Failed to close file %q", path))
-  end
-
-  -- notification.info(string.format("Added %d rules to %q", #rules, path))
+  path:write(table.concat(rules, "\n") .. "\n", "a+", 438)
 end
 
 local operation = require("neogit.operations")
 M.shared = operation("ignore_shared", function(popup)
-  local Path = require("plenary.path")
-
   local git_root = popup.state.env.git_root
   local rules = vim.tbl_map(function(v)
-    return tostring(Path:new(v):make_relative(git_root))
+    return Path:new(v):make_relative(git_root)
   end, popup.state.env.paths)
 
-  add_rules(git_root .. "/.gitignore", rules)
+  add_rules(Path:new(git_root, "/.gitignore"), rules)
 end)
 
 M.private = operation("ignore_private", function(popup)
-  local Path = require("plenary.path")
-
   local git_root = popup.state.env.git_root
   local rules = vim.tbl_map(function(v)
-    return tostring(Path:new(v):make_relative(git_root))
+    return Path:new(v):make_relative(git_root)
   end, popup.state.env.paths)
 
-  add_rules(git_root .. "/.git/info/exclude", rules)
+  add_rules(Path:new(git_root .. "/.git/info/exclude"), rules)
 end)
 
 M.at_subdirectory = operation("ignore_subdirectory", function(popup)
-  local Path = require("plenary.path")
-
   for _, path in ipairs(popup.state.env.paths) do
     local path = Path:new(path)
-    local parent = tostring(path:parent())
+    local parent = path:parent()
 
-    add_rules(tostring(parent) .. "/.gitignore", { tostring(path:make_relative(parent)) })
+    add_rules(path:parent() .. "/.gitignore", { path:make_relative(parent) })
   end
 end)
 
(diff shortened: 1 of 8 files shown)


ten3roberts:
Alrighty, the ignore popup, or at least the beginnings of it is done

lua/neogit/popups/ignore/actions.lua
```diff
@@ -0,0 +1,58 @@
...
+  if err then
+    return notification.error(string.format("Failed to read ignore file %q\n\n%s", path, err))
+  end
+
+  async.uv.fs_write(fd, table.concat(rules, "\n") .. "\n")
+
+  local err = async.uv.fs_close(fd)
+  if err then
+    return notification.error(string.format("Failed to close file %q", path))
+  end
```
CKolkey, line 19: Is there an advantage to using this low-level API versus the plenary Path abstraction? I had a lot of success in the past replacing these type of calls with Path.

lua/neogit/popups/ignore/actions.lua
```diff
@@ -0,0 +1,58 @@
...
+
+  -- notification.info(string.format("Added %d rules to %q", #rules, path))
```
CKolkey, line 21: Remove?

lua/neogit/popups/ignore/actions.lua
```diff
@@ -0,0 +1,58 @@
...
+end
+
+local operation = require("neogit.operations")
+M.shared = operation("ignore_shared", function(popup)
```
CKolkey, line 25: I have a feeling `git.repo.git_path()` could be of use here: https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/lib/git/repository.lua#L15-L17

tests/util/git_harness.lua
```diff
@@ -69,9 +69,14 @@ end
...
-
-  local lines = vim.split(result.stdout, "\n")
+  -- Wake me up when when this get into stable
+  -- local result = vim.system(cmd, { text = true }):wait()
+  -- local lines = vim.split(result.stdout, "\n")
+  -- return lines
+
+  local lines = vim.fn.system(cmd)
+  local lines = vim.split(lines, "\n")
+  -- return lines
```
CKolkey, line 79: Can remove this line, yea?

CKolkey:
Some questions about the FS API calls, and some cleanup, but all-in-all looks good :)

lua/neogit/popups/ignore/actions.lua
ten3roberts: No, the only one I can think of is error handling, rather than the asserts inside `Path:write`.

How do you go about the docs for plenary.path? I usually read the source file, but it is easy to miss things that way.

lua/neogit/popups/ignore/actions.lua
ten3roberts: Fixed :smile:

ten3roberts:
Thanks for the suggestions, I've fixed all of them up now :blush:

lua/neogit/popups/ignore/actions.lua
```diff
@@ -0,0 +1,42 @@
...
+---@param path
+---@param rules string[]
+local function add_rules(path, rules)
+  path:write(table.concat(rules, "\n") .. "\n", "a+", 438)
+end
+
+local operation = require("neogit.operations")
+M.shared = operation("ignore_shared", function(popup)
+  local git_root = git.repo.git_root
+  print("git root", git_root)
```
CKolkey, line 14: Remove Before Flying™️

lua/neogit/status.lua
```diff
@@ -1031,10 +1031,17 @@ local set_folds = function(to)
 end
 
 --- Handles the GoToFile action on sections that contain a hunk
----@param item StatusItem
+---@param item File
 ---@see section_has_hunks
 local function handle_section_item(item)
-  local path = item.name
+  local path = item.absolute_path
+  print("Opening: " .. path)
```
CKolkey, line 1038: Remove Before Flying™️

tests/specs/neogit/popups/ignore_spec.lua
```diff
@@ -0,0 +1,39 @@
...
+  vim.fn.feedkeys(vim.api.nvim_replace_termcodes(normal_cmd, true, true, true))
+  vim.fn.feedkeys("", "x") -- flush typeahead
+  status.wait_on_current_operation()
+end
+
+describe("ignore popup", function()
+  it(
+    "top level ignore",
+    in_prepared_repo(function()
+      print("Here")
```
CKolkey, line 21: And this

ten3roberts:
Right, fixed up some imports that were unused as well 😄

CKolkey:
Code looks fine - I've got a question on usage. In the status buffer, with no untracked files, I press `i t` and `.gitignore` is modified to have a new empty line at the end. That... probably shouldn't happen?

If I have the cursor over a filepath, that gets appended to the `.gitignore` fine, so I think it's just an issue of an `""` being treated as a valid path, maybe? Maybe something else.

Should the fuzzy finder get invoked here? You could do something like https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/popups/branch/actions.lua#L64-L66 

(maybe you have that planned in a revision)

ten3roberts:
> If I have the cursor over a filepath, that gets appended to the `.gitignore` fine, so I think it's just an issue of an `""` being treated as a valid path, maybe? Maybe something else.

Most likely, the inteded usage is a visual selection or a cursor on over the file(s) you want to ignore, similar to stage. With my refactoring of the selection api this also works if you do `i t` on the `Untracked` section header. I'll fix the empty case :)

CKolkey (approved):
Looks good!

Issue #864, opened 2023-10-02
[Feature] Ability to specify popup arguments from `:Neogit` command


b0o:
I have been using Fugitive for years and have developed muscle memory for many of my Fugitive-based mappings. For example:

```vim
nmap <leader>gcc :Git commit --verbose<cr>
nmap <leader>gca :Git commit --verbose --all<cr>
nmap <leader>gcA :Git commit --verbose --amend<cr>
``` 

I would like to transition fully over to Neogit, but I want to keep my workflow more or less the same. I propose being able to pass arguments to Neogit popups directly from the command line, like so:

```vim
nmap <leader>gcc :Neogit commit -v c<cr>
nmap <leader>gcA :Neogit commit -a -v c<cr>
nmap <leader>gcA :Neogit commit -v a<cr>
```

I imagine it would work like this:

If an argument is passed, it is pre-toggled in the Neogit popup.

If an action is passed, like `c` for `commit` or `a` for `amend`, the popup would not be displayed at all.

CKolkey:
First off, thanks for the idea. However... Neogit isn't a fugitive replacement, in terms of how it's used. The two plugins have different intended modes of interaction. Neogit is a git client in the spirit of https://magit.vc/, driven by popup buffers. Honestly, you miss out on so much by just invoking the popups directly, I don't want to make it so you can skip THOSE.

For example, almost all popups will consume whatever commit(s) are selected or under the cursor. You completely miss this if you invoke a popup directly.

Honestly, if it wasn't already part of the API, I never would have made it possible to invoke a popup without going through the status buffer, but it's there already, so, it stays.

Of course, this is not for everyone! So no worries there, but.. If you like fugitive, why not keep using it?

Also, there are some technical reasons this would be tough to accomplish. If you're _really_ keen, you could call the actions directly `neogit/lua/popups/*/actions.lua` and pass a popup object that responds to `:get_arguments()`. That would have to return the table of arguments/options `{ "--verbose", "--all" }`.

b0o:
Hey, thanks for the quick response.

> You miss out on so much by just invoking the popups directly

Yeah, absolutely. I usually *do* use the popups, and I mostly interact with Neogit through the `:Neogit` buffer. But, sometimes, I stage hunks using gitgutter, and I know I want to go straight to creating a commit, so I habitually do `<leader>gcc`.

> If you like fugitive, why not keep using it?

[...]

PR #865, opened 2023-10-02
Add experimental API for accessing popup actions directly.

CKolkey:
https://github.com/NeogitOrg/neogit/issues/864
Credit to @b0o

@ten3roberts How do you feel about this?

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -111,6 +111,42 @@ function M.open(opts)
   end
 end
 
+-- EXPERIMENTAL!
+-- This can be used to create bindable functions for custom keybindings:
+--   local neogit = require("neogit")
+--   vim.keymap.set('n', '<leader>gcc', neogit.action('commit', 'commit', { '--verbose', '--all' }))
+--
+---@param popup  string Name of popup, as found in `lua/neogit/popups/*`
+---@param action string Name of action for popup, found in `lua/neogit/popups/*/actions.lua`
+---@param args   table? CLI arguments to pass to git command
+---@return function
+function M.action(popup, action, args)
+  local notification = require("neogit.lib.notification")
+  local a = require("plenary.async")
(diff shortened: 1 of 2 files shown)


lua/neogit.lua
```diff
@@ -111,6 +111,42 @@ function M.open(opts)
...
+---@param action string Name of action for popup, found in `lua/neogit/popups/*/actions.lua`
+---@param args   table? CLI arguments to pass to git command
+---@return function
+function M.action(popup, action, args)
+  local notification = require("neogit.lib.notification")
+  local a = require("plenary.async")
+
+  return function()
+    a.run(function()
+      local ok, actions = pcall(require, "neogit.popups." .. popup .. ".actions")
```
ten3roberts, line 129: How will this handle populating the `env` and `popup` field of the actions first parameter?

Will the user need to supply a dummy `Popup` instance?

ten3roberts (changes requested):

lua/neogit.lua
CKolkey: The `env` gets skipped entirely: this bypasses the popup, and all associated state, to allow easy access to the _actions_.

They user can pass a table of CLI flags that populate the return value of a mock `get_arguments()` function. This _won't_ work for stuff that we use internally like `--graph`, `--color`, and `--decorate` for `git log`, but thats it.

The value of exposing an API like this is that a user could pre-compose a set of CLI arguments and bind that to a key. Then, they still get all the advantages of neogit like the `client.wrap` RPC stuff, more custom function like `spin_out`, etc. 

_I_ still think the UI is the way to go, but I don't see the harm in this. The biggest downside of exposing this is that we're making the popup names/action names part of the public api, and changing them becomes a breaking change. Which... I think thats fine. Don't see why we would rename them.

Issue #469, opened 2023-03-12
Neogit can't start up after PR #453 on Windows


Strongleong:
### Description

After merging PR #453 Neogit does not work on Windows anymore. 
Here is error message with `Lazy.nvim`:
```
Failed to run `config` for neogit
...a/Local/nvim-data/lazy/plenary.nvim/lua/plenary/path.lua:497: We couldn't mkdir: C:\Users\strongleong\AppData\Local\nvim-data/neogit/C:
# stacktrace:
  - plenary.nvim\lua\plenary\path.lua:497 _in_ **mkdir**
  - plenary.nvim\lua\plenary\path.lua:630 _in_ **touch**
  - neogit\lua\neogit\lib\state.lua:51 _in_ **read**
  - neogit\lua\neogit\lib\state.lua:33 _in_ **setup**
  - neogit\lua\neogit.lua:61 _in_ **setup**
  - neogit\lua\neogit.lua:76
  - ~\AppData\Local\nvim/lua/lazy-init.lua:16
  - ~\AppData\Local\nvim\init.lua:5
```

Error message with `minimal.lua`
```
packer.nvim: Error running config for neogit: ...site\pack\packer\start\plenary.nvim/lua/plenary/path.lua:497: We couldn't mkdir: C:\Users\strongleong\AppData\Local\nvim-data/neogit/C:
```

P.S. Maybe add CI tests for Windows? 

### Neovim version

NVIM v0.8.2
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

### Operating system and version

Windows 10

### Steps to reproduce

1. `nvim -nu minimal.lua`

### Expected behavior

Neovim opens as usual, Neogit works properly 

### Actual behavior

Neovim opens with error, Neogit is not loaded

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
[...]

PR #877, opened 2023-10-08
format drive letter for per_project_settings

setibeings:
Absolute paths in windows always start with a drive letter which has to be stripped or formatted. As far as I can tell from my testing this makes detecting the platform unnecessary, as the base path and filename don't need further modification, and stripping a drive letter that isn't there is harmless.

This resolves issue https://github.com/NeogitOrg/neogit/issues/469

### DIFF
--- a/lua/neogit/lib/state.lua
+++ b/lua/neogit/lib/state.lua
@@ -16,12 +16,7 @@ function M.filepath()
   local filename = "state"
 
   if config.values.use_per_project_settings then
-    filename = vim.loop.cwd():gsub("/", "%%")
-  end
-
-  if vim.loop.os_uname().sysname == "Windows_NT" then
-    base_path = base_path:gsub("/", "\\")
-    filename = filename:gsub("\\", "%%")
+    filename = vim.loop.cwd():gsub("^%a:", ""):gsub("/", "%%")
   end
 
   return Path:new(base_path .. filename)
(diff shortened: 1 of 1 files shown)


setibeings:
I want to take a crack at writing or rewriting some tests around this, so if you were to hold off til I get time to take another look (a day or two, I hope) I can better demonstrate that this code works. I probably should have done it before opening the PR, but I would still like to do it now.

CKolkey:
Sure, fine with me :)

wrightjjw:
Is plenary path normalize not sufficient in this case?

setibeings:
well, the branch name isn't really accurate anymore, I realized that simply discarding drive letters would be bad say if somebody has a project with the same folder name on the roots of different drives. I never wrote automated tests, but in my manual testing the files get equivalent names to what they got before on linux. I didn't test it on windows with `:set shellslash` because that setting works poorly with other packages I have installed, but I don't know of a reason that this change in particular would break that option in neovim.

setibeings:
Although I wanted to write unit tests, this could probably be merged as is, if there isn't a hard requirement to unit test all code changes. An end to end manual test of this fix is just to check that the state file is created with the expected name, in the expected place, and of course that Neovim doesn't crash on startup when `vim.loop.os_uname().sysname` returns something other than"Windows_NT" on windows.

setibeings:
I applied the changes from stylua

CKolkey:
Looks good to me! Thanks for the fix.

PR #883, opened 2023-10-09
feat: interactive rebase should include current commit

star-szr:
As discussed in https://github.com/NeogitOrg/neogit/discussions/862

I really tried to write a test for this, but couldn't figure out how to wait until the `git-rebase-todo` was present to check its contents. Maybe later :)

Feedback very welcomed, as always.

### DIFF
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -483,6 +483,15 @@ local configurations = {
     },
   },
 
+  ["rev-list"] = config {
+    flags = {
+      parents = "--parents",
+    },
+    options = {
+      max_count = "--max-count",
+    },
+  },
+
   ["rev-parse"] = config {
     flags = {
       verify = "--verify",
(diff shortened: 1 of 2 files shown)


CKolkey:
Is something like this useful? https://stackoverflow.com/questions/2221658/whats-the-difference-between-head-and-head-in-git

I was imagining (though maybe I'm wrong) that using `<commit>^` would work here.

CKolkey:
Ah, ok, that makes sense. Thanks for the explanation :)

CKolkey:
Maybe as a commit message, or message here, but inline can get... overwhelming.

CKolkey:
> Breaking news: I let it run (only the status_spec test) while I was in a meeting and it failed on run 895 in the same way.

As they say, fool me 894 times, shame on you. Fool me 895 times, shame on me.

The state-with-side-effects nature of testing a local git repo is... rough sometimes.

Issue #887, opened 2023-10-11
Random test failure in tests/specs/neogit/status_spec.lua


star-szr:
### Description

Occasionally one of the assertions in `tests/specs/neogit/status_spec.lua` will fail.

Example: https://github.com/NeogitOrg/neogit/actions/runs/6480916741/job/17597408980

### Neovim version

n/a

### Operating system and version

n/a

### Steps to reproduce

1. Set the environment variable `TEST_FILES` to `tests/specs/neogit/status_spec.lua`
2. Run `make test` many times (for me it failed on run 895 but being a random fail, this will vary, the second time I ran it in a loop it failed on run 1102)

### Expected behavior

The assertion will pass consistently.

### Actual behavior

Sometimes, this failure happens:

```
 Fail	||	status buffer unstaging files - u can unstage from a selection in a hunk	
            ...enary.nvim/start/plenary.nvim/lua/plenary/async/util.lua:37: Blocking on future timed out or was interrupted.
            ./tests/util/git_harness.lua:63: ...er/work/neogit/neogit/tests/specs/neogit/status_spec.lua:178: Expected objects to be the same.
            Passed in:
            (string) 'M  b.txt
            '
            Expected:
            (string) 'MM b.txt
            '
            
            stack traceback:
            	...enary.nvim/start/plenary.nvim/lua/plenary/async/util.lua:37: in function 'block_on'
            	./tests/util/git_harness.lua:60: in function <./tests/util/git_harness.lua:55>
```

### Minimal config

```Lua
-- NOTE: See the end of this file if you are reporting an issue, etc. Ignore all the "scary" functions up top, those are
-- used for setup and other operations.
local M = {}

local base_root_path = vim.fn.fnamemodify(debug.getinfo(1, "S").source:sub(2), ":p:h") .. "/.min"
function M.root(path)
  return base_root_path .. "/" .. (path or "")
end

function M.load_plugin(plugin_name, plugin_url)
  local package_root = M.root("plugins/")
  local install_destination = package_root .. plugin_name
  vim.opt.runtimepath:append(install_destination)

  if not vim.loop.fs_stat(package_root) then
    vim.fn.mkdir(package_root, "p")
  end

  if not vim.loop.fs_stat(install_destination) then
    print(string.format("> Downloading plugin '%s' to '%s'", plugin_name, install_destination))
    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
[...]

PR #888, opened 2023-10-14
test: fix random test fail in status spec

star-szr:
After reviewing the test and the code in [lua/neogit/status.lua](https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/status.lua) I noticed that stage and discard set `M.current_operation` but unstage does not.

The assertion for unstaging that occasionally fails comes right after an `act()` line, and the last line in `act()` is `status.wait_on_current_operation()`, which checks if `status.current_operation` is set.

## Local testing results

I tested locally with just the four unstage tests (the other tests in the status spec were commented), and without this change I ran the test in a loop three times and it failed at runs 122, 119, and 74. With the fix in place, I was not able to reproduce a failure. I ran the test in a loop two times and stopped the loop at runs 1130 and 1360, satisfied that the random failure has been squashed.

Fixes: #887

### DIFF
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -879,6 +879,8 @@ local stage = function()
 end
 
 local unstage = function()
+  M.current_operation = "unstage"
+
   local selection = M.get_selection()
   local mode = vim.api.nvim_get_mode()
 


ten3roberts (commented):
The all too familar one line bugfix.

Thanks for your search and contribution.

~Could you use the newer operation wrapper instead, which makes sure to reset in in the end too?~

~https://github.com/NeogitOrg/neogit/blob/d3fbf8b48f4f8bc406eb23f8b06eb912776f8cb0/lua/neogit/popups/ignore/actions.lua#L16~

~Because I am not sure if it resets on every branch it to false/nil at the end~

ten3roberts (approved):
My bad, I see they are two different albeit similar APIs, old vs new.

The current is good, and it is reset without any apparent early returns that are missed if you expand the whole function 😄

Merging these two wait APIs are definitely out of scope for this PR 😅

PR #889, opened 2023-10-15
Implemented push tags action

OZoneGuy:
Implemented command to push all tags.


CKolkey:
Nice! Love when it's that simple. Thanks for the contribution :)

PR #890, opened 2023-10-15
test: remove status wait api in favor of generic operations

star-szr:
Seems pretty straightforward but let me know if anything could be improved!

References: #888


CKolkey:
Oh, nice, I like this a lot.

Issue #870, opened 2023-10-03
Enter on a submodule should open a Status buffer *inside* the modules cwd


ten3roberts:
**Is your feature request related to a problem? Please describe.**
Changes in a submodule show as changed content (with an extra .S flag)

![image](https://github.com/NeogitOrg/neogit/assets/25723553/b6e2298b-3e68-4064-8298-1506a74ac6a1)

**Describe the solution you'd like**
Clicking `<cr>` on the submodule should open neogit's status buffer inside the submodule, allowing you to commit its changes.

**Describe alternatives you've considered**
`:Neogit cwd=<path_to_submodule>`

CKolkey:
Always fun when you try to do something at work and it's not supported, then you remember it's _your_ project XD

ds82:
my workaround until submodule support lands:
```
		keymap.set("n", "<leader><leader>g", function()
			local file_dir = vim.fn.expand("%:h")
			local exec_result = vim.system({ "git", "rev-parse", "--show-toplevel" }, { text = true, cwd = file_dir })
				:wait()
			local root = vim.fn.trim(exec_result.stdout)
			neogit.open({ kind = "replace", cwd = root })
		end, { desc = "Neogit status", silent = true })
```
`vim.system` requires neovim nightly!

PixsaOJ:
This should be in `README`, works great!

PR #893, opened 2023-10-16
Integrate with submodules

ten3roberts:
This enables neogit to work with submodules

image

Fixes: #870


lua/neogit/status.lua
```diff
@@ -289,18 +306,26 @@ local function draw_buffer()
...
         elseif f.mode then
-          output:append(string.format("%s %s", label, f.name))
+          line = string.format("%s %s", label, f.name)
         else
-          output:append(f.name)
+          line = f.name
+        end
+
+        if f.submodule then
+          line = line .. " " .. format_submodule_mode(f.submodule)
```
CKolkey, line 324: Definitely `string.format` here. Lua performance can go to shit with a lot of string concatenation.

ten3roberts:
In order to get submodules working properly I found it necessary to make the status buffer not be a global state but rather instance based. The reason is that opening and going back and forth between submodules status panes would make these status buffers trample that same global state.

This approach makes the status buffer an object which can be manipulated and returned/found based on the git repositories root, and will in theory allow having multiple instances of neogit for different repositories open at the same time in tabs, splits, or the background.

CKolkey:
Cool, the change makes sense. I always thought the global status buffer thing was weird conceptually, but :shrug: it's from before my time here.

I'd say don't worry too much about the status buffer - I'm going to be replacing it soon with a rewrite built using the UI framework. I finally got a couple building blocks in place to make that work:
- `on_open` callback function for "folded" content for lazy-loading diffs
- an index built into the renderer so finding components under the cursor is (nearly) `o(1)`. All components are indexed by line number in a table, so finding the right one is much faster than iterating through all components starting at line 1.

Still a WIP... lots of functions to pull in from the existing buffer. If you're up for it I wouldn't mind help :) But all good either way.

https://github.com/NeogitOrg/neogit/pull/1161

ten3roberts:
Awesome!

Thanks for the heads up

I think both of these rewrites would go quite well together.

Will daily run this one for a bit and see how it fairs and fix any issues (and get the tests to pass).

Have some more time and mental energy now to pursue my side projects, which feels fantastic :smile: 

Don't worry too much about merge conflicts, as the Swedish saying goes: "As one makes their bed, so they must lie"

CKolkey:
Given the number of conflicts, I think it's safe to assume this can be closed. However, I've copied the core change - displaying submodule status - here: 51450329e04860f71a31d8c7ef5178fce6e60524

PR #894, opened 2023-10-17
Add show log with reflog action to log command

guillaumeboehm:
Hi, I feel like the git reflog command is very confusing and always find myself using git log --reflog instead.
If adding this action to the log command is not too confusing I'd really like to be able to use it directly in Neogit.

If it is too confusing, could we maybe explorer adding the flag in the same fashion as --graph and such ? Although I don't really see where it would fit.


CKolkey:
I'm a little confused - I don't see much difference between `git log HEAD` and `git log HEAD --reflog`. What is the current reflog view missing?

guillaumeboehm:
> I'm a little confused - I don't see much difference between `git log HEAD` and `git log HEAD --reflog`. What is the current reflog view missing?

The reflog command isn't missing anything, it just is a nightmare to read I believe. As for the difference between log and log --reflog, try doing a quick git commit --amend and you'll see the dangling older commit in your log tree.

CKolkey:
Screenshot 2023-10-17 at 10 10 44

(magit left, neogit right)

I'm a bit perplexed - how it is a nightmare? Genuinely curious.

guillaumeboehm:
Yes on a tiny streamline repo it is not confusing at all grant you, but on a big repo with several local branches, dangling commits coming from different branches etc, I think it gets really confusing real fast.

guillaumeboehm:
For example I did some convoluted operations here and there on my dotfiles repo where I keep an x11 branch and a wayland branch on top of master.

With git log --reflog it is pretty easy by looking at the branching paths and commit names to know what I did and look for a commit in particular let's say.


With git reflog I find it much more complicated, the information is there sure, and you even get a bit more context with the exact operations being performed. But it is much heavier and less obvious in my opinion.

Issue #900, opened 2023-10-22
Unable to stage hunk in other root directory


ten3roberts:
So this is still an issue actually. It seems like it goes hand-in-hand with #691.

Steps to produce:
1. Open nvim inside a project where git is initialized that's not the root folder.
> In my case I would nvim open `dotfiles/.config/nvim`. But the root folder is the `dotfiles`
2. Open up `neogit`
2. Try hunk stashing a change.
3. I get an error saying:
```
git --no-pager -c color.ui=always --no-optional-locks apply --cached:
    error: packages/common/.config/nvim/../../../mac/Brewfile: does not exist in index

Open the console for details
Git Error (1), press $ to see the git command history
```

What fixes the issue is opening nvim on the root folder where git is initialized. In my case, that's the `dotfiles/` folder. If you need more info let me know. I don't mind helping out the most I can.

_**SideNote: Just used `nvim-tree` to jump back up to my root folder and that worked as well. Granted that defeats the purpose since sadly I would've opened neovim there to begin with.**_

_Originally posted by @CarlosMed in https://github.com/NeogitOrg/neogit/issues/395#issuecomment-1773991615_

ten3roberts:
Thank you for reopening the issue. I moved it to a new issue.

Is it only when hunk staging or file staging as well, as well as other git operations.

Hugobsb:
@ten3roberts Hello 👋 the same is happening to me with the latest version in `master`. Is it happening to you?

PR #901, opened 2023-10-22
fix: git apply denies paths outside the current working directory

ten3roberts:
Fixes: #900


CarlosMed:
Just tried it out and it works! Thank you for that quick fix.

PR #904, opened 2023-10-24
feat: Prefill branch rename prompt with the selected branch name

bloznelis:
I feel like the most common use case for branch renames is fixing typos. For me (and likely others) DX would be greatly improved if the prompt would come already prefilled with the name of the branch which is being renamed.

To our luck `input.get_user_input` already supports default value as its second parameter, DRY code achieved!


CKolkey:
Heyyy, yeah, I'm onboard. Nice job ;)

Issue #915, opened 2023-11-04
"open pull request" opens wrong URL when remote contains many path segments


miduddin:
### Description

GitLab for example allows projects to be under a (nested) group. For example with the remote URL `git@gitlab.com:org/group/project`, neogit will open a pull request on `https://gitlab.com/org/group` when it should be on `https://gitlab.com/org/group/project` instead.

### Neovim version

NVIM v0.10.0-dev-4c32927
Build type: Release
LuaJIT 2.1.1697887905

### Operating system and version

Linux (WSL)

### Steps to reproduce

On a git project with more than 2 path segments in its URL (e.g. `https://gitlab.com/org/group/project`), open a pull request with neogit (e.g. `bo` from `:Neogit` buffer).

### Expected behavior

Correct pull request URL is opened, e.g. `https://gitlab.com/org/group/project/merge_requests/new?merge_request[source_branch]=some_branch`

### Actual behavior

Wrong pull request URL is opened, e.g. `https://gitlab.com/org/group/merge_requests/new?merge_request[source_branch]=some_branch`

### Minimal config

```Lua
I don't have one ready at the moment, but will try to provide when further info is needed.
```

miduddin:
I've tried to fix it in https://github.com/miduddin/neogit/commit/c486f262df1a29ef8ca327422529de2e920d29e6 which seems to work well with my use cases, but I haven't figured out how to make the tests for it, is it ok to just make a pull request as is?

CKolkey:
Cool, no need for a minimal config, thats a pretty clear problem. Feel free to open a PR and I'll have a look. This also seems like something that should be pretty straightforward to write specs for

PR #916, opened 2023-11-04
fix open pull request with many path segments in URL

miduddin:
Fixes https://github.com/NeogitOrg/neogit/issues/915


CKolkey:
Good stuff, thanks for the contribution :)

PR #917, opened 2023-11-05
issue #469 neogit can't start on windows

ray-x:
> local uv = vim.uv or vim.loop

vim.loop deprecated in neovim 0.10

> or os_name:find("MINGW")

for `mingw/msys`

> filename = filename:gsub("^C:", "C%%")

replace "C:/folder/file" to "C%folder/file" as `: (colon)` is reserved in windows

### DIFF
--- a/lua/neogit/lib/state.lua
+++ b/lua/neogit/lib/state.lua
@@ -12,15 +12,17 @@ end
 
 ---@return Path
 function M.filepath()
+  local uv = vim.uv or vim.loop
   local base_path = vim.fn.stdpath("state") .. "/neogit/"
   local filename = "state"
 
   if config.values.use_per_project_settings then
-    filename = vim.loop.cwd():gsub("/", "%%")
+    filename = uv:cwd():gsub("/", "%%")
   end
-
-  if vim.loop.os_uname().sysname == "Windows_NT" then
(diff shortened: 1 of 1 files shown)


setibeings:
can you still reproduce the issue on the master branch?

setibeings:
I believe I've made the platform detection here unnecessary with my recent changes to the same area of the code.

https://github.com/NeogitOrg/neogit/commits/master/lua/neogit/lib/state.lua

What error, if any, do you get if you use the fully up-to-date version of neogit?

ray-x:
The issue I see now is different. Each time I am promoted I need to Initialize a repository in my working folder where `.git` exists.

setibeings:
The code you changed in this pull request has conflicting changes with my PR that I had opened before this one was opened, and that addresses the same initial issue. I noticed your code also uses `vim.uv` which was a change I neglected to make. I asked elsewhere about the minimum version, and depending on the answer, using `vim.uv` all the time might be the way to go. https://github.com/NeogitOrg/neogit/discussions/953

Opening a new issue would be a good way for you to clarify the new issue you're seeing.

ray-x:
Submit a issue here: https://github.com/NeogitOrg/neogit/issues/954

PR #929, opened 2023-11-08
refactor: set Neogit `cwd` to git root if `cwd` unspecified

### LINKED ISSUE #906: Allow to change how the relative paths are shown
When we have untracked directories and files and we are not at the git root, neogit shows them using relative paths to the current directory:

Captura de Tela 2023-10-25 às 12 41 16

Hence, it is sometimes very difficult to see what we have to stage.

It should be possible to make neogit always show the paths with respect to the git root, as Magit does:

Captura de Tela 2023-10-25 às 12 42 06

I did not consider any alternative.

treatybreaker:
This sets the Neogit `cwd` to the git root if left unspecified. This makes Neogit more in line with Magit in its presentation of files.

Some sceenshots below.
As a note, my git root is `~/dots` for the repo I am demonstrating this in and I am opening Neovim in my `dots/.config/neovim/lua` directory.

Before:
image

After:
image

Closes #906

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -67,7 +67,7 @@ function M.open(opts)
   end
 
   if not opts.cwd then
-    opts.cwd = vim.fn.getcwd()
+    opts.cwd = require("neogit.lib.git.cli").git_root()
   end
 
   if not did_setup then


CKolkey:
Simple! Nice idea.

Issue #931, opened 2023-11-09
Changing CWD breaks action on hunks


smatting:
### Summary

If you start nvim and  `:cd` into a different git repository you have a Neogit where all operations on hunks are broken with errors such

```
git --no-pager -c color.ui=always --no-optional-locks apply --cached:

    error: repos/neogit/lua/neogit/lib/git/index.lua: does not exist in index
```
where the wrong filename is an incorrect relative path to the git repo.

This is very confusing, because other operations such as staging whole files or making commits work fine for the git repo you just `:cd` 'ed into.

### Analysis

When `setup()` is called `neogit.lib.git.repository` gets initialized and `git_root` is set. When you `:cd` into another git repo and `open()` Neogit then the target repo is the one at CWD, however `neogit.lib.git.repository` is still at the state of repo determined at `setup()`. If you try to stage a hunk `neogit.lib.git.index.apply` is called with `{ use_git_root = true }` which will in turn use the outdated `neogit.lib.git.repository.git_root` causing the bug.

### Neovim version

I'm using master:
```
commit d3c5687a78cffc25026ff31d2fcbd61a2a3e067f (HEAD -> master, origin/master, origin/HEAD)
Merge: 29583c2b c105aa7c
Author: Cameron <Alleyria@gmail.com>
Date:   Wed Nov 8 09:00:32 2023 +0100

    Merge pull request #929 from treatybreaker/git-root-cw
```

### Operating system and version

linux

### Steps to reproduce

1. Start neovim in $HOME (or any other directory that is _NOT_ the git repo you want to work in)
```
$ cd $HOME
$ nvim
```

2. Change directory to a git repo that you want to work in, e.g.
```
:cd ~/repos/neogit
```

3. Start Neogit
```
:Neogit
```

4. Try to stage a _hunk_. This will fail with

```
git --no-pager -c color.ui=always --no-optional-locks apply --cached:

    error: repos/neogit/lua/neogit/lib/git/index.lua: does not exist in index
    
Open the console for details
```

Note the filename is `repos/neogit/lua/neogit/lib/git/index.lua`, but the correct file relative to the repo would be `lua/neogit/lib/git/index.lua`.

### Expected behavior

`:cd` into a git repo and then starting neogit should give you avneogit for the repo at cwd where everything works.

### Actual behavior

All hunk operations are broken

### Minimal config

```Lua
-- NOTE: See the end of this file if you are reporting an issue, etc. Ignore all the "scary" functions up top, those are
[...]

PR #932, opened 2023-11-09
Always reset status when opening neogit

smatting:
Fixes https://github.com/NeogitOrg/neogit/issues/931

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -106,6 +106,7 @@ function M.open(opts)
         status.reset()
         status.refresh(true)
       else
+        status.reset()
         status.create(opts.kind, opts.cwd)
       end
     end)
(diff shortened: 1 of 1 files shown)


ten3roberts (approved):
Amazing, simple and easy solution.

The dreaded enemy of all programmers: *cache invalidation*

ten3roberts:
As for the tests, it is likely they fail as they assume neogit opens and has the correct state immediately, but now such happens in the next event loop tick.

There may be an easy "wait for everything to open" for those tests

smatting:
> The dreaded enemy of all programmers: cache invalidation

Maybe the cleanest way would be not to cache this at all. Any objections against trying this?

CKolkey:
I'm up to try, sure :)

smatting:
Closing this in favor of https://github.com/NeogitOrg/neogit/pull/944

PR #934, opened 2023-11-13
Graph colors

gollth:
Hi there,

first of all, thanks again for this amazing package! Long time emacs user decided to switch to nvim just because of you ^^

I was trying to configure the look and feel of the log graph view of `neogit` to match more the original color scheme of magit:

magit-graph

However with my config the colors of neogit look rather dull:

neogit-graph

I tried to configure the `NeogitGraph*` highlight groups without success. After digging a bit through the code of neogit, I noticed that you use `Macro` and `String` highlight groups for local/remote branch names. Maybe these could reuse the [NeogitBranch & NeogitRemote](https://github.com/NeogitOrg/neogit/blob/master/doc/neogit.txt#L201-L202) highlight groups instead?

I don't have much experience with Lua or highlight groups, so please let me know if there is another (better?) way to configure this =)

### DIFF
--- a/doc/neogit.txt
+++ b/doc/neogit.txt
@@ -265,6 +265,7 @@ NeogitFilePath              Applied to filepath
 NeogitCommitViewHeader      Applied to header of Commit View
 
 LOG VIEW BUFFER
+NeogitGraphAuthor           Applied to the commit's author in graph view
 NeogitGraphBlack            Used when --colors is enabled for graph
 NeogitGraphBlackBold
 NeogitGraphRed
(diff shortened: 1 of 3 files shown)


CKolkey:
Hey! I appreciate the kind words. No objection to the changes here, but check out `lua/neogit/lib/hl.lua` for the definition of the highlight groups. If you're going to add a group, it should get a definition/link there. :)

gollth:
Sure thing, added the `NeogitGraphAuthor` group there

CKolkey:
Good call reusing the highlight groups. Thanks!

Issue #442, opened 2023-02-03
Can't stage a hunk from file up a dir from PWD


mangelozzi:
### Description

I have a project structure that looks like this (these dirs all exist in the git project):
```
config/
info/
install/
scripts/
src/
   |---dir1/
   |---dir2/
```

I generally have the `PWD`  of VIM set to `src/` for python linting and thats the dir the project starts up in.
When I review a file say in the `config/` dir, and try to stage hunk it with Neogit, it shows an error message (image below). One can stage the whole file succesfully though, just not the hunk

### Neovim version

NVIM v0.8.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by runner@fv-az190-151

### Operating system and version

Ubuntu 22.04

### Steps to reproduce

1. Open a git project
2. make a change in a file higher up in the project structure
3. CD into a lower dir
4. Try stage the file in the higher up dir.

### Expected behavior

The hunk to be staged

### Actual behavior

The following error message:
![image](https://user-images.githubusercontent.com/19764314/216595228-1b41cbbf-caa2-4707-a1c2-19a95193ef46.png)

The file is actually `../README.md`, and not `src/../README.md`. Which it gets right if one stages the whole file.

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "TimUntersberger/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing neogit and dependencies.")
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
```

dusty-phillips:
This is happening for me in a monorepo setup as well.

PR #944, opened 2023-11-16
Update git_root on refresh

smatting:
Fixes https://github.com/NeogitOrg/neogit/issues/931 and fixes https://github.com/NeogitOrg/neogit/issues/442

This PR updates `in_prepared_repo` to a different scenario (called "Scenario 2"), namely that the user changes directory to a git repo after `setup()` has already been called. This makes 14 tests fail, also including the tests for hunk operations which explains the issue https://github.com/NeogitOrg/neogit/issues/931.

The second commit fixes all failing tests, by making sure `repository.git_root` and `repository.git_path` is updated on every refresh of the status. Also this commit removes the unused `repository.cwd`.

Maybe this PR could be extended so that `in_prepared_repo` would run each test for both scenarios:
1) Scenario 1: User runs setup() while already being in the right git repo (i.e. you start `nvim` inside the git repo).
2) Scenario 2: User changes directory after setup() has been called (i.e. you start `nvim` and then `:cd` into the git repo)

I was not able to fit both scenarios into `in_prepared_repo` because running Scenario 1 first would influence the outcome of running Scenario 2, so some critical cleanup was missing between running both scenarios. Maybe somebody can help me with this? With the bugfix included in this PR all tests pass when running them in either Scenario 1 and Scenario 2 separately. We could of course also duplicate all tests, but that would be ugly. What we really want is that that every test passes in either scenario, because that assert that neogit behaves the same.

### DIFF
--- a/lua/neogit/lib/git/repository.lua
+++ b/lua/neogit/lib/git/repository.lua
@@ -3,7 +3,7 @@ local logger = require("neogit.logger")
 
 -- git-status outputs files relative to the cwd.
 --
--- Save the working directory to allow resolution to absolute paths since the
+-- Save the git_root to allow resolution to absolute paths since the
 -- cwd may change after the status is refreshed and used, especially if using
 -- rooter plugins with lsp integration
 -- stylua: ignore start
@@ -15,7 +15,6 @@ local function empty_state()
     git_path     = function(...)
       return Path.new(root):joinpath(".git", ...)
     end,
-    cwd          = vim.fn.getcwd(),
     git_root     = root,
     head         = {
       branch = nil,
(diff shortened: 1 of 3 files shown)


smatting:
@CKolkey  I'm no longer sure this PR makes sense, now that I have read more of the codebase: it appears like everything related to cwds is much more complicated that it has to be. Wouldn't it much simpler if all git commands would be executed with `cwd = git.repo.git_root` and this state would only be updated when you call `:Neogit` (and only then)?

CKolkey:
Hmm. Yeah, sounds reasonable to me. A lot of the CWD logic is from before I got involved with this project. If you think there's a better (simpler) way, I'd love to see what you have in mind :)

smatting:
@CKolkey I manged to pull it off: https://github.com/NeogitOrg/neogit/pull/963

PR #946, opened 2023-11-16
Show HEAD as `NeogitPopupBranchName` in Log Graph

gollth:
Follow-up after #934: In the log graph it's a bit non-intuitive where the currently checked out branch is. To make this more customizable for the user, we could reuse the `NeogitPopupBranchName` highlight group.

### DIFF
--- a/doc/neogit.txt
+++ b/doc/neogit.txt
@@ -199,9 +199,9 @@ these yourself before the plugin loads, that will be respected. If they do not
 exist, they will be created with sensible defaults based on your colorscheme.
 
 STATUS BUFFER
-NeogitBranch                Header showing currently checked out branch
-NeogitBranchHead            Currently checkoued branch (in log graph)
-NeogitRemote                Header showing current branch's remote (if set)
+NeogitBranch                Local branches
+NeogitBranchHead            Accent highlight for current HEAD in LogBuffer
+NeogitRemote                Remote branches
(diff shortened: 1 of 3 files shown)


gollth:
Refactored like you suggested. One question, though. In the `neogit.txt` for the explanation of the highlight groups it is mentioned that:

```txt
NeogitBranch                Header showing currently checked out branch
NeogitRemote                Header showing current branch's remote (if set)
```

which is not entirely true, right? E.g in the graph it will be used not only for the currently checked out branch, but for all branches. Should we reword this section a bit. say sth. like:

```txt
NeogitBranch                Local branches (e.g. currently checked out branch in status)
NeogitBranchHead            Currently checked out branch (in log graph)
NeogitRemote                Remote branches (e.g. current branch's remote if set) 
```

CKolkey:
Yeah, agreed, there's some inconsistency there. I think we could just say:

```
NeogitBranch                Local branches
NeogitBranchHead            Accent highlight for current HEAD in LogBuffer
NeogitRemote                Remote branches
```

In _general_ my goal is to have the same thing (ref) look the same, regardless of where it is. I'll have a spin with this tonight, possibly.

gollth:
> In general my goal is to have the same thing (ref) look the same, regardless of where it is.

This is also what I was trying to achieve with reusing highlight groups as much as possible. Makes total sense. Reformulated the txt as you suggested

PR #963, opened 2023-11-22
Remove dependency on cwd

### LINKED ISSUE #442: Can't stage a hunk from file up a dir from PWD

### LINKED ISSUE #931: Changing CWD breaks action on hunks

smatting:
This pull request fixes #931 and fixes #442 by ensuring the independence of nearly all library functions from the current working directory (cwd).

Specifically, this pertains to functions that directly or indirectly use cwd, which can change anytime, either manually by the user or automatically through plugins. Cwd dependent functions make it difficult to reason about the state of Neogit and potentially lead to numerous bugs. Throughout the codebase, you can observe various attempts to circumvent this issue, often involving the tracking of the cwd via `state.cwd`.

Notably, all functions in `neogit.lib.git.cli` are dependent on the cwd, with `require("neogit.lib.git.cli").git_root()` being a noteworthy example.

This pull request aims to eliminate all cwd dependencies by designating `require("neogit.lib.git.repository").git_root` as the source of truth for the git root that Neogit is targeting. Unlike the cwd, which can change at any point, `repository.git_root` only updates through `repository.refresh()` triggered by manual user intervention (e.g., `Ctrl+R` or `:Neogit`) and should - hopefully - also invalidate all other state which depends on the git root.

Key changes in this pull request include:
- Ensuring that all `cli` functions are executed with `cwd = repository.git_root`. This change is particularly impactful as it ensures that all cwd-relative file paths returned by `git` do not need to be converted to paths relative to the git root, as `cwd == git_root` always holds.
- Replacing all instances of `cli.git_root()` with `repository.git_root` in the codebase to maintain consistency.
- Renaming `cli.git_root()` to `cli.git_root_of_cwd()` to discourage its use and prevent undoing the modifications introduced by this pull request.
- Removing workaround code related to changing cwd, such as `diffview.root_prefix()`, `repository.cwd`, and the `use_git_root` argument of hunk operations.

Smaller changes include:
- Renaming `git_is_repository_sync` to `is_inside_worktree` for clarity.
- Removing `neogit.get_repo()` due to its solitary use.
- Eliminating unused functions `cli.git_root_sync()` and `cli.git_dir_path_sync()`.
- Making `repository.git_path` a function of the module rather than part of the state, as `git_path` depends on `git_root`.

Additionally, two tests that relied on Neogit being cwd dependent were adjusted to ensure that `status.reset()` is called after the cwd has been changed.

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -67,7 +67,7 @@ function M.open(opts)
   end
 
   if not opts.cwd then
-    opts.cwd = require("neogit.lib.git.cli").git_root()
+    opts.cwd = require("neogit.lib.git.cli").git_root_of_cwd()
   end
 
   if not did_setup then
@@ -76,7 +76,7 @@ function M.open(opts)
     return
   end
 
-  if not cli.git_is_repository_sync(opts.cwd) then
+  if not cli.is_inside_worktree(opts.cwd) then
     if
       input.get_confirmation(
         string.format("Initialize repository in %s?", opts.cwd or vim.fn.getcwd()),
@@ -168,10 +168,6 @@ function M.complete(arglead)
   return vim.tbl_filter(function(arg)
     return arg:match("^" .. arglead)
   end, { "kind=", "cwd=", "commit" })
-end
-
-function M.get_repo()
-  return require("neogit.lib.git").repo
 end
 
 function M.get_log_file_path()
--- a/lua/neogit/integrations/diffview.lua
+++ b/lua/neogit/integrations/diffview.lua
@@ -9,6 +9,7 @@ local dv_lib = require("diffview.lib")
 local dv_utils = require("diffview.utils")
 
 local neogit = require("neogit")
+local repo = require("neogit.lib.git.repository")
 local status = require("neogit.status")
 local a = require("plenary.async")
 
@@ -35,47 +36,12 @@ local function cb(name)
   return string.format(":lua require('neogit.integrations.diffview').diffview_mappings['%s']()<CR>", name)
 end
 
----Resolves a cwd local file to git root relative
-local function root_prefix(git_root, cwd, path)
-  local t = {}
-  for part in string.gmatch(cwd .. "/" .. path, "[^/\\]+") do
-    if part == ".." then
-      if #t > 0 and t[#t] ~= ".." then
-        table.remove(t, #t)
-      else
-        table.insert(t, "..")
-      end
-    else
-      table.insert(t, part)
-    end
-  end
-
-  local git_root_parts = {}
-  for part in git_root:gmatch("[^/\\]+") do
-    table.insert(git_root_parts, part)
-  end
-
-  local s = {}
-  local skipping = true
-  for i = 1, #t do
-    if not skipping or git_root_parts[i] ~= t[i] then
-      table.insert(s, t[i])
-      skipping = false
-    end
-  end
-
-  path = table.concat(s, "/")
-  return path
-end
-
 local function get_local_diff_view(selected_file_name)
   local left = Rev(RevType.STAGE)
   local right = Rev(RevType.LOCAL)
-  local git_root = neogit.cli.git_root_sync()
 
   local function update_files()
     local files = {}
-    local repo = neogit.get_repo()
     local sections = {
       conflicting = {
         items = vim.tbl_filter(function(o)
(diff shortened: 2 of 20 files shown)


CKolkey:
Very insightful and thorough. I _do_ love a PR that removes nearly 2x the lines it adds.

Thank you for the contribution :)

PR #967, opened 2023-11-23
Hide internal CLI calls from git history

CKolkey:
Remove unused code for "parallel builder"
Extract some Constants
Allow internal state commands to be hidden from Git History buffer


smatting:
Nice! The git history buffer was so noisy it was very hard to find anything

CKolkey:
> Nice! The git history buffer was so noisy it was very hard to find anything

Yeah, it's a bit much right now. I've been meaning to do this for ages haha

PR #970, opened 2023-11-25
Allow to open pull popup from within log view

gollth:
Another small thing, that worked in Magit: When pressing `p` in the log graph view you could open the "pull" popup. On my setup this fails with:

image

Added this open to the pop up mappings in `log_view`:
pull-popup gif


CKolkey:
Ah, ok! The implementation is actually even more simple than you have it - we don't do anything with the passed in commit, so you can omit that second function as an argument entirely. Since it's so simple I just went ahead and did it: https://github.com/NeogitOrg/neogit/pull/972

Thanks for pointing that out :)

gollth:
Ah okay, good to know. Thanks for fixing!

Issue #955, opened 2023-11-18
Bad scrolling performance in git log buffer


mortezadadgar:
### Description

Hey,
It's the first time I'm using this plugin and looks pretty promising, although noticed that scrolling in a new file containing 1K lines of change is so slow

	

https://github.com/NeogitOrg/neogit/assets/27911493/bc4ebe53-e138-469b-aa1c-52dfcc053faa

### Neovim version

NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1697887905

### Operating system and version

Gentoo Linux

### Steps to reproduce

1. create a file like https://0x0.st/HvUQ.txt
2. init git repo and commit
3. open nvim with minmal config and `:Neogit`
4. open git log in neogit
5. scroll in buffer and watch neovim struggles to scroll smoothly

### Expected behavior

scroll smoothly

### Actual behavior

scrolling in git log buffer is choppy and slow

### Minimal config

```Lua
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
	vim.fn.system({
		"git",
		"clone",
		"--filter=blob:none",
		"https://github.com/folke/lazy.nvim.git",
		"--branch=stable",
		lazypath,
	})
end
vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
	{ "NeogitOrg/neogit", config = true },
	"nvim-lua/plenary.nvim",
})
```

CKolkey:
Yeah. There's some gnarly `o(n)` logic [here](https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/lib/ui/init.lua#L133-L138). I've been meaning to do something about it, but haven't found the time yet.

CKolkey:
Well, it just so happens that nvim 0.10 added an API that will let us do native folding instead of the custom stuff we do now! https://github.com/NeogitOrg/neogit/pull/959

So, this won't look nice for the current stable release, but it's a HUGE performance win.

mortezadadgar:
Thanks for fixing it can't wait for it to be merged!

CKolkey:
Same, this has been bugging me for ages.

mortezadadgar:
@CKolkey the slowness is not fixed I believe it has gone even worse you can reproduce it with the file that i have attached in issue

CKolkey:
[...]

Issue #951, opened 2023-11-17
Show ongoing merge or rebase in overview


agraven:
**Is your feature request related to a problem? Please describe.**
An active merge with conflicts or rebase that's stopped can significantly change how the unstaged/staged status of files should be interpreted, and changes what the desirable next course of action to perform is.

Displaying if there's an active rebase or merge in the overview would make this immediately clear, and would remove a lot of ambiguity of the current state of the repo. I am often uncertain about whether a merge or a rebase actually succeeded when I currently look at the overview.

**Describe the solution you'd like**
An extra field next to `Head:` and `Merge:` at the top that appears if there's an ongoing merge or rebase.

**Describe alternatives you've considered**
Displaying the information elsewhere in the UI

**Additional context**
This information is present in a regular `git status` invocation. fugitive also displays ongoing rebases, though not ongoing merges.

CKolkey:
Oh, there should be a section in the status buffer that shows up during a rebase:
<img width="478" alt="Screenshot 2023-11-17 at 13 02 23" src="https://github.com/NeogitOrg/neogit/assets/7228095/3e154c7e-1751-4509-a6e3-8a6f0efa32b7">


If the rebase is over without conflict, it's probably too fast to appear, but you'll see it with a conflict.

agraven:
I don't think that appeared for me. I'll see if I can reproduce it. Either way, displaying an ongoing merge would still be a nice addition.

Issue #909, opened 2023-10-29
Correctly focus the last tab when closing Neogit* buffers


asmodeus812:
**Is your feature request related to a problem? Please describe.**
Would be great if the kind "tab" worked correctly with closing buffers, or at least provide a way to customize the generic Buffer mappings. The feature request is that when closing a tabbed neogit buffer and there is no alternate buffer on the  current tab, of any kind the last focused tab should be opened i.e g tab functionality. The current behaviour would just focus the next tab instead.

**Describe the solution you'd like**
Either a user configuration, or implemented in neogit

**Describe alternatives you've considered**
Overriding the Buffer:close but the buffer file has this at the top `package.loaded["neogit.buffer"] = nil` which would prevent us from overriding in a robust way.

**Additional context**
None

gegoune:
I use `<Cmd>tabnext # <Bar> tabclose #<CR>` to close Diffview's tab, something similar can be done for Neogit too.

Issue #714, opened 2023-08-07
Highlights don't update on cursor move


ghost:
### Description

The highlighting doesn't update for me on cursor move. Only updates when folds are expanded or closed.

### Neovim version

NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3

### Operating system and version

Debian 6.1.38-2

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. :Neogit

### Expected behavior

Hunks under the cursor to become highlighted. 

### Actual behavior

The highlights only update when folds are expanded or closed.

### Minimal config

```Lua
vim.opt.termguicolors = true
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "NeogitOrg/Nneogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing neogit and dependencies.")
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
```

PR #978, opened 2023-11-28
[Neovim Nightly] 0.10

### LINKED ISSUE #1250: Making a new branch should refresh the status buffer
### Description

This is similar to #1240. After creating a new branch, the status buffer doesn't update.

### Neovim version

NVIM v0.10.0-dev-2818+gae28ef327
Build type: RelWithDebInfo
LuaJIT 2.1.1710088188

Linux 6.7.9-amd64

### Steps to reproduce

1. Go to any git directory
1. `nvim -nu minimal.lua`
1. `:Neogit`
1. Make a new branch with `b c` from any existing branch

### Expected behavior

The Ui should update to show the new branch details.

### Actual behavior

The previous state is still shown.

### Minimal config

```Lua
local M = {}

function M.root(path)
end

  end

    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
    if vim.v.shell_error > 0 then
        vim.log.levels.ERROR)
    end
  end
end

function M.setup(plugins)

  -- Install required plugins
  if plugins ~= nil then
    end
  end

    callback = function()
      vim.fn.system({
        "rm",
        "-r",
        "-f",
        M.root("xdg")
      })
    end
  })
end

M.setup({
})
```

CKolkey:
Todo: Set up CI action to auto merge master into Nightly, presuming no conflicts.

Changes that will land in Master when Neovim 0.10 is released.
- Rewritten folding to use native folds instead of internal state.
- Added RefsView buffer. 
- Use new extmarks api for line highlighting/signs
- Use new options api
- Rewritten status buffer using UI framework
- Improved Rebase section for status buffer
- Added rebase_editor and commit_editor insert mode specific mappigns
- Added ability to remap `OpenOrScrollUp` and `OpenOrScrollDown`
- Removed remaining `ftpugin/*.vim` files
- Removed `require("neogit").refresh_viml_compat()`
- ? Removed `require("neogit").refresh_manually()`
- Completely rewritten status buffer
- Remove old bootstrap.lua file
- Added a header to the commit-select view, with instructions
- Added status column override
- Improved how "is the buffer open" is checked internally
- Better support for showing the correct SHA hash length in UI
- Fix height of fuzzy finder always being `0.3` instead of a computed value
- Use mapped key to close git command history
- Use native folds in git command history
- Added `NEOGIT_DEBUG` env var, which will show, among other things, git CLI calls that are otherwise hidden from the user.
- Use native folds for log view
- Added `OpenOrScrollDown` and `OpenOrScrollUp` to status buffer, reflog buffer, and refs buffer
- Extend the "Close" mapping to close the topmost "layer", even if unfocused.
- Adds a diff view to the commit editor, so you can see the changes you're about to commit.
- Adds `Untrack` (bound to `K`)

### Refs View:
- [ ] Finalize Keybindings
### Docs:
- [ ] Bisect Popup Documentation

### Status:
- [ ] Status buffer loses context highlighting when telescope picker is activated (`b l`). Commit view doesn't.
- [ ] Status buffer might not render a diff with highlights when a notification (popup window) is present. Probably related to above.

Closes #1250, #1240, #1160, #1192, #1170, #1171, #1156, #1155, #1140, #1130, #1109, #1096, #955, #951, #909, #515, #714, #509, #122, #334, #58, #335, #1248, #1291


CKolkey:
Yeah, there are some API's that dont exist on 0.9 that I'm using. The only one so far is for fold-text to use the extmarks api for nice highlighting, but there are (loose) plans to use the new `vim.system` api to replace the async jobs. I'm not going to go out of my way to break stuff, but... i expect it to happen.

What I can do is tag the commit before I merge this for 0.9 compatibility, so people can pin to that if they need.

CKolkey:
Thanks @ofseed - fixed

ofseed:
Thank you for your fix. I found that after the fix, diffview works well but the diff buffers have a `'statuscolumn'` set by neogit. Maybe it is not needed?

CKolkey:
> Is this expected to break on older versions of neovim when merged?

You can pin to: https://github.com/NeogitOrg/neogit/releases/tag/v0.0.1

PR #980, opened 2023-11-29
Merge + Help Popups in {Commit, Log, Reflog} Views

gollth:
Found another couple of missing keybinds =)

### DIFF
--- a/lua/neogit/buffers/commit_view/init.lua
+++ b/lua/neogit/buffers/commit_view/init.lua
@@ -154,6 +154,7 @@ function M:open()
           p { commit = self.commit_info.oid }
         end),
         [popups.mapping_for("FetchPopup")] = popups.open("fetch"),
+        [popups.mapping_for("HelpPopup")] = popups.open("help"),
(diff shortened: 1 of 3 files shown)


CKolkey:
Merge popup is an easy ✅ 

Help popup I think is a little weird. It's pretty tightly coupled to the status buffer (the stage/unstage commands are all baked in) and it doesn't have the ability to get the current commit under the cursor so will behave differently when opening a popup from the help menu vs directly.

My plan for that is to do some/all of the following:
- Make the help popup more composable so callers can modify what gets shown
- Pass `commits` into help popup constructor so they can be passed into any popups that are opened up from the help menu
- Create different help popups for each view

gollth:
Ah okay, didn't look that closely to the help popup (was just wondering why trying to open it brought up a vim search :D)

I will revert the keybind commit for now, than until the preps for the customizable help are ready. All points sound very resonable to me

CKolkey:
Cool! Thanks as always :)

gollth:
Thank you for this plugin =)

Issue #938, opened 2023-11-14
Log graph does not update on push, pull, reset...


gollth:
### Description

When I do some git action like pushing, fetching, rebasing or resetting, from the log graph view, the changes in the history are not reflected immediately. I have to `q` out of the log graph view and then `l b` back into the view to see the changes. `<C-r>` does not help either.

Ideally the log graph automatically updates, similar to how magit is doing it:

| Neogit | Magit |
|:---|:---|
|![neogit-push](https://github.com/NeogitOrg/neogit/assets/31999281/6cd33dfc-428f-4478-a705-83a5bcf0a4e6)|![magit-push](https://github.com/NeogitOrg/neogit/assets/31999281/2ef6ecda-6656-4f85-9733-08a1ac302326)|

### Neovim version

NVIM v0.10.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Run "nvim -V1 -v" for more info

### Operating system and version

Ubuntu 20.04

### Steps to reproduce

1. Open Neogit 
2. Do any git action, which changes the log graqh, e.g. pushing or resetting
3. Note how the log graph is not updating

### Expected behavior

Similar behaviour to Magit, i.e. that the log graph automatically reflects changes in the history (or at least an easy way to refresh the log graph view, e.g. via autocmds...)

### Actual behavior

Log graph is not updating after a git action is performed

### Minimal config

```Lua
-- NOTE: See the end of this file if you are reporting an issue, etc. Ignore all the "scary" functions up top, those are
-- used for setup and other operations.
local M = {}

local base_root_path = vim.fn.fnamemodify(debug.getinfo(1, "S").source:sub(2), ":p:h") .. "/.min"
function M.root(path)
  return base_root_path .. "/" .. (path or "")
end

function M.load_plugin(plugin_name, plugin_url)
  local package_root = M.root("plugins/")
  local install_destination = package_root .. plugin_name
  vim.opt.runtimepath:append(install_destination)

  if not vim.loop.fs_stat(package_root) then
    vim.fn.mkdir(package_root, "p")
  end

  if not vim.loop.fs_stat(install_destination) then
    print(string.format("> Downloading plugin '%s' to '%s'", plugin_name, install_destination))
    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
    if vim.v.shell_error > 0 then
[...]

PR #981, opened 2023-11-29
More `Neogit*` Events

gollth:
This provides a working solution for my problem in #938. Let me explain:

With the current four events the user (me^^) can set up a vim autocmd which closes and reopens the Graph View. Not pretty, but does the job for me. However it's a bit cumbersome to being only limited to update the view on commit, push, pull & fetch. With these additional events the log view also does update on other commands:

more-events

This is my user config for auto-updating:

```lua
local group = vim.api.nvim_create_augroup("MyCustomNeogitEvents", { clear = true })
vim.api.nvim_create_autocmd("User", {
  pattern = {
    "NeogitCherryPick",
    "NeogitBranchCheckout",
    "NeogitBranchCreated",
    "NeogitBranchDelete",
    "NeogitBranchReset",
    "NeogitBranchRename",
    "NeogitRebase",
    "NeogitReset",
    "NeogitTagCreate",
    "NeogitTagDelete",
    "NeogitCommitComplete",
    "NeogitPushComplete",
    "NeogitPullComplete",
    "NeogitFetchComplete",
  },
  group = group,
  callback = function(event)
    print("Event: " .. vim.inspect(event.file))
    local buffername = vim.api.nvim_buf_get_name(event.buf)
    if buffername:match("NeogitLogView$") then
      vim.fn.feedkeys("q", "x") -- Close Log Graph
      require("neogit").open({ "log" })
      vim.fn.feedkeys("b", "x") -- Open log graph (branches) again    end
    end
  end,
})
```

If you want this example config could become part of some "Recipes" documentation or so...

---

FWIW, maybe other people might benefit from these events, that's why this PR =)

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,6 +399,8 @@ Neogit emits the following events:
 | `NeogitBranchRename`    | Branch was renamed                       | `{ branch_name: string, new_name: string }`     |
 | `NeogitRebase`        | A rebase finished                        | `{ commit: string, status: "ok"\|"conflict" }`    |
 | `NeogitReset`         | A branch was reset to a certain commit   | `{ commit: string, mode: "soft"\|"mixed"\|"hard"\|"keep"\|"index" }` |
+| `NeogitTagCreate`     | A tag was placed on a certain commit     | `{ name: string, ref: string }`                   |
+| `NeogitTagDelete`     | A tag was removed                        | `{ name: string }`                                |
 
 You can listen to the events using the following code:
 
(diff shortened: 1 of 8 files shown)


CKolkey:
Cool! I'm planning to handle the buffer update thing with the watcher module, but these could be handy for people regardless, so I don't see the harm. Thanks!

CKolkey:
(I really like the gif's, btw. Great way to show whats up)

PR #1030, opened 2023-12-14
Remote Popups in {Commit, Log, Reflog} Views

gollth:
...and [another one](https://github.com/NeogitOrg/neogit/pull/980) :)


CKolkey:
_Nice._

PR #1046, opened 2023-12-18
Fixed commit SHA check preventing line actions in rebase editor

Aphosis:
Hey there !

The rebase editor actions stopped working recently and I was about to open an issue but the fix was really straightforward so I figured it would be even quicker to open a pull request.

In the rebase editor, there is a check trying to match a commit SHA of length 10 instead of 8. Looking at other parts of the code base it doesn't look intentional.

This was preventing most rebase editor actions from modifying the current line, at least on my repository.

Thanks for the awesome plugin by the way, I started using it recently and that's a really nice workflow !

### DIFF
--- a/lua/neogit/buffers/rebase_editor/init.lua
+++ b/lua/neogit/buffers/rebase_editor/init.lua
@@ -21,7 +21,7 @@ local function line_action(action)
       return
     end
 
-    if line[2] and line[2]:match("%x%x%x%x%x%x%x%x%x%x") and line[1] ~= "Rebase" then
+    if line[2] and line[2]:match("%x%x%x%x%x%x%x%x") and line[1] ~= "Rebase" then
       line[1] = action
       vim.api.nvim_set_current_line(table.concat(line, " "))
       buffer:write()


CKolkey:
Hey! Thanks for the patch. Definitely just an oversight :)

CKolkey:
Huh, just checked again, and when I open the rebase buffer it _is_ a 10-character hash. Wonder why yours might be 8. No big deal - if 8 is valid 10 is valid.

Aphosis:
It may be because some repositories need more than 8 characters to avoid clashes in an abbreviated commit hash ?

What's weird is that I've tried on a repo at work with more than 5000 commits, so I would expect to have a similar commit hash length of 10.

The documentation on this topic is a bit lacking, but it looks like if no formatting option is given to `git-rebase` it will use a format similar to `git-log` `oneline` format, with a commit hash abbreviated. The minimum number of characters git will trim the commit hash to abbreviate it is hardcoded to 7.

- https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt-rebaseinstructionFormat
- https://stackoverflow.com/a/73250185
- https://github.com/git/git/blob/624eb90fa8f65a79396615f3c2842ac5a3743350/object-name.h#L128

I see that someone else had the issue on a small repo so with !1049 you should definitely be fine.

CKolkey:
Collision avoidance makes perfect sense. I feel like I should have noticed, because there's plenty of `item.oid:sub(1,7)` in the view layer here, and the fact that mine was 10 _did_ stand out. Oh well. No harm done.

For reference, the work repo I tested it on has about 50,000 commits, so maybe it's by order-of-magnitude?

Aphosis:
It looks like it is indeed !

https://github.com/git/git/blob/624eb90fa8f65a79396615f3c2842ac5a3743350/object-name.c#L795

PR #1066, opened 2023-12-24
Add worktree popup

CKolkey:
Worktrees!


lua/neogit/popups/init.lua
```diff
@@ -153,6 +153,9 @@ function M.mappings_table()
       },
     },
     { "RemotePopup", "Remote", M.open("remote") },
+    { "WorktreePopup", "Worktree", M.open("worktree", function(f)
+      f {}
+    end) },
```
CKolkey, line 158: Do something with env or don't pass it

PR #1075, opened 2023-12-28
Use the Selene binary instead of building from source

azzamsa:
The `taiki-e/install-action` action pulls the target binary and resorts to building from source if not found. However, Selene is compatible with `cargo-binsall`, eliminating the need for source building.

The `Swatinem/rust-cache` action inherently caches `~/.cargo` directories, including the location where Selene is stored.


CKolkey:
Nice, thank you :)

PR #1077, opened 2023-12-28
Add `typos` linter

azzamsa:
We have identified numerous typos in the past. Hopefully, incorporating `typos` into our linter tools will assist us in minimizing such errors.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,9 +16,9 @@ jobs:
 
       - uses: taiki-e/install-action@v2
         with:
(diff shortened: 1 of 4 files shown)


azzamsa:
@CKolkey, what about the GitHub CI job name? I don't think naming it `selene` is suitable anymore. However, changing it to `linters` doesn't make sense either because `stylua` is also a linter.

CKolkey:
Good point. We could combine stylua/selene/typos into one "lint" action, since I think "test" is still going to be slower anyhow.

PR #1095, opened 2024-01-02
Handle core.commentChar in commit and rebase editors

lechat:
add config variable to hide commit help - disable_editor_help

I have the same problem as #320, I have to change `core.commitChar` to ";" in git. This PR fixes problem in commit and rebase editors.

If it seems that adding common function to `neogit.lib.git` is not appropriate, happy to change that to some other place.

### DIFF
--- a/lua/neogit/buffers/editor/init.lua
+++ b/lua/neogit/buffers/editor/init.lua
@@ -77,30 +77,30 @@ function M:open(kind)
         return pad(mapping[name] and mapping[name][1] or "<NOP>", padding)
       end
 
-      if not config.values.disable_editor_help then
-        local comment_char = git.config.get("core.commentChar"):read() or git.config.get_global("core.commentChar"):read() or "#"
-        -- stylua: ignore
-        local help_lines = {
-          string.format("%s", comment_char),
-          string.format("%s Commands:", comment_char),
-          string.format("%s   %s Close", comment_char, pad_mapping("Close")),
-          string.format("%s   %s Submit", comment_char, pad_mapping("Submit")),
-          string.format("%s   %s Abort", comment_char, pad_mapping("Abort")),
-          string.format("%s   %s Previous Message", comment_char, pad_mapping("PrevMessage")),
-          string.format("%s   %s Next Message", comment_char, pad_mapping("NextMessage")),
-          string.format("%s   %s Reset Message", comment_char, pad_mapping("ResetMessage")),
-        }
+      local comment_char = git.config.get("core.commentChar"):read()
+        or git.config.get_global("core.commentChar"):read()
+        or "#"
 
-        help_lines = util.filter_map(help_lines, function(line)
-          if not line:match("<NOP>") then -- mapping will be <NOP> if user unbinds key
-            return line
-          end
-        end)
+      -- stylua: ignore
+      local help_lines = {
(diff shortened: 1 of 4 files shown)


lua/neogit/buffers/editor/init.lua
```diff
@@ -77,26 +77,27 @@ function M:open(kind)
...
-      end)
+      if not config.values.disable_editor_help then
+        local comment_char = git.config.get_comment_char()
+        -- stylua: ignore
+        local help_lines = {
+          string.format("%s", comment_char),
+          string.format("%s Commands:", comment_char),
+          string.format("%s   %s Close", comment_char, pad_mapping("Close")),
+          string.format("%s   %s Submit", comment_char, pad_mapping("Submit")),
+          string.format("%s   %s Abort", comment_char, pad_mapping("Abort")),
```
CKolkey, line 88: I think the `Previous`, `Next`, and `Reset` commands got dropped on accident.

lua/neogit/lib/git/config.lua
```diff
@@ -147,4 +147,16 @@ function M.unset(key)
...
+  local char = M.get("core.commentChar").value
+  if char == "" then
+    char = M.get_global("core.commentChar").value
+    if char == "" then
+      char = "#"
+    end
+  end
+
+  return char
+end
```
CKolkey, line 160: I don't really wanna have bespoke value getters in this library, but we can simplify how it's written in the callers to obviate the need for a function:

```lua
local comment_char = git.config.get("core.commentChar"):read() or git.config.get_global("core.commentChar"):read() or "#"
```

lua/neogit/config.lua
```diff
@@ -135,6 +135,7 @@ end
 ---@field fetch_after_checkout? boolean Perform a fetch if the newly checked out branch has an upstream or pushRemote set
 ---@field telescope_sorter? function The sorter telescope will use
 ---@field disable_insert_on_commit? boolean|"auto" Disable automatically entering insert mode in commit dialogues
+---@field disable_editor_help? boolean|"auto" Disable displaying commit/rebase editor help
```
CKolkey, line 138: I don't think I want to expose this - at least not now.

CKolkey (changes requested):
Ok! We can do something about this. Left some notes

lua/neogit/buffers/editor/init.lua
lechat: Not dropped by me. I just changed what was there.

lua/neogit/lib/git/config.lua
lechat: That's exactly what I was asking about. Will change code in a moment.

lua/neogit/config.lua
lechat: Will remove this line.

lua/neogit/buffers/editor/init.lua
lechat: Ah, there were more changes in editor. Will  fix that as well.

CKolkey: All good, I merged it very recently, probably some time between you forking and opening a PR :)

CKolkey:
Thanks a ton for the contribution :)

PR #1104, opened 2024-01-04
Use for-the-badge style for Dotfyle badge

codicocodes:
Thanks for adding the Dotfyle badge to the repo! Noticed that the dotfyle badge had a different style, so changed it. This styling aligns with the other badges in the Readme.


CKolkey:
Hey, thanks!

PR #1118, opened 2024-01-10
Rebase to Reword

### LINKED ISSUE #1114: Tracking issue for _Rebase to reword a commit_
I (would like to) use the `r` > `w` (Rebase to reword a commit) feature quite regularly, so here is a tracking issue for you to get insight if other people are in need of it. It's probably not the highest priority on your list, but from the surface it also doesn't seem to hard. If you give me a pointer where to start, I can also have a look at implementing this

neogit-reword-a-commit

gollth:
Implementation based on https://peterevans.dev/posts/how-to-rewrite-git-commit-messages-non-interactively/

Closes #1114

### DIFF
--- a/lua/neogit/lib/git/rebase.lua
+++ b/lua/neogit/lib/git/rebase.lua
@@ -75,6 +75,10 @@ function M.onto(start, newbase, args)
   end
 end
 
+---@param commit string rev name of the commit to reword
+---@param message string new message to put onto `commit`
+---@param arguments table additional arguments from the popup (unused)
+---@return nil
 function M.reword(commit, message)
   local result = cli.commit.allow_empty.only.message("amend! " .. commit .. "\n\n" .. message).call()
   if result.code ~= 0 then
--- a/lua/neogit/lib/git/rebase.lua
+++ b/lua/neogit/lib/git/rebase.lua
@@ -77,7 +77,6 @@ end
 
 ---@param commit string rev name of the commit to reword
 ---@param message string new message to put onto `commit`
----@param arguments table additional arguments from the popup (unused)
 ---@return nil
 function M.reword(commit, message)
   local result = cli.commit.allow_empty.only.message("amend! " .. commit .. "\n\n" .. message).call()
--- a/lua/neogit/lib/git/rebase.lua
+++ b/lua/neogit/lib/git/rebase.lua
@@ -75,6 +75,19 @@ function M.onto(start, newbase, args)
   end
 end
 
+function M.reword(commit, message, _args)
+  local result = cli.commit.allow_empty.only.message("amend! " .. commit .. "\n\n" .. message).call_sync()
+
+  if result.code ~= 0 then
+    return
+  end
+  result = cli.rebase.env({ GIT_SEQUENCE_EDITOR = ":" }).interactive.autosquash.commit(commit).call_sync()
+  if result.code ~= 0 then
+    return
+  end
+  fire_rebase_event("ok")
+end
+
 function M.continue()
   return rebase_command(cli.rebase.continue)
 end
--- a/lua/neogit/lib/git/rebase.lua
+++ b/lua/neogit/lib/git/rebase.lua
@@ -75,13 +75,14 @@ function M.onto(start, newbase, args)
   end
 end
 
-function M.reword(commit, message, _args)
-  local result = cli.commit.allow_empty.only.message("amend! " .. commit .. "\n\n" .. message).call_sync()
-
+function M.reword(commit, message)
+  local result = cli.commit.allow_empty.only.message("amend! " .. commit .. "\n\n" .. message).call()
   if result.code ~= 0 then
     return
   end
-  result = cli.rebase.env({ GIT_SEQUENCE_EDITOR = ":" }).interactive.autosquash.commit(commit).call_sync()
+
+  result =
+    cli.rebase.env({ GIT_SEQUENCE_EDITOR = ":" }).interactive.autosquash.autostash.commit(commit).call()
   if result.code ~= 0 then
     return
   end
--- a/lua/neogit/popups/rebase/actions.lua
+++ b/lua/neogit/popups/rebase/actions.lua
@@ -103,8 +103,9 @@ function M.reword(popup)
       return
     end
   end
+  -- TODO: Support multiline input for longer commit messages
   local old_message = git.log.message(commit)
-  local new_message = vim.fn.input("Message: ", old_message)
+  local new_message = input.get_user_input("Message: ", old_message)
 
   git.rebase.reword(commit, new_message, popup:get_arguments())
 end
--- a/lua/neogit/popups/rebase/actions.lua
+++ b/lua/neogit/popups/rebase/actions.lua
@@ -93,6 +93,22 @@ function M.interactively(popup)
   end
 end
 
+function M.reword(popup)
+  local commit
+  if popup.state.env.commit then
+    commit = popup.state.env.commit
+  else
+    commit = CommitSelectViewBuffer.new(git.log.list()):open_async()[1]
+    if not commit then
+      return
+    end
+  end
+  local old_message = git.log.message(commit)
+  local new_message = vim.fn.input("Message: ", old_message)
+
+  git.rebase.reword(commit, new_message, popup:get_arguments())
+end
+
 function M.subset(popup)
   local newbase = FuzzyFinderBuffer.new(git.branch.get_all_branches())
     :open_async { prompt_prefix = "rebase subset onto" }
--- a/lua/neogit/popups/rebase/actions.lua
+++ b/lua/neogit/popups/rebase/actions.lua
@@ -107,7 +107,10 @@ M.reword = operation("rebase_reword", function(popup)
 
   -- TODO: Support multiline input for longer commit messages
   local old_message = git.log.message(commit)
-  local new_message = input.get_user_input("Message: ", old_message)
+  local new_message = input.get_user_input("Message", { default = old_message })
+  if not new_message then
+    return
+  end
 
   git.rebase.reword(commit, new_message or "", popup:get_arguments())
 end)
--- a/lua/neogit/popups/rebase/actions.lua
+++ b/lua/neogit/popups/rebase/actions.lua
@@ -112,7 +112,7 @@ M.reword = operation("rebase_reword", function(popup)
     return
   end
 
-  git.rebase.reword(commit, new_message or "", popup:get_arguments())
+  git.rebase.reword(commit, new_message)
 end)
 
 function M.subset(popup)
--- a/lua/neogit/popups/rebase/actions.lua
+++ b/lua/neogit/popups/rebase/actions.lua
@@ -1,6 +1,7 @@
 local git = require("neogit.lib.git")
 local input = require("neogit.lib.input")
 local notification = require("neogit.lib.notification")
+local operation = require("neogit.operations")
 
 local CommitSelectViewBuffer = require("neogit.buffers.commit_select_view")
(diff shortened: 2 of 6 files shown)


lua/neogit/popups/rebase/actions.lua
```diff
@@ -106,6 +106,22 @@ function M.interactively(popup)
...
+  if popup.state.env.commit then
+    commit = popup.state.env.commit
+  else
+    commit = CommitSelectViewBuffer.new(git.log.list()):open_async()[1]
+    if not commit then
+      return
+    end
+  end
+  local old_message = git.log.message(commit)
+  local new_message = vim.fn.input("Message: ", old_message)
```
CKolkey, line 120: Line 2 in this file we import the input lib - use that for getting input.

lua/neogit/lib/git/rebase.lua
```diff
@@ -58,6 +59,56 @@ function M.onto(start, newbase, args)
   end
 end
 
+function M.reword(commit, message, args)
```
CKolkey, line 62: For new functions, I always add type annotations. A lot of the codebase predates their use, but every little bit helps :)

lua/neogit/lib/git/rebase.lua
```diff
@@ -58,6 +59,56 @@ function M.onto(start, newbase, args)
...
+        "--allow-empty",
+        "--only",
+        "-m",
+        "amend! " .. commit .. "\n\n" .. message,
+      },
+      on_error = function()
+        return false
+      end,
+    })
+    :spawn_blocking()
```
CKolkey, line 79: I think something like this would work:
`cli.commit.allow_empty.only.message(("amend! %s\n\n%s"):format(commit, message)).call_sync()`

I don't know if all of those fields are defined in the git.cli lib, but thats not too hard. Look for `commit = config { ... }`. I'm on mobile, so if that doesn't actually exist I'll add it tonight.

lua/neogit/lib/git/rebase.lua
```diff
@@ -58,6 +59,56 @@ function M.onto(start, newbase, args)
...
+        "rebase",
+        "-i",
+        "--autosquash",
+        commit .. "^",
+      },
+      on_error = function()
+        return false
+      end,
+    })
+    :spawn_blocking()
```
CKolkey, line 102: Same as above, but for the `-c` I think you have to pass it to `prefix()` or something like that. If you launch nvim with NEOGIT_LOG_LEVEL=trace and NEOGIT_LOG_FILE=true you can tail the log and see the exact cli strings being made. Or a well placed print() call in cli down towards the bottom where it does the string format call to build it all.

Again, on mobile, so its a bit harder to look up :)

CKolkey (changes requested):
Cool technique! Didnt know that.. and I think it can solve my "instant fixup" not being "instant" issue.

lua/neogit/lib/git/rebase.lua
gollth: Ahh okay I see. This is a method chaining API 🤦 Think I could solve this now. Thanks

lua/neogit/popups/rebase/actions.lua
gollth: Done

lua/neogit/lib/git/rebase.lua
gollth: added

lua/neogit/lib/git/rebase.lua
gollth: Using the `git.cli...` now

tests/specs/neogit/popups/rebase_spec.lua
```diff
@@ -0,0 +1,46 @@
...
+end
+
+describe("rebase popup", function()
+  it(
+    "rebase to reword HEAD~1",
+    in_prepared_repo(function()
+      local new_commit_message = "foobar"
+      local original_branch = gb.current()
+      input.values = { new_commit_message }
+      -- TODO: need to mock `CommitSelectViewBuffer`
```
gollth, line 25: @CKolkey I want to add unit tests for this, but I'm failing to mock the `CommitSelectViewBuffer` =S Do you have any idea how this could be achieved, so that this test can pass the commit to reword into the SUT?

gollth: ...or maybe there is a way to mock `popup.state.env.commit = "HEAD~1"` ?

lua/neogit/lib/git/rebase.lua
gollth: Simply using `cli.rebase.env({ GIT_SEQUENCE_EDITOR = ":" })....` is working for me

tests/specs/neogit/popups/rebase_spec.lua
CKolkey: Sure thing - does something like this work?
```lua
-- tests/mocks/commit_select_buffer.lua

local CommitSelectViewBuffer = require("neogit.buffers.commit_select_view")

local M = {
  values = {},
}

function M.add(value)
  table.insert(M.values, value)
end

CommitSelectViewBuffer.open_async = function()
  return table.remove(M.value, 1)
end

return M
```

```lua
-- tests/spec/whatever_spec.lua
local CommitSelectViewBufferMock = require("tests.mocks.commit_select_buffer")

-- ...
it("tests something", in_prepared_repo(function()
    CommitSelectViewBufferMock.add("commit to return")
    -- ...
end))
```

CKolkey: > ...or maybe there is a way to mock `popup.state.env.commit = "HEAD~1"` ?

If you wanna do something like that you can do a unit test for the action and just pass a popup table to the function - skip the integration test and treat the action like a function. I _think_ that'll work, unless some coroutine stuff doesn't like that

gollth: Cool, the `CommitSelectViewBufferMock` worked quite nicely!

gollth:
I added now integration tests for the "rebase to reword" functionality into `tests/specs/neogit/popups/rebase_spec.lua`. Probably it makes sense to add tests for the other kinds of rebase as well at some point. I can have a look at that, if I find the time

lua/neogit/popups/rebase/actions.lua
```diff
@@ -93,6 +94,24 @@ function M.interactively(popup)
...
+  else
+    commit = CommitSelectViewBuffer.new(git.log.list()):open_async()[1]
+    if not commit then
+      return
+    end
+  end
+
+  -- TODO: Support multiline input for longer commit messages
+  local old_message = git.log.message(commit)
+  local new_message = input.get_user_input("Message: ", old_message)
```
CKolkey, line 110:
```suggestion
  local new_message = input.get_user_input("Message", old_message)
```

I cleaned up the `get_user_input()` function the other day, so the ": " at the end is added automatically.

lua/neogit/lib/git/rebase.lua
```diff
@@ -58,6 +58,20 @@ function M.onto(start, newbase, args)
   end
 end
 
+function M.reword(commit, message)
```
CKolkey, line 61: In the action you're passing the popup arguments as a third parameter to this function:
```lua
git.rebase.reword(commit, new_message or "", popup:get_arguments())
```

Presumably this can be used in the cli call on 67 to determine if autosquash/autostash should be enabled? Unless those are a hard-requirement for this to function properly.

Whichever way this needs to be changed is alright - but can you add type-annotations to this function?

```lua
---@param commit string
---@param message string
---@param arguments table
---@return nil
```

tests/mocks/commit_select_buffer.lua
```diff
@@ -0,0 +1,18 @@
+local CommitSelectViewBuffer = require("neogit.buffers.commit_select_view")
+local git = require("neogit.lib.git")
+
+local M = {
+  values = {},
+}
+
+--- Add a rev name to the mocked list of commits the user selected.
+--- @param rev string the rev name of the commit which the user will select once `CommitSelectViewBuffer:open_async()` is called
```
CKolkey, line 9:
```suggestion
---Add a rev name to the mocked list of commits the user selected.
---@param rev string the rev name of the commit which the user will select once `CommitSelectViewBuffer:open_async()` is called
```

I don't know the reason, but lua-lsp doesn't want a space here 🤷🏼

tests/specs/neogit/popups/rebase_spec.lua
```diff
@@ -0,0 +1,76 @@
...
+  it(
+    "rebase to reword HEAD~1 from log view",
+    in_prepared_repo(function()
+      act("ll") -- log branches and go down one commit
+      operations.wait("log_current")
+      test_reword("HEAD~1", "foo")
+    end)
+  )
+
+  it(
```
CKolkey, line 49: Ooo, this is a cool way to test the autocmds. Very nice.

CKolkey (changes requested):
Some minor changes needed, but overall very slick. Much appreciated :D

lua/neogit/popups/rebase/actions.lua
CKolkey: Ohh, just realized that we can't really tell the difference between "cancelled" and "empty input". Hmm.

tests/specs/neogit/popups/rebase_spec.lua
gollth: Wasn't sure if this is the way to do the async stuff in lua, but with the timeout, it should be pretty robust I hope

lua/neogit/lib/git/rebase.lua
gollth: Hmm, so `autosquash` is needed for sure for this to work, otherwise you get this "amend! ... " stuff as a separate commit.

gollth: `autostash` is a different story. Technically this is not required. But during testing I found it a bit annoying to type `r w` and then get an error just because I had a dirty worktree. I think the reason why this is not a default in git is because the rebase can fail during the `git stash pop`. However since we do not change content during the rewording, this popping can really never fail AFAICS. I would personally leave it enabled by default like this, but if you say it should come from the popup args I'll refactor =)

CKolkey: Nah, I think you made the right call. Leave it as is :)

tests/specs/neogit/popups/rebase_spec.lua
CKolkey: Yea, it's... weird. I think this is fine :) We'll find out how robust it is in time haha

CKolkey:
As I suspected, this "instant rebase" tricks worked a treat for instant-fixup and instant-squash :) https://github.com/NeogitOrg/neogit/pull/1131

lua/neogit/popups/rebase/actions.lua
gollth: User can cancel just fine with ESC no? Then the whole rebase is never even started

lua/neogit/popups/rebase/actions.lua
```diff
@@ -93,6 +94,27 @@ function M.interactively(popup)
...
+  end
+
+  -- TODO: Support multiline input for longer commit messages
+  local old_message = git.log.message(commit)
+  local new_message = input.get_user_input("Message", { default = old_message })
+  if not new_message then
+    return
+  end
+
+  git.rebase.reword(commit, new_message or "", popup:get_arguments())
```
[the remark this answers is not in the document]
gollth: done

gollth:
@CKolkey feel free to fix all remaining issues yourself and push directly to my branch

CKolkey:
Looks good to me - I'll see about getting the editor set up so it does a proper buffer to edit the message in a bit, but this is definitely good enough to ship. Thanks for the contribution :D

Issue #807, opened 2023-09-01
Issues Related to "reverse" color schemes


chuck-flowers:
### Description

This is related to another issue that I opened and was partially fixed (#793).

Instead of a hard error now, my highlighted changes are not legible because I believe the foreground and background highlighting are the same color.

### Neovim version

NVIM v0.9.1
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

### Operating system and version

WSL2 Ubuntu 22.04 / Arch Linux Latest

### Steps to reproduce

1. Open neovim in a project with git changes
2. Open the neogit status page
3. Highlight

### Expected behavior

I expect the highlighted code blocks to be readable instead of solid colors (like in the below image)

### Actual behavior

![image](https://github.com/NeogitOrg/neogit/assets/9027535/58e1ea75-987e-403a-afee-bc952a5832ce)

### Minimal config

```Lua
-- gruvbox config (lazy)
return {
	'ellisonleao/gruvbox.nvim',
	config = function()
		local gruvbox = require 'gruvbox'

		gruvbox.setup({
			transparent_mode = false
		})

		vim.opt.termguicolors = true
		vim.o.background = 'dark'
		vim.cmd('colorscheme gruvbox')
	end
}

-- neogit config (lazy)
return {
	'NeogitOrg/neogit',
	dependencies = {
		'nvim-lua/plenary.nvim',
		'nvim-telescope/telescope.nvim',
		'sindrets/diffview.nvim'
	},
	opts = {},
	keys = {
		{
			'<leader>gs',
			function()
				require('neogit').open()
			end
		},
		{
			'<leader>gc',
			function()
				require("neogit").open({ "commit" })
			end
		},
		{
			'<leader>gl',
			function()
				require("neogit").open({ "log" })
			end
		}
	}
}
```

zdcthomas:
I'm seeing the exact same behavior +1

chuck-flowers:
@zdcthomas I might take a crack at writing a fix this weekend if I have time.

CKolkey:
IMO the real fix is to open a PR for your colorschemes and get them to support all the highlight groups.. All the highlight logic I added is (mostly) a stop-gap. But, by all means, I'd happily accept improvements :)

jamesbvaughan:
[...]

PR #1126, opened 2024-01-12
Fix highlighting when showing the commit message

sealor:
I had issues with highlighting on light background and fixed them.

Issue https://github.com/NeogitOrg/neogit/issues/807 describes a similar problem and can IMHO be closed.


CKolkey:
Cool :) I dig the change

PR #1127, opened 2024-01-12
Fix color shading on light background

sealor:
I had issues with highlighting on light background and fixed them.

Issue https://github.com/NeogitOrg/neogit/issues/807 describes a similar problem and can IMHO be closed.


CKolkey:
Makes sense - do you mind posting some before/after screenshots?

sealor:
Left: before fix
Right: after fix

On the left side one hunk header is invisible and some background colors are "missing".

@CKolkey Thanks for implementing such a nice Neovim plugin!

PR #1129, opened 2024-01-12
docs: update README.md

trimclain:
Sync README with the [default config](https://github.com/NeogitOrg/neogit/blob/d51199a3552571a02b6d3d88b95acc4102ef2fd2/lua/neogit/config.lua#L198)


CKolkey:
Thanks!

PR #1131, opened 2024-01-14
Allow instantly rebasing what would otherwise require the editor

CKolkey:
Use technique from https://peterevans.dev/posts/how-to-rewrite-git-commit-messages-non-interactively/


gollth:
Works like a charm!

PR #1132, opened 2024-01-14
Update rebase/reword to use full editor experience

CKolkey:
Builds on top of @gollth's work in #1118 

todo: 
- Gotta fix the tests now that they don't use the input...


gollth:
Hey, great to see the original Magit-way of doing the rewording! Was just testing this and noticed it does not (yet) work with multiline git messages. Probably you know, just wanted to mention =)

PR #1135, opened 2024-01-15
Add diff range functionality

smatting:
This PR implements missing `(d)iff (r)ange` functionality. The user can choose ranges like this:

- select two refs `<a>` and `<b>` via fuzzyfinder then choose between `<a>...<b>` and `<a>..<b>` 
- choose from predefined ranges `<b>...HEAD`, `<b>..HEAD` for `<b>` in `upstream`, `neogit.baseBranch`. These predefined ranges are useful:
   - If you configure `neogit.baseBranch` to the target branch of your PR then `<b>...HEAD` is the same diff that github shows you for the PR. 
   - For `<b>` as the upstream `<b>..HEAD` are the unpushed changes
- manually specify a range via input


smatting:
I noticed that if I open the `diff range` popup and a mouse click outside makes the popup disappear and subsequently the popup cannot be opened after. This problem doesn't occur if I close the popup via `<ESC>`. I'm not sure if this a result of the changes of this PR. However with different popups such a "(b)ranch (c)reate" this doesn't happen.

CKolkey (changes requested):
Hell yeah! Some minor notes, but otherwise looks alright :)

smatting:
Thanks for the review! I applied your suggestions.

CKolkey:
Nice. Doesn't have to be now, doesn't have to be you, but at some point, if the user has made a selection of commits, _that_ should be used as the range (first/last commit in selection). 

Thanks for the contribution :)

PR #1142, opened 2024-01-21
Fix: respect disable_line_numbers(=false) setting

taeruh:


CKolkey:
Classic `nil`s. Thanks :)

PR #1151, opened 2024-01-26
Fix section expansion sometimes not working

### LINKED ISSUE #1150: Expanding the "Recent commits" section sometimes doesn't work
### Description

As per title - I've been running into this issue for ages and I've finally managed to figure out a consistent way to cause the issue.

### Neovim version

NVIM v0.10.0-dev-2191+g4e59422e1d
Build type: RelWithDebInfo
LuaJIT 2.1.1702233742

Arch Linux

### Steps to reproduce

1. Open nvim in a git repo with a few commits, and one stash
2. Open Neogit
3. Expand, and then collapse, the "Recent commits" section
4. Expand the "Stashes" section
5. Try to expand the "Recent commits" section, and observe that nothing happens

This also happens in other situations, after expanding unstaged/changes as well - these are just the simplest steps to consistently trigger the bug.

### Expected behavior

Expanding the "Recent commits" section should work.

### Actual behavior

Here's a screen recording demonstrating the issue:

https://github.com/NeogitOrg/neogit/assets/49614525/bfc40e46-1425-489e-abb9-bbd2281dd624

### Minimal config

```Lua
-- No changes made

local M = {}

function M.root(path)
end

  end

    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
    if vim.v.shell_error > 0 then
        vim.log.levels.ERROR)
    end
  end
end

function M.setup(plugins)

  -- Install required plugins
  if plugins ~= nil then
    end
  end

    callback = function()
      vim.fn.system({
        "rm",
        "-r",
        "-f",
        M.root("xdg")
      })
    end
  })
end

M.setup({
})
```

AThePeanut4:
While attempting to work out consistent reproduction steps for #1150, I added a bunch of debug prints in the `toggle` function to try and figure out what was going on. This ended up giving me a fairly good idea as to what was actually causing the issue, after which I couldn't resist just trying to fix the issue myself :joy:

I've tried my best to make sure that my changes to `get_selection` haven't broken anything else, but I'm not at all familiar with the codebase so I could easily have missed something somewhere.

Fixes #1150

### DIFF
--- a/lua/neogit/popups/init.lua
+++ b/lua/neogit/popups/init.lua
@@ -40,18 +40,18 @@ end
 --- Returns an array useful for creating mappings for the available popups
 ---@return table<string, Mapping>
 function M.mappings_table()
-  ---@param commit CommitLogEntry|nil
   ---@return string|nil
-  local function commit_oid(commit)
-    return commit and commit.oid
+  local function get_selected_commit_oid()
+    local s = require("neogit.status").get_selection()
+    return s.item and s.item.commit and s.item.commit.oid
   end
 
-  ---@param commits CommitLogEntry[]
   ---@return string[]
-  local function map_commits(commits)
+  local function get_selected_commit_oids()
+    local s = require("neogit.status").get_selection()
     return vim.tbl_map(function(v)
       return v.oid
-    end, commits)
+    end, s.commits)
   end
 
   return {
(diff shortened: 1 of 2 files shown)


CKolkey:
So, I really appreciate you taking time to get into this, but I'm actually nearly done rewriting the status buffer here: https://github.com/NeogitOrg/neogit/pull/1161

Main reason was that it was the last buffer to _not_ use the internal UI framework, and I dislike inconsistency. I don't think this issue occurs with the new version. Hopefully that'll be done soon :)

Anyways, I'm going to close this since I don't really wanna add more stuff I need to reconcile. Just wanted to say that if it _wasn't_ for the rewrite, this would have been a great addition.

PR #1152, opened 2024-01-27
fix: typo in the kind type

Danielkonge:
I ran into this typo in the type annotations while changing my config. (Alternatively, if the string is supposed to be "float", a change elsewhere in the code needs to be made.)


CKolkey:
Keen eye 👁️

PR #1154, opened 2024-01-30
Add count support for line_action in reabse todo

PostCyberPunk:
and a fix for skipping "edit"

### DIFF
--- a/lua/neogit/buffers/rebase_editor/init.lua
+++ b/lua/neogit/buffers/rebase_editor/init.lua
@@ -12,28 +12,38 @@ local M = {}
 
 local function line_action(action, comment_char)
   return function(buffer)
-    local line = vim.split(vim.api.nvim_get_current_line(), " ")
-    if line[1] == comment_char then
-      table.remove(line, 1)
-    end
-
-    -- Check if line is "break" or "exec"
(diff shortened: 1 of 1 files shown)


lua/neogit/buffers/rebase_editor/init.lua
```diff
@@ -63,6 +74,7 @@ function M:open(kind)
     filetype = "NeogitRebaseTodo",
     buftype = "",
     kind = kind,
+    disable_line_numbers = false,
```
CKolkey, line 77: This could be set via config instead - there's already a "disable_line_numbers" value that could be read.

CKolkey (changes requested):
Small change about the line numbers, otherwise looks good to me :)

lua/neogit/buffers/rebase_editor/init.lua
PostCyberPunk: here's my config,but i have no line number in rebase editor. is my config wrong?
```lua
opts={
      disable_line_numbers = false,
      rebase_editor = {
        disable_line_numbers = false,
      },
```
any way ,hard coded line number is still a bad idea,i will delete this commit.

CKolkey:
Good stuff, thanks for the contribution :)

PR #1161, opened 2024-02-02
Rewrite status buffer to use the same UI lib as all other buffers

CKolkey:
Longer WIP to finally rewrite the status buffer to use the same UI lib as every other buffer. It'll be nice to have a single UI lib :)

Todo:
- on-close hook with all the teardown stuff. Probably in bufunload autocmd
- Visual Mode:
	- Stage
	- Unstage
	- Discard
	- Ignore Popup Filepaths
- When launching the fuzzy finder, any refresh attempted will raise an exception because the set_folds() function cannot be called when the buffer is not focused, as it's not a proper API. We could implement some kind of freeze mechanism to prevent the buffer from refreshing while the fuzzy finder is open. 
- Add logging to status buffer
- Add notifications to status buffer
- Discard Stash
- Discard Section
- Discarding a RENAME should set the file back to original
- Goto{Prev,Next}HunkHeader doesn't cross file boundaries
- Deprecate status.refresh_viml_compat
- Deprecate status.refresh_manually - just use refresh
- finish cursor placement logic around refreshing buffer


lua/neogit/buffers/commit_view/init.lua
```diff
@@ -114,12 +114,10 @@ function M:open(kind)
           -- in order to use them as match patterns.
           local selected_path = vim.fn.trim(c.value)
 
-          local diff_headers = {}
-
           -- Recursively navigate the layout until we hit NeogitDiffHeader leafs
           -- Forward declaration required to avoid missing global error
```
CKolkey, line 118: remove this note

lua/neogit/popups/commit/init.lua
```diff
@@ -19,6 +19,10 @@ function M.create(env)
     :option("C", "reuse-message", "", "Reuse commit message")
     :group_heading("Create")
     :action("c", "Commit", actions.commit)
+    :action_if(vim.fn.has("git-absorb") == 0, "b", "Absorb", function()
+      -- TODO: https://github.com/SuperBo/fugit2.nvim/blob/dev/lua/fugit2/git_absorb.lua
+      vim.system({ "git", "absorb", "--and-rebase" }, { env = { GIT_SEQUENCE_EDITOR = ":" } }):wait()
+    end)
```
CKolkey, line 25: remove this from PR

PR #1162, opened 2024-02-03
FIX: Proper quiting of `stash rename`

gollth:
Sorry just noticed a little bug, with the `stash rename`: When input is `nil` (e.g. when user quit the message prompt with ESC) then the stash is dropped =S


CKolkey:
Good catch - this is one of the more annoying things in lua, all the "if nil then return" business. Maybe theres a nicer way to go about it, but its pretty much -everywhere- we receive user input 😓

PR #1164, opened 2024-02-05
Fix #1047

### LINKED ISSUE #1047: Stage operation fails
### Description

When trying to Stage some hunk, I get an error :

```
git apply --cached:

    error: C:\Users\Cibrario\code\pygris\pygris/src/pygris/macros/rf_pll.py: does not exist in index

An error occured.
```
Given the mix of `/` and `\`, I can guess it's windows related ?

### Neovim version

NVIM v0.9.4
Build type: RelWithDebInfo
LuaJIT 2.1.1696883897

Windows 11

### Steps to reproduce

Open Neogit buffer, go over an unstaged hunk and type `s`.

### Expected behavior

the hunk should be staged

### Actual behavior

the error occur

### Minimal config

```Lua
local M = {}

function M.root(path)
end

  end

    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
    if vim.v.shell_error > 0 then
        vim.log.levels.ERROR)
    end
  end
end

function M.setup(plugins)

  -- Install required plugins
  if plugins ~= nil then
    end
  end

    callback = function()
      vim.fn.system({
        "rm",
        "-r",
        "-f",
        M.root("xdg")
      })
    end
  })
end

M.setup({
})
```

saggitar:
_Cause of bug_:
Command used to get initial repository directory returns unix path seperators on windows.
This confuses the `plenary.path` module.

_Fix_:
Use `plenary.path` to convert the output of the git command to windows style path

### DIFF
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -1,6 +1,7 @@
 local logger = require("neogit.logger")
 local process = require("neogit.process")
 local util = require("neogit.lib.util")
+local Path = require("plenary.path")
 
 local function config(setup)
   setup = setup or {}
@@ -572,7 +573,8 @@ local function git_root_of_cwd()
     :spawn_blocking()
 
   if process ~= nil and process.code == 0 then
-    return process.stdout[1]
+    local out = process.stdout[1]
+    return Path:new(out):absolute()
   else
     return ""
   end


CKolkey:
Very much appreciated :)

PR #1175, opened 2024-02-12
 Fix: graph invalid UTF-8 codepoint

wroyca:
This PR fixes a scenario where UTF-8 codepoints passed to gsub would be swapped out with their Unicode replacement.

#### before:
image

#### after:
image


wroyca:
cc @CKolkey

CKolkey:
Brilliant, thank you :)

Issue #1108, opened 2024-01-05
Diff view shows a lot of noise caused by blank lines


miguno:
### Description

Opening the diff side-by-side view with `dd` shows a lot of noise caused by blank lines that are (seemingly) being considered as removed (`--------------`) and then added as a true blank linkes (`         `) as shown in the screenshot below.

<img width="1339" alt="Screenshot 2024-01-05 at 19 07 28" src="https://github.com/NeogitOrg/neogit/assets/294849/0e666a89-cbe5-4653-b9b3-89968cb80b00">

Notice how the only true change in the screenshot above is the line on the right in which "// foo" and a newline was added. The other highlighted changes are simply unchanged newlines.

Could this be a legitimate neogit issue, or is it more likely an issue with (say) neovim's highlight groups in my setup?  I can work on a minimal config if needed.

Environment:
- My neovim theme is https://github.com/folke/tokyonight.nvim
- Neovim version and OS: see GH issue fields
- Update 2024-01-08: I am using [diffiview.nvim](https://github.com/sindrets/diffview.nvim), and I don't this issue when calling `:DiffviewOpen` manually.

### Neovim version

NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1692716794

### Operating system and version

macOS 14.2.1

### Steps to reproduce

1. Create and commit a file in a git repository that includes newlines.

Example:

```
1

2

3

4
```

2. Open the file in neovim and make a simple change, such as modifying an existing line:

```
1

2 Hello, modification!

3

4
```

3. Run `:Neogit` and open the diff side-by-side view with `dd`.

### Expected behavior

Only the actual change should be highlighted with red vs. green colors.  In my screenshot above, only the two lines for the "// foo" change.

### Actual behavior

See screenshot.  All newlines are incorrectly highlighted as having been changed, at least that's what the coloring indicates.

### Minimal config

```Lua
can work on this if needed
```

dbatten5:
Getting this issue as well. I'll add that I'm using `diffview` and if I run `:DiffviewOpen` then I don't see this issue

miguno:
I am using [diffiview.nvim](https://github.com/sindrets/diffview.nvim), too, but I see this issue both when calling neogit and when calling `:DiffviewOpen` manually.

murtaza64:
hmm... I get this both with neogit and diffview

miguno:
[...]

PR #1181, opened 2024-02-21
Fixed unwanted trim for file from diffview

eternity74:
call_sync method trims output by default. DiffView needs to read file as it is.
issue:'Diff view shows a lot of noise caused by blank lines #1108'


miguno:
Thanks for this fix, @eternity74 !

PR #1184, opened 2024-02-23
Use plenary.job for call_sync

eternity74:
This change helps to capture output of the command correctly. 
Current neogit.process implementation doesn't capture stdout correctly regarding empty lines and EOF.

This would help to remove the last empty line issue on #1108


CKolkey:
seems reasonable, but I'll revisit the job handling when 0.10 drops and we get the `vim.system` api.

CKolkey:
Going to shelve this as it's more likely we'll use `vim.system` instead

Issue #1176, opened 2024-02-14
Status buffer broken during rebase with unmerged paths


Aphosis:
### Description

Hey !

I noticed an issue recently, if I stop during a rebase with a mix of modified (no conflict) and unmerged paths (conflicts), the status buffer is broken.

More specifically, two lines that were displayed as `Modified` and `Both Modified` previously are now merged together on a single line, separated by a tab character.

I did some `git bisect`ing and identified 0281db3089503ddadcdabffe279e2f96fb7f7483 as the commit that introduced the bug.

Here is an exemple of the output I get now:

```
Hint: [<tab>] toggle diff | [s] stage | [u] unstage | [x] discard | [c] commit | [?] help

Head:     sha (detached) Previous commit
Tag:      tag (0)

Rebasing: my-branch (1/2)
Staged changes (4)
Modified       a.txt
Modified       b.txt
Modified       c.py	u UU N... 100644 100644 100644 100644 sha1 sha2 sha3 d.txt

Stashes (4)
Recent commits (10)
```

Here is the output I get when I use dab4e50: 

```
Hint: [<tab>] toggle diff | [s] stage | [u] unstage | [x] discard | [c] commit | [?] help

Head:     sha (detached) Previous commit
Tag:      tag (0)

Rebasing: my-branch (1/2)
Unstaged changes (1)
Both Modified  d.txt

Staged changes (4)
Modified       a.txt
Modified       b.txt
Modified       c.txt

Stashes (4)
Recent commits (10)
```

I noticed that `update_status` was mentioned in !691, and I can see the `string.format` in https://github.com/NeogitOrg/neogit/commit/0281db3089503ddadcdabffe279e2f96fb7f7483#diff-70ea688606f2254c83e78a2d5fc72635f23e5b8505042d44b4157e1b54aca666R65 is likely what is causing the issue, but I failed to find an easy way to fix my issue without reverting 0281db3089503ddadcdabffe279e2f96fb7f7483.

I think this has to do with how those results are later matched using the various `match_1`, `match_2` and `match_u`, but my investigation ended here.

The minimal config was enough to reproduce on my side, but this may be a bit cumbersome to reproduce since it requires both conflicting and non conflicting files during a rebase (at least I think this is how I end up with the bug).

If you need more info don't hesitate to reach out, thank you for your time, and for the awesome plugin !

### Neovim version

NVIM v0.10.0-dev-594ff34
Build type: RelWithDebInfo
[...]

PR #1189, opened 2024-02-27
Discard Conflicted Files

gollth:
Hi @CKolkey, me again (=

AFAIK so far there is no option to discard files which are in conflict. I really liked the Magit way of handling this, that is asking the user to use "ours" or "theirs", so I recreated that behavior in Neogit. 

[Neogit-discard-conflict.webm](https://github.com/NeogitOrg/neogit/assets/31999281/d26bd443-6562-40a7-be28-9ac5b9435d48)

## Notes

1. This does not work yet rebased on master, because of #1176. Best is not to merge before this is fixed
2. This does only for for entire files up to now. Discarding conflicted hunks does not work. I want to try this next, but have to understand the "patching" logic for hunks. Ideas/explanations welcome

### DIFF
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -214,6 +214,8 @@ local configurations = {
     flags = {
       _track = "--track",
       detach = "--detach",
+      ours = "--ours",
+      theirs = "--theirs",
     },
     aliases = {
       track = function(tbl)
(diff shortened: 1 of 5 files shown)


CKolkey:
Nice! This is great - something I've wanted to add for a while, but I've been holding off on adding stuff to the status buffer until https://github.com/NeogitOrg/neogit/pull/1161 is done. Which... is close! 

Anyways, for the hunk stuff, it looks like Magit uses *smerge* under the hood: https://github.com/emacs-mirror/emacs/blob/master/lisp/vc/smerge-mode.el#L411

... So maybe the hunk stuff can be in a follow-up PR.

CKolkey:
No... the CI situation is a bit lame in that respect. And since it feels a lot like work, I haven't dug into it much, haha. 

I'll just bench test the feature, then port it to the new status buffer

CKolkey:
This is awesome. Thank you :)

PR #1190, opened 2024-03-02
Use ; and , to scroll commit view inside log view

saep:
When viewing the commit log, pressing `;` or `,` can be used to open the commit view without focusing it. If you press `;` again, it will scroll down the commit view. With `,` you can scroll up. Because the commit view window isn't focused, you can quickly focus the next commit with `j` or the previous with `k`.

This reduces the amount of key presses significantly when reviewing changes of multiple commits because you don't have to switch windows twice for every commit you want to look at.

Since `f` and `t` are bound differently to normal vim usage, `;` and `,` are also basically free to use. On the US layout (and others), you only need the right hand to review the changes and can use the left hand to drink coffee or tee. ;-)

These bindings implement the same functionality as `magit-diff-show-or-scroll-down` and `magit-diff-show-or-scroll-up`.

This workflow is something that I've grown very accustomed to and one of the reasons I'm still using (Doom) Emacs at work. I also think, that it is useful for other people, so I've opened this PR. This is literally my whole magit-configuration:

```emacs-lisp
(map!
 :map magit-log-mode-map
 :g ";" 'magit-diff-show-or-scroll-down
 :g "," 'magit-diff-show-or-scroll-up
 )
```

### DIFF
--- a/lua/neogit/buffers/commit_view/init.lua
+++ b/lua/neogit/buffers/commit_view/init.lua
@@ -71,6 +71,32 @@ function M:close()
   self.buffer = nil
 end
 
+---Opens the CommitViewBuffer if it isn't open or performs the given action
+---which is passed the window id of the commit view buffer
+---@param commit_id string commit
+---@param filter string[]? Filter diffs to filepaths in table
+---@param action fun(window_id)
+function M.open_or_run_in_window(commit_id, filter, action)
+  if not commit_id then
+    return
+  end
+  local cvb = M.instance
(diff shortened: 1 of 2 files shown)


CKolkey:
Cool - I'm a bit backlogged right now with stuff, but we'll get this merged in good time. Thanks for the PR

CKolkey:
Alright. This is sweet. Nice work :)

CKolkey:
Just a heads up, on the nightly branch I made these keys user configurable

PR #1196, opened 2024-03-12
fix: unlock gpg key on commit

### LINKED ISSUE #1195: GPG key unlock doesn't work on commit
### Description

With gpgSign set in gitconfig and pinentry-loopback or pinentry-tty enabled, commits cannot be made using neogit.

```sh
# ~/.gnupg/gpg.conf
 pinentry-mode loopback
```
```sh
# ~/.gnupg/gpg-agent.conf
pinentry-program /usr/bin/pinentry-tty
allow-loopback-pinentry
```

Relates to : https://github.com/NeogitOrg/neogit/issues/370

### Neovim version

NVIM v0.10.0-dev-2489+gb413f5d04
Build type: RelWithDebInfo
LuaJIT 2.1.1707061634

Linux 6.7.9

### Steps to reproduce

1. :Neogit kind=auto
2. Stage some changes
3. `cc`
4. Errors or freezes, depending on pinentry setting

### Expected behavior

The user is prompted to unlock the gpg key and the changes are committed and the commit is signed.

### Actual behavior

When pinentry loopback is enabled, there is just an error.
When pinentry-tty is enabled without loopback, the UI gets messed up and nvim freezes

### Minimal config

```Lua
local M = {}

function M.root(path)
end

  end

    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
    if vim.v.shell_error > 0 then
        vim.log.levels.ERROR)
    end
  end
end

function M.setup(plugins)

  -- Install required plugins
  if plugins ~= nil then
    end
  end

    callback = function()
      vim.fn.system({
        "rm",
        "-r",
        "-f",
        M.root("xdg")
      })
    end
  })
end

M.setup({
})
```

trevarj:
Make do_commit an interactive call in case there is a need to type in a gpg password before signing a commit.

Add an `interactive` flag to the opts for a wrapped command.

Fixes #1195


CKolkey:
I'll have a chance to review this to see what kind of other things might be impacted soon. Really trying to make sure https://github.com/NeogitOrg/neogit/pull/978 is ready for when `0.10` drops.

CKolkey:
If you like, sure, but the surface area of your change is so small I don't think it would make a difference. Your call ;)

CKolkey:
Seems to be fine - very nice fix! Thanks for digging into the issue.

PR #1211, opened 2024-03-22
Rebase modify a commit

kjughx:
I often use interactive rebase to make changes to a single commit that's not HEAD. With this change that's possible with "rm<cr>" on a selected commit or "rm<cr>" will open git log where a commit can be selected.

A naive approach to unit test did not work. There are not enough commits in the "prepare repo" so I added an initial commit, then my test _sometimes_ succeeds but it also causes the test_reword test to fail so there's something that I don't understand.


lua/neogit/popups/rebase/actions.lua
```diff
@@ -115,6 +115,24 @@ M.reword = operation("rebase_reword", function(popup)
...
+    commit = popup.state.env.commit
+  else
+    commit = CommitSelectViewBuffer.new(git.log.list()):open_async()[1]
+    if not commit then
+      return
+    end
+  end
+  local message = git.log.message(commit)
+  if
+    input.get_confirmation('Modify commit "' .. message .. '"?', { values = { "&Yes", "&No" }, default = 1 })
```
CKolkey, line 130: I'm not sure we need to ask for confirmation here - the user has already hit `rm` and (possibly) chosen a commit.

lua/neogit/lib/git/rebase.lua
```diff
@@ -92,6 +92,17 @@ function M.reword(commit, message)
   fire_rebase_event("ok")
 end
 
+function M.modify(commit)
+  local short_commit = string.sub(commit, 1, 7)
+  local editor = "sed -i 's/^pick " .. short_commit .. "/edit " .. short_commit .. "/'"
```
CKolkey, line 97: Couple notes here. Something I learned the hard way is that a 7 character hash isn't a constant thing. Turns out (https://github.com/NeogitOrg/neogit/pull/1046), for every order-of-magnitude of commits in a repo, git automatically adds another character to the hash length! So, we can't count on it being 7.

CKolkey: Also, tried this out, and I get this error:

```
git rebase -i --autosquash --autostash 4c29337d72cbdb8187965e6839c18274d09508ea^:

    sed: 1: "/Users/cameron/code/neo ...": command c expects \ followed by text
    error: There was a problem with the editor 'sed -i 's/^pick 4c29337/edit 4c29337/''.
```

CKolkey (changes requested):
Good start - we can make this work. Out of curiosity, how portable is `sed -i`? I know there are a few different versions of sed around... I'd like to avoid these kind of dependencies _if possible_ and only rely on neovim.

Would it be possible to use nvim via a startup option to modify the file? It can run ex commands this way, so I think it's doable. Check out `:h cli-arguments`

lua/neogit/popups/rebase/actions.lua
kjughx: Fair enough, I removed it.

lua/neogit/lib/git/rebase.lua
kjughx: Hmm. I agree it would be nice to only depend on neovim. This works for me in the shell:
`GIT_SEQUENCE_EDITOR="nvim` -c '%s/^pick \(<SHORT_SHA>.*\)/edit \1/' -c 'wq'" git rebase -i  <LONG_SHA>^`
but not when I do:
` local editor="nvim -c '%s/^pick \\(" .. short_commit .. ".*\\)/edit \\1/' -c 'wq'"`.

The downside of matching like this in a huge repo is that there could be multiple matches on the 7 character hash which would edit two commits... not sure if that's acceptable or how to work around that.

kjughx: I added `in_pty(true)` to the call and then it worked.

CKolkey: Does it work with the full sha?

CKolkey: We could make a helper that calls `git rev-list --count --all` to determine the correct length of the hash. It would be handy in some other places to, so the ui layer shows the correct length of hash

kjughx: The command you sent just gives me some large number, e.g. in this repo 2973... however `git log -n 1 --format=%h` returns a hash where we can count the characters. I added a helper function in lib.git.log.

This repo has 8 characters in the hash and that works fine for me now.

kjughx: I made a similar implementation for "drop a commit". Should I add that in a separate pr or can I add it here?

lua/neogit/lib/git/log.lua
```diff
@@ -515,4 +515,8 @@ function M.reflog_message(skip)
     .call_sync({ ignore_error = true }).stdout
 end
 
+function M.abbreviated_size()
+  return string.len(cli.log.format("%h").max_count(1).call_sync().stdout[1])
```
CKolkey, line 519:
```suggestion
  return string.len(cli.log.format("%h").max_count(1).call({ hidden = true }).stdout[1])
```

`hidden = true` will prevent this from showing up in the git command history buffer.

lua/neogit/lib/git/rebase.lua
CKolkey: Awesome - let's just do that in a new PR :)

CKolkey (approved):

lua/neogit/lib/git/rebase.lua
CKolkey: Lets just do it in a new PR - sounds great :D

tests/specs/neogit/popups/rebase_spec.lua
```diff
@@ -15,34 +15,74 @@ local function act(normal_cmd)
 end
 
 describe("rebase popup", function()
-  function test_reword(commit_to_reword, new_commit_message)
+   before_each(function()
```
CKolkey, line 18:
```suggestion
  before_each(function()
```

PR #1213, opened 2024-03-23
Rebase drop a commit

kjughx:
Based on [this](https://github.com/NeogitOrg/neogit/commit/a5d5cc68dcd844eedc2db5a7c399dbec6ef1d49d) we can now also drop a commit.

I suggest to change the keybinding for dropping to "d" since it's unused in that view and feels more natural.

### DIFF
--- a/lua/neogit/lib/git/rebase.lua
+++ b/lua/neogit/lib/git/rebase.lua
@@ -106,6 +106,20 @@ function M.modify(commit)
   fire_rebase_event { commit = commit, status = "ok" }
 end
 
+function M.drop(commit)
+  local short_commit = require("neogit.lib.git").rev_parse.abbreviate_commit(commit)
+  local editor = "nvim -c '%s/^pick \\(" .. short_commit .. ".*\\)/drop \\1/' -c 'wq'"
+  local result = cli.rebase
+    .env({ GIT_SEQUENCE_EDITOR = editor }).interactive.autosquash.autostash
+    .in_pty(true)
+    .commit(commit)
+    .call()
+  if result.code ~= 0 then
+    return
+  end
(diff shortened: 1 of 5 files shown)


kjughx:
I realize that I broke a lot of tests with the previous PR, I will try to fix them.

lua/neogit/lib/git/rebase.lua
```diff
@@ -107,6 +107,20 @@ function M.modify(commit)
   fire_rebase_event { commit = commit, status = "ok" }
 end
 
+function M.drop(commit)
+  local short_commit = string.sub(commit, 1, log.abbreviated_size())
```
CKolkey, line 111: I forgot about this until now - you can use `git.rev_parse.abbreviate_commit(oid)` to get the correct short commit 😅

CKolkey (changes requested):

CKolkey:
Thanks - but don't stress about it.. The test suite kinda sucks and is _far_ flakier than I like. It's on my list of things to eventually rewrite it using either mini.test or https://github.com/lewis6991/nvim-test

lua/neogit/lib/git/rebase.lua
kjughx: Oh okay, then [this](https://github.com/NeogitOrg/neogit/commit/87b7fc6cfb516fe639687acf90f320f1f83321c9) is not very useful then. Should I remove it?

CKolkey: yeah, that's fine :)

CKolkey:
Great stuff - Since we were in the neighborhood, I went ahead and implemented Autosquash, which was the last action to go! Thanks for helping with this :)

PR #1226, opened 2024-03-29
rspec WIP

CKolkey:
experimental RPC test suite


deril:
It looks nice. But I am not sure that we need to bring another language to the codebase, which is needed to be supported.

CKolkey:
@deril if this was in common lisp, you would have an _entirely_ different attitude :P

CKolkey:
All good, man, I appreciate what you did regardless :)

Yeah, using ruby has a few advantages here:
- robust git gem to verify correctness
- non-async runtime
- robust stdlib for managing child processes/filesystem stuff
- rspec is a great testing framework

And of course... i thought it would be interesting to use RPC to test neovim. Thats probably the only real reason hahaha.

I guess the downside is that people would have to write tests in ruby, but it's basically just me doing this anyways so... whatever. I like ruby 🤷‍♂️

Oh, and you can only really do E2E tests, but I think thats ok. I already unit test lua stuff when it's hard to get "correct", which wouldn't change.

Issue #1233, opened 2024-04-03
Tracking Issue: Status buffer bugs on `nightly` branch


AThePeanut4:
Firstly, thank you to everyone involved in making this plugin, and specifically thank you @CKolkey for all the fixes, new features and improvements you've made over the last year or so. It really makes working with Git a breeze, and without it I certainly would not have been able to fully switch over to Neovim. 

### Description

So this is a list of several small issues I've come across while (ab)using the rewritten status buffer. A few might also have existed before #1161. For brevity I'm just going to give short descriptions of each, if you're unable to reproduce any of them I'll be happy to provide more detailed steps.

#### Bugs:
- [ ] #1242
- [x] When the Recent Commits section is collapsed, pressing j while the cursor is on the last line of the buffer (the Recent Commits header) now moves the cursor up by 1, rather than doing nothing.
- [x] I'm also seeing the cursor jump after returning from a popup - if the cursor was on the Hint/Head/Merge/Push lines, it jumps down to the Untracked files section.
- [x] Most of the time, when returning to the status buffer from a popup (e.g. view a commit and then close it), the Stashes and Recent Commits sections automatically close. This is by far the most annoying bug on this list.
- [x] There's some inconsistency when pressing `<Tab>` in the blank space between expanded sections. Sometimes it will collapse the section above, and sometimes it will do nothing. I've also come across a likely related issue, see below for reproduction steps.
- [x] Attempting to discard changes (`x`) to multiple _unstaged_ files at once (with Visual mode) throws an error.
- [x] Using `1`, `2`, `3` or `4` when the cursor is in any blank line acts as if the cursor was in the "Recent commits" section.
- [x] Opening the help popup (with `?`) while the cursor is on the "Hint:", "Head:", Merge:" or "Push:" lines throws an error.
- [x] Running `:Neogit` when Neogit is already open in another tabpage closes the current tabpage rather than switching to the previous Neogit tabpage.
- [x] Closing the Neogit tabpage with `q` if it is the last open tabpage throws an error. It should probably open a blank file in this case, or the alternate file.
- [x] ~I get a lot of `attempt to compare nil with number ... in function 'get_cursor_location'` errors. These also seem to happen when returning to the status buffer from a popup, but it again doesn't happen every time.~ Seems to no longer be happening.
[...]

PR #1244, opened 2024-04-07
Status buffer fixes

AThePeanut4:
This fixes the issues with `<Tab>` in blank spaces. The problem is that the `EmptyLine` component was actually an *instance* of a component, and that instance was being shared between all the sections in the status buffer. That means that when the component tree is rendered, that single instance gets added to `RendererIndex.index` multiple times, and its `position.row_start` field gets overwritten multiple times. This results in the `RendererIndex:find_by_line(line)` method returning a component with a `position.row_start` that is not equal to the `line` argument, and that breaks everything.

I've also fixed an error that pops up if the upstream or pushRemote branches don't have any commits, and made the Head line show `(detached)` again when detached. When detached, the commit hash will always be shown, irrespective of `show_head_commit_hash`. If `show_head_commit_hash` is enabled, and the branch doesn't have any commits, then it displays "0000000".

Ref: #1233

### DIFF
--- a/lua/neogit/buffers/common.lua
+++ b/lua/neogit/buffers/common.lua
@@ -14,6 +14,10 @@ local range = util.range
 
 local M = {}
 
+M.EmptyLine = Component.new(function()
+  return col { row { text("") } }
+end)
+
 local diff_add_start = "+"
 local diff_delete_start = "-"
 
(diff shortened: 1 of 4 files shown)


CKolkey:
Ahhhhhh, right 🤦🏼 . Much appreciated.

PR #1251, opened 2024-04-15
Add option to edit patch before staging

leiteg:
Hello! Thanks for this awesome plugin, I have been having a blast using it :)

I added a command to edit the patch before staging the file (shortcut `e`). This is equivalent to running `git add --edit -- file` in the command line. Do you guys have any interest in merging this?

In case this PR is missing something, I am willing to update it!

### DIFF
--- a/lua/neogit/config.lua
+++ b/lua/neogit/config.lua
@@ -376,6 +376,7 @@ function M.get_default_values()
         ["s"] = "Stage",
         ["S"] = "StageUnstaged",
         ["<c-s>"] = "StageAll",
+        ["e"] = "StageEdit",
         ["u"] = "Unstage",
         ["U"] = "UnstageStaged",
         ["$"] = "CommandHistory",
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -292,6 +292,7 @@ local configurations = {
     flags = {
       update = "-u",
       all = "-A",
+      edit = "--edit",
     },
   },
 
--- a/lua/neogit/lib/git/status.lua
+++ b/lua/neogit/lib/git/status.lua
@@ -186,6 +186,9 @@ local status = {
   stage_all = function()
     git.cli.add.all.call()
   end,
+  stage_edit = function(files, env)
+    return git.cli.add.edit.files(unpack(files)).env(env).show_popup(true):in_pty(true).call()
+  end,
   unstage = function(files)
     git.cli.reset.files(unpack(files)).call()
   end,
--- a/lua/neogit/status.lua
+++ b/lua/neogit/status.lua
@@ -925,6 +925,34 @@ local stage = operation("stage", function()
   }, "stage_finish")
 end)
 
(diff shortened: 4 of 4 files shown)


CKolkey:
Cool stuff! I'm not sure I understand the use case of this instead of just editing the buffer, though. Care to enlighten me?

leiteg:
@CKolkey it is a niche use case indeed! Sometimes in my workflow, I do a bunch of changes at once and would only then create several commits from these changes. Editing the buffer directly would require me to (i) delete the changes I do not want to stage, (ii) stage the remaining changes, (iii) bring the deleted changes back, (iv) commit, (v) repeat. This is perfectly doable for simple changes but it gets clumsy when there are many files.

You might argue that I can stage individual hunks already, but more often than not it is useful to stage only part of a hunk, that's where the `--edit` flag comes in handy!

CKolkey:
Ahh, ok! Try doing a visual selection of part of a hunk and staging that - I think that accomplishes what you want?

leiteg:
It does! I didn't know this was possible, I guess this PR is redundant after all.

Thanks for you attention and keep up the awesome work!

CKolkey:
Glad to hear it!

rmacklin:
Howdy folks! I'm a long time git CLI user but I'm experimenting with adopting a neogit workflow.

Re:
> Try doing a linewise visual selection (`V`) of part of a hunk and staging that - I think that accomplishes what you want?

That's quite handy - thanks for the tip!

However, there are still some cases where I find `git add -p`'s `e` (manually edit the current hunk) option useful that aren't solved this way. Basically, there are times when you really do want to change some part of a hunk before staging it, but leave the file as-is in your working directory.

For example: you've added a new call of a function `foo`, and then you've renamed the function (and all its usages) to `bar`. You might decide you want to stage the new invocation _without the `bar` rename_ to be committed right now. With `git add -p` you can use `e` on that hunk and make a temporary change of `bar` back to `foo` and then commit. (And then later you can commit the `bar` rename once you're ready, which would still be there in the working directory.)

I admit this is more rare than just staging individual hunks or lines inside a hunk, but it is something I do often enough that I'd love to have an equivalent option in neogit.

That said, @leiteg it sounds like you were proposing something slightly different, which was an equivalent to `git add --edit file` rather than operating on individual hunks (like `git add -p` -> `e`), is that correct? Personally, I'd want the hunk granularity...

leiteg:
> it sounds like you were proposing something slightly different, which was an equivalent to git add --edit file rather than operating on individual hunks (like git add -p -> e), is that correct?

Hey @rmacklin, yes, that's correct. I proposed the file-wise edit because it was simpler to implement, but I agree that hunk-wise editing would be a better feature to have. For my use cases, doing a visual select and then staging did the trick, as suggested by @CKolkey. If you wish, and the maintainers agree, you can pick up my patch and move it forward.

rmacklin:
I thought I'd have time to try moving your patch forward, but I haven't been able to. Just wanted to follow up here to make it clear I'm not actively working on it.

PR #1267, opened 2024-04-29
Handle git@server:repo.git as a shorthand for ssh://git@server/repo.git in git.remote.parse(url)

### LINKED ISSUE #1266: improve handling of user@host:repo.git and git://host/repo parsing
### Description

Per the Git Book (https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols), the protocols can be the following:

- Local Protocol (e.g. `/srv/git/project.git`)
- HTTP / HTTPS (`https://example.com/gitproject.git`)
- The SSH protocol (`ssh://[user@]server/project.git`)
- The Git protocol (`git://example.com/project.git`)

The `git@server:project.git` shorthand is handled by the `lib/git/remote.lua` parser as if it were a shorthand for the Git protocol, but is in fact a shorthand for the SSH protocol (see the page from the Git Book linked above).

While only used for parsing and not for communicating, this leads for a few issues, especially for self-hosted GitLab installs where you can have urls like "git@gitlab.priv:group/project/repo.git" which weren't correctly handled (noticed as a result of https://github.com/NeogitOrg/neogit/issues/1259#issuecomment-2079268027)

### Neovim version

NVIM v0.10.0-dev-2883+g2c5fd3f96-Homebrew
Build type: Release
LuaJIT 2.1.1710088188

MacOS 14.4.1

### Steps to reproduce

Parse url "git@gitlab.priv:group/project/repo.git". Parse url "ssh://git@gitlab.priv/group/project/repo.git"

### Expected behavior

Both should parse correctly and identically. repo should be `repo.git`, owner should be  `group/project`

### Actual behavior

Both parse differently.

franklouwers:
Per the Git Book (https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols), `git@remote:repo.git` is a shorthand for `ssh://git@remote/repo.git`. This handles this by explicitly parsing as ssh://... in the `git.remote.parse(url)` funciton.

Closes #1266

### DIFF
--- a/lua/neogit/lib/git/remote.lua
+++ b/lua/neogit/lib/git/remote.lua
@@ -72,35 +72,36 @@ end
 function M.parse(url)
   local protocol, user, host, port, path, repository, owner
 
-  for _, v in pairs { "git", "https", "http", "ssh" } do
-    if url:sub(1, #v) == v then
+  for _, v in pairs { "https", "http", "ssh", "git" } do
+    if url:sub(1, #v + 3) == (v .. "://") then
       protocol = v
       break
     end
   end
 
+  if protocol == nil then
+    -- handle case where url is in the user@ form by translating it to ssh://user@
+    -- note that this will fail for complex structures, but these would fail now as well.
+    url = url:gsub("^(.*@[^:]+):(.+)$", "ssh://%1/%2")
+    protocol = "ssh"
+  end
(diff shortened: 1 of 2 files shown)


CKolkey:
Great work, thanks man ;)

PR #1276, opened 2024-05-05
Fix error params in process.lua

davidmh:
Instead of concatenating it's passing the inspected object as a second parameter.

That's throwing an error when files are staged or branches switched from outside Neogit.

### DIFF
--- a/lua/neogit/process.lua
+++ b/lua/neogit/process.lua
@@ -375,7 +375,7 @@ function Process:spawn(cb)
   })
 
   if job <= 0 then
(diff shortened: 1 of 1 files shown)


CKolkey:
good catch, though I can't recreate it by staging or switching outside neovim. Still, the second argument should be the level, not another string.

davidmh:
Fwiw, I've seen it happen in three scenarios:

- Staging the file using vim-fugitive's `:GWrite`
- Cherry-picking from within a neovim terminal
- Switching into remote branches

PR #1277, opened 2024-05-05
Added config value for relative line numbers

Ziyan-Imran:
The original source code only has one optional value for `disable_line_numbers`. This value initially controls both regular line numbers and relative line numbers if toggled `true/false`. 

I have added an additional config value of the name `disable_relative_line_numbers` with its default value set to `false` which will only control the relatively line number option. 

Each value can then be used independently to give the end-user's more flexibility.


CKolkey:
Looks good, thanks :)

Issue #1256, opened 2024-04-18
NeogitStatus sometimes messes up filename, cant toggle file open


mangelozzi:
### Description

In the Neogit status it shows files name like:
```
Modified       s -> Gateway/Gateway.NextClient/src/app/app-route-strategy.t
```
Instead of (this is from the CLI):
```
modified:   Gateway.NextClient/src/app/app-route-strategy.ts
```

Because it has the wrong filenames, it can't toggle it open

### Neovim version

NVIM v0.10.0-dev-1224+gdfa8b582a
Build type: RelWithDebInfo
LuaJIT 2.1.1695653777
Run "nvim -V1 -v" for more info

### Operating system and version

Ubuntu 22.04

### Steps to reproduce

Not sure, I tried to create blank report, with the same dir structure, relative to the current PWD, but it did not do it then. I think the file has to be modified and renamed.

### Expected behavior

The correct file name to show, if I exit neovim, and perform a git status at the command line:
```
$ gs
On branch feature/prod-det8
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        renamed:    Gateway.NextClient/src/app/shopping/shopping.route-strategy.ts -> Gateway.NextClient/src/app/app-route-strategy.ts
        modified:   Gateway.NextClient/src/app/app.module.ts
        renamed:    Gateway.NextClient/src/shared/utils/reuse-route.ts -> Gateway.NextClient/src/shared/utils/route-snapshot-tools.ts

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   Gateway.NextClient/src/app/app-route-strategy.ts
        modified:   Gateway.NextClient/src/app/shopping/items/components/product-page/product-page.component.ts
        modified:   Gateway.NextClient/src/modules/cart-preview/cart-preview.component.html
        modified:   Gateway.NextClient/src/modules/cart-preview/cart-preview.component.ts
        modified:   Gateway.NextClient/src/modules/cart-preview/shared.cart-preview.module.ts
        modified:   Gateway.NextClient/src/modules/filtering/inventory-filters/inventory-filters.component.ts
        modified:   Gateway.NextClient/src/styles.scss

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        Gateway.NextClient/extra-webpack.config.js
        Gateway.NextClient/scripts/
[...]

Issue #954, opened 2023-11-18
Failed to detect git repo under windows


ray-x:
### Description

The issue I see now is different. It may releated to
https://github.com/NeogitOrg/neogit/issues/469
But after updating the plugin the error message is slightly different. ATM,  I am promoted I need to Initialize a repository in my working folder where .git exists.
I never see this message on Mac/Linux

### Neovim version

It existed in both 0.9.4 release and nightly

### Operating system and version

Windows + Powershell

### Steps to reproduce

Open a file inside a git repo with neovim and start Neogit. The neogit is complaining it is not a git repo. Seems somewhere in the code interpreting the folder to C:/xxx/xxx while it should be C:\xxx\xxx under windows.

### Expected behavior

Neogit should recognize current working folder is a git repo

### Actual behavior

Git repo not recognized. 

### Minimal config

```Lua
-- NOTE: See the end of this file if you are reporting an issue, etc. Ignore all the "scary" functions up top, those are
-- used for setup and other operations.
local M = {}

local base_root_path = vim.fn.fnamemodify(debug.getinfo(1, "S").source:sub(2), ":p:h") .. "/.min"
function M.root(path)
  return base_root_path .. "/" .. (path or "")
end

function M.load_plugin(plugin_name, plugin_url)
  local package_root = M.root("plugins/")
  local install_destination = package_root .. plugin_name
  vim.opt.runtimepath:append(install_destination)

  if not vim.loop.fs_stat(package_root) then
    vim.fn.mkdir(package_root, "p")
  end

  if not vim.loop.fs_stat(install_destination) then
    print(string.format("> Downloading plugin '%s' to '%s'", plugin_name, install_destination))
    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
    if vim.v.shell_error > 0 then
      error(string.format("> Failed to clone plugin: '%s' in '%s'!", plugin_name, install_destination),
        vim.log.levels.ERROR)
    end
  end
end

---@alias PluginName string The plugin name, will be used as part of the git clone destination
---@alias PluginUrl string The git url at which a plugin is located, can be a path. See https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols for details
---@alias MinPlugins table<PluginName, PluginUrl>

---Do the initial setup. Downloads plugins, ensures the minimal init does not pollute the filesystem by keeping
[...]

PR #1281, opened 2024-05-06
Use plenary.job to fix .git detection on windows, add regex fix for renamed files 

cjgibbons:
Multiple fixes for issues are contained in this PR:

- This PR fixes Issue #954 by changing:
   - the call to `vim.fn.system()` with `plenary.job`, and then testing the return value after the job finishes.
   - the call to `git_root_of_cwd()` with `plenary.job`, testing the return value and if in error defaults to `vim.fn.getcwd()` otherwise returns the directory from `git rev-parse --show-toplevel`

   On Windows, the `vim.fn.system()` was returning 127, instead of the expected 0, resulting in `false` whenever `is_inside_worktree()` is invoked. In turn, this always means the prompt for initializing the git repos is displayed for every Neogit operation. Plenary.job produces a 0 on success.

   Similarly, the call to `git_root_of_cwd()` kept returning the empty string "" (due to the non-zero success values) so `vim.fn.getcwd()` was always being called. The above change reduces the abstraction by making `opts.cwd`  have either the git toplevel directory or the current working directory whenever `git_root_of_cwd()` is invoked.

- This PR also fixes Issue #1256:
   - By matching on the beginning of the line and looking for `2 R` only then the regex matches the case for renamed files AND renamed and staged files. Since the file paths are matched properly they can be edited normally.

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -83,11 +83,11 @@ function M.open(opts)
   if not cli.is_inside_worktree(opts.cwd) then
     if
       input.get_confirmation(
-        string.format("Initialize repository in %s?", opts.cwd or vim.fn.getcwd()),
+        string.format("Initialize repository in %s?", opts.cwd),
         { values = { "&Yes", "&No" }, default = 2 }
       )
     then
-      lib.git.init.create(opts.cwd or vim.fn.getcwd(), true)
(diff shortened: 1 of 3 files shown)


CKolkey:
Good stuff! I don't have a windows installation, so I appreciate the assistance in ensuring a smooth experience with the platform :)

PR #1282, opened 2024-05-07
docs: update readme set verify_commit to default

nick22985:
Updates the default config to include the suppression of the stdout message.

https://github.com/NeogitOrg/neogit/blob/d51199a3552571a02b6d3d88b95acc4102ef2fd2/lua/neogit/config.lua#L222


CKolkey:
Thanks!

PR #1283, opened 2024-05-10
Fix the ability to open a files from within neogit when cwd is set to a subdirectory or external location, fixes Issue #1140

cjgibbons:
This PR fixes the bug in Issue #1140.

The problem stems from the default directory being returned instead of the detected `.git`'s parent directory. This issue existed prior to PR #1281, but is now properly fixed in this PR.

Specifically, the text of plenary.job's `job_output` is in a table element retrieved by calling `result()` against the table.

With this PR the proper directory top-level is returned wherever `cwd` is defined, including sub-directories, and it also fixes the behavior with `cwd=%:p:h` so that opening files in the subdirectory returns the correct path and one can still open files above the defined `cwd` from within Neogit.


CKolkey:
Thanks for the follow up :)

PR #1286, opened 2024-05-13
Add d key in commit view for diffview integation

ptn:
I didn't see any tests regarding the diffview integration, that's why this commit doesn't add any.


CKolkey:
Ahh, thanks for this. Didn't notice it was missing. But, it shouldn't go to the diffview integration directly, but the diff popup. I took care of this on the `nightly` branch, so no worries. Thanks for the heads up :)

PR #1301, opened 2024-05-17
Allow opening the staged diff in a vsplit

### LINKED ISSUE #1293: Allow `show_staged_diff` to open in a vertical split
No.

I'd like to be able to configure `show_staged_diff` to open in a vertical split. At the moment it always opens the diff buffer at the bottom, but I'd love to see it to the right.

**Additional context**
I'd love to send you a PR if you help me by pointing out where the code should go.

ptn:
Fixes #1293

### DIFF
--- a/lua/neogit/config.lua
+++ b/lua/neogit/config.lua
@@ -69,9 +69,16 @@ end
 ---@class NeogitConfigPopup Popup window options
 ---@field kind WindowKind The type of window that should be opened
 
+---@alias StagedDiffSplitKind
+---| "split" Open in a split
+---| "vsplit" Open in a vertical split
+---| "split_above" Like :top split
+---| "auto" "vsplit" if window would have 80 cols, otherwise "split"
+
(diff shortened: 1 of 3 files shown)


lua/neogit/config.lua
```diff
@@ -69,9 +69,14 @@ end
 ---@class NeogitConfigPopup Popup window options
 ---@field kind WindowKind The type of window that should be opened
 
+---@alias StagedDiffSplitKind
+---| "split" Open in a split
+---| "vsplit" Open in a vertical split
```
CKolkey, line 74: This can also include `split_above` and `auto`. `tab` and `floating` would _work_, but I think the result would be poor.

CKolkey:
Thanks! Small addition to the type annotation, but otherwise looks good :)

lua/neogit/config.lua
ptn: Makes sense. Pushed.

CKolkey: Thanks!

PR #1318, opened 2024-05-23
Fix typo in highlight name

DNLHC:
In commit view diff deletion highlighting wasn't working due to a typo
The correct highlight name defined in `lua/neogit/lib/hl.lua` is `NeogitDiffDeletions`


CKolkey:
Good catch, thanks

PR #1328, opened 2024-05-25
feat: add `vsplit_left`

stevenxxiu:
When I used *Neogit* a few days ago, the diff view was always on the left and the commit message editor on the right. This appears to have switched places.

This adds a `vsplit_left` option to open a new window to the left. In the case of the diff view, this can be set in `opts.commit_editor.staged_diff_split_kind` to open the diff view to the left. This restores the previous behavior.


CKolkey:
Thanks!

PR #1343, opened 2024-05-30
Remove unused keybind

ptn:
Space is super common to use as leader, so this was shadowing very important keymaps.


ptn:
If you have plans to add that debug functionality, I'd suggest mapping it to something else other than Space.

CKolkey:
Absolutely right, that didn't need to be there :)

PR #1350, opened 2024-06-01
Refs view: pass ref details to branch popup actions

actionshrimp:
Hi! First of all, thanks so much for all your amazing work on neogit. I've been wanting to switch over to neovim for a while, but magit was the one thing that held me back, so when I came across neogit and how much functionality it already has, I was blown away and have been fully able to switch over ❤️!

This PR adds something I did really love from magit - adding a bit of contextual information about the selected ref to the normal-mode branch popup, to save a bit of typing and to allow you to see what you're selecting a bit more easily from the name picker vs just a commit sha. Currently it's passed for:

- checking out a branch
- create a branch from an existing ref (a name based on the existing ref name is used to prefill the new branch name which is handy when you're checking out a remote ref)
- deleting a branch

One thing I noticed while testing is the refs view doesn't seem to refresh on the first reopen, but is refreshed if you close it and re-open it again. However, I thought investigating that might start getting a bit involved for a first PR! Let me know if it makes sense to investigate separately though.

Demo of the behaviour here:
https://github.com/NeogitOrg/neogit/assets/286704/33f034b7-2635-42a4-bb86-5a0a891c3289


CKolkey:
Nice! I saw the issues with the CI tests and fixed that - you can merge/rebase master to grab those changes.

The refresh thing is because only the status buffer and log buffer have "rerender" logic. I think it would make sense to use the `watcher` class to trigger the buffer to rerender, depending on how cleanly the state can be passed in :)

actionshrimp:
Great, thanks. I have rebased and fixed the formatting issue. Thanks for the pointer about the refresh - I will have a look and try to fix in a separate PR!

CKolkey:
Great contribution - I never used Magit, so these subtle things (that really make it shine) sometimes are overlooked. Really appreciate any suggestions for how to polish this up, if you have any more :)

PR #1352, opened 2024-06-02
Contributor tweaks

actionshrimp:
A couple of minor tweaks:
- I couldn't get `logger.fmt_debug` to work - strangely it didn't error but it didn't seem to log anything either. I updated `CONTRIBUTING.md` to match how the rest of the code did it.
- The tests were giving me an error as I'm on macos, but have GNU mktemp installed (there are at least 5 of us! 😆) - this detects GNU mktemp via `--version` rather than based on OS.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,8 +60,9 @@ verbose logging.
 ```lua
 local logger = require("neogit.logger")
 
-logger.fmt_info("This is a log message: %d", 2)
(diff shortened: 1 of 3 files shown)


CKolkey:
Yeah, no idea why `fmt_*` doesn't work.. Never dug too deeply into that. It's a vendored version of what plenary ships with some minor tweaks. Not a huge deal to just pass a pre-formatted string in, so I haven't bothered. 

Thanks for the other fixups :)

PR #1353, opened 2024-06-02
Ref view: refresh

actionshrimp:
I had a go at the refreshing of the ref view. This works as-is, but I'm wondering if you had any thoughts on the approach. Some details:
- I originally added the refs to the `git.repo.state` that the existing `git.repo:refresh` updated already, and was using that in some of the earlier commits, but I realised I could just refresh the refs state directly in the watcher callback which seems simpler for now, although there may be some drawbacks here. I also don't have a lock or anything like the main repo state refresh does, do you think it's worth adding one? I'm not super familiar with how async/concurrent the various functions are in neogit/nvim (yet :))
- The branch popup from the ref view kept taking me back to the status view, because the default options have `refocus_status` in - I've overridden this to false from the branch popup, but I guess there might be some situations where this is undesireable? It's quite nice to stay on the refs view if you're e.g. mass-deleting a load of local branches or something.
- If you have multiple remotes, I noticed sometimes they would change order so I added a sort in the UI rendering code to keep things stable here.
- I noticed that the setup of the Watcher in the status buffer wasn't actually getting assigned to the instance state [here](https://github.com/NeogitOrg/neogit/compare/master...actionshrimp:ref-view-refresh?expand=1#diff-b47379ff487cddc334f1b85eb050891d6f3ce3090579c7866bf9066815dfc829R204), as `:start()` returned `nil`, so this should hopefully help the status buffer clean up its watcher a bit better. 

Happy to tweak anything or split stuff out into separate PRs if that makes more sense!

### DIFF
--- a/lua/neogit/buffers/refs_view/init.lua
+++ b/lua/neogit/buffers/refs_view/init.lua
@@ -4,6 +4,10 @@ local ui = require("neogit.buffers.refs_view.ui")
 local popups = require("neogit.popups")
 local status_maps = require("neogit.config").get_reversed_status_maps()
 local CommitViewBuffer = require("neogit.buffers.commit_view")
+local Watcher = require("neogit.watcher")
+local logger = require("neogit.logger")
+local a = require("plenary.async")
+local git = require("neogit.lib.git")
 
 --- @class RefsViewBuffer
 --- @field buffer Buffer
@@ -18,11 +22,14 @@ local M = {
 
 --- Creates a new RefsViewBuffer
 --- @return RefsViewBuffer
-function M.new(refs)
+function M.new(refs, root)
   local instance = {
     refs = refs,
+    root = root,
     head = "HEAD",
     buffer = nil,
+    watcher = nil,
+    last_refreshed = 0,
   }
 
   setmetatable(instance, { __index = M })
@@ -33,6 +40,11 @@ function M:close()
   if self.buffer then
     self.buffer:close()
     self.buffer = nil
+  end
+
+  if self.watcher then
+    logger.debug("[REFS] Stopping Watcher")
+    self.watcher:stop()
   end
 
   M.instance = nil
@@ -57,6 +69,11 @@ function M:open()
     filetype = "NeogitRefsView",
     kind = config.values.refs_view.kind,
     context_highlight = false,
+    on_detach = function()
+      if self.watcher then
+        self.watcher:stop()
+      end
+    end,
     mappings = {
       v = {
         [popups.mapping_for("CherryPickPopup")] = popups.open("cherry_pick", function(p)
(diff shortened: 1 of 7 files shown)


actionshrimp:
Ok cool 👍 I've been using it for a few days and I'm fairly happy with it, so I've unmarked it as a draft in case you're interested in merging it in the short term for that watcher fix. Happy to wait until your refresh changes are in and I can base this off that if you prefer though.

CKolkey:
Hey :) Just merged the performance improvements that touched the watcher. If you don't mind rebasing these changes, I'll have a look at this very soon. 

I was a little concerned about commenting out the logging from the watcher callback - it would be ideal to find out _why_ that's a problem, since I use that logging quite a bit when trying to reason about the concurrency 😅 

The last thing I changed was a bit of an inversion of how the repo/status buffer handled debouncing. With more buffers using the refresh callback, it didn't make sense to tie it to the status buffer specifically.

actionshrimp:
No prob! I've just rebased and cleaned up the commits a bit 👍 Seems to be working nicely again now.

CKolkey:
Hey :) Much appreciated. I think I need to refactor how the watcher/buffers work together - but that's a good thing, this branch is definitely how I'd like things to work, it just is a bit over-eager in how many refreshes it does 😅 

I'll handle it and get this merged soon :)

actionshrimp:
No problem at all. Yeah I spotted a couple of oddities with the refreshing, sometimes 'changed' entries on the status buffer appear twice etc. No rush on this PR or anything, definitely worth ironing out all the kinks first and I'm happy to rebase as much as necessary - it's gonna be badass!

One thing I did wonder was if it was worth pushing the watcher down a level so it directly updated the repo state as it changed, rather than having it at a per-buffer level (if that's actually possible with the way neovim works)? But not entirely sure if how that'd work is clear in my mind.

actionshrimp:
Also just to add if it helps at all, having the status buffer knowing about the watcher so it can cleanup (https://github.com/NeogitOrg/neogit/pull/1353/commits/57d7e44c6b7791d2d7f85df337acb83a8284912f) definitely seemed to help a bit vs when I was directly testing against mater.

CKolkey:
As a matter of fact, I had the same thought today:

- there should be one watcher per .git/ directory, not buffer. Buffers can register with a watcher.
- watcher should trigger the repo directly, with a callback that rerenders registered buffers

Oh, and git.refs.list_parsed() is -horridly- slow on my repo at work. Like, 700ms to refresh, compared to 20ms for the other refreshes. So, I need to either figure out how to handle that better, or... donno yet. But that needs some rethinking. I'll make a branch and push what I was toying around with in the morning 😀 

It did expose some dumb things that are going on elsewhere, though, so thats at least something haha

CKolkey:
Well, tomorrow turned into... now. But I think I got it in a nice place :)

https://github.com/NeogitOrg/neogit/pull/1386

actionshrimp:
Awesome! I wont be able to have a look at this for a while as I’m without a laptop for a few days, but I’ll take a look early next week. (No need to wait until then though if I’m holding things up.)

CKolkey:
I included all these commits in a different branch and have merged that to `master` :) 

Thanks again for the contribution!

PR #1365, opened 2024-06-07
feat: make head section foldable

DanilaMihailov:
I really like what you did with folds in status view. The fact that I can use default key binds to fold/unfold AND configure some folds to be folded by default is really helpful for me. I usually have a few dozen untracked files and now I just fold them by default. Thank you for that :)

Magit allows to collapse the HEAD section and it seems neat. So I put all HEAD components inside a Section. And added config option to fold them by defalt.

Screenshot 2024-06-08 at 00 49 05
Screenshot 2024-06-08 at 00 49 11

I am not sure about name for config, called it `config.status.HEAD_folded` for now


CKolkey:
Cool! I didn't know that. Could you commit the config file and add the new line to the readme? :)

DanilaMihailov:
Added field to README and doc. And added types in config.lua

CKolkey:
Awesome stuff - thanks again. And really glad to hear you like the native folding - that took like... two months... to implement 😅

Issue #1245, opened 2024-04-08
Neogit breaks on GPG commit signing password prompt


Strongleong:
### Description

When I have enabled GPG signing and I try to commit this happens: 
https://github.com/NeogitOrg/neogit/assets/17177071/6a8bbd8b-b9a7-446e-9a67-e4d9cef47b7b

No matter what I type, GPG rejects it, neovim behaves unpredictable and only <Ctrl+c> helps. But console now in broken state (typing `reset` helps)

### Neovim version

NVIM v0.9.5
Build type: Release
LuaJIT 2.1.0-beta3
Compilation: /usr/lib/ccache/bin/x86_64-pc-linux-gnu-gcc  -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wvla -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -fno-common -Wno-unused-result -Wimplicit-fallthrough -fdiagnostics-color=always -fstack-protector-strong -DUNIT_TESTING -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -I/usr/include/luajit-2.1 -I/usr/include -I/usr/include -I/var/tmp/portage/app-editors/neovim-0.9.5/work/neovim-0.9.5_build/src/nvim/auto -I/var/tmp/portage/app-editors/neovim-0.9.5/work/neovim-0.9.5_build/include -I/var/tmp/portage/app-editors/neovim-0.9.5/work/neovim-0.9.5_build/cmake.config -I/var/tmp/portage/app-editors/neovim-0.9.5/work/neovim-0.9.5/src -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include

   system vimrc file: "/etc/vim/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

### Operating system and version

Gentoo (kernel version 6.1.69)

### Steps to reproduce

1. Make sure you have GPG signing enabled with password protected key
2. Create git repository
3. `nvim -nu minima.lua`
4. Create file
5. `:Neogit`
6. Stash file
7. Commit

### Expected behavior

Neogit prompts for GPG password and creates signed commit. Nothing is broken 

### Actual behavior

GPG password prompt pops up, Neogit tries to "fight" it.
Broken console.
Commit is not created

### Minimal config

[...]

PR #1376, opened 2024-06-13
Updated GPG integration docs

dandyrow:
I was having difficulties getting GPG integration to work on my system as explained in #1245. The issue ended up being that my GPG conf files were placed in the incorrect directory so the config wasn't being picked up. I thought I should add the command I used to work out what directory I needed to use in a note within the docs to help people who may come across this issue in the future.


CKolkey:
Thanks for this! Always great to add the comment you wish had been there :)

Issue #767, opened 2023-08-19
Neogit no longer respects showUntrackedFiles


AThePeanut4:
### Description

Since #722 (specifically f7811cc2024a63e70d0dfcfccda73ac9482c9d6c), Neogit no longer respects `showUntrackedFiles` in `.git/config`. Previously, untracked files were not shown if `showUntrackedFiles = no`, but now they are always shown, irrespective of what this is set to.

Context:

I use git to manage my dotfiles, with the bare repository in the `~/.dotfiles` folder, and the worktree at `~`. In order to be able to use git from the command-line, I have a shell alias that wraps the git command to add the `--git-dir` argument.

I also have Neovim set up to export the `GIT_DIR` environment variable when I am editing my config files, which allows plugins like diffview, gitsigns and Neogit to work correctly. For Neogit specifically, this works great - I am able to stage, commit, push, etc without issue.

However, because the worktree is at `~`, I need to be able to ignore untracked files, since every non-configuration file in my home directory is, and should be, untracked. For the git CLI, setting `showUntrackedFiles = no` does the trick, and before f7811cc2024a63e70d0dfcfccda73ac9482c9d6c, this also worked in Neogit.

Additionally, untracked files are now listed in Neogit individually instead of as directories, which means that now tens of thousands of files are loaded and displayed as untracked, making the status buffer extraordinarily slow - both to load and to navigate.

I have tried to disable the "Untracked files" section using the `sections = { untracked = { hidden = false } }` config option, but this just hides the section from being displayed - the data is still being loaded. The status page still takes forever to load, where it was close to instant before.

I think either `showUntrackedFiles` should be respected, or hiding the section should actually stop the data from being loaded. I would prefer `showUntrackedFiles`, since it would make my neovim config simpler, but either will work. Perhaps even an analogous config option to `showUntrackedFiles`, where you can specify whether to show nothing (git's `no`), files and root directories (git's `normal` - git's default) or all files in all directories (git's `all` - Neogit's new default).

### Neovim version

NVIM v0.10.0-dev-882+gd7ae9ae3e5
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

### Operating system and version

Arch Linux

### Steps to reproduce

1. Set `status.showUntrackedFiles = no` in a git repository (`git config status.showUntrackedFiles no`)
2. Add untracked files
3. Open the Neogit status buffer in that repository

### Expected behavior

Untracked files should not be listed under "Untracked files".

### Actual behavior

[...]

PR #1382, opened 2024-06-17
feat: show all untracked files in status buffer

titaniumtraveler:
This enables showing **all** unstaged files in the status buffer
instead of just the directories containing them.

While this could theoretically be enabled by setting `git config status.showUntrackedFiles all`,
I think showing all the unstaged files would be a pretty good default for neogit, especially when staging files.

### DIFF
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -129,6 +129,7 @@ local configurations = {
     },
     options = {
       porcelain = "--porcelain",
+      untracked_files = "--untracked-files",
     },
   },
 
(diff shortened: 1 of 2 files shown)


CKolkey:
Hey! I appreciate the thought - however, in the past I've had complaints of not respecting the git config setting: https://github.com/NeogitOrg/neogit/issues/767

And I subsequently added documentation for how to enable this: https://github.com/NeogitOrg/neogit/blob/11dea1b67bb153cd5b4a2de348ceefa69508dfce/doc/neogit.txt#L1618-L1621

So, given that, I'm going to leave things as-is. I _personally_ agree with you, that it's nice to see things individually, but since not everyone agrees and this is configurable externally, I don't want to mess with people's workflows :)

Thanks!

titaniumtraveler:
Fair. Thanks for the answer.
Though maybe it would be nice to have some config to set it *specifically* for neogit.
While I like having all files shown for neogit, for the normal `git status` I prefer just having the short overview.

(I would have looked into doing that myself, but the code base is kind of intimidating for someone who is not that experienced with lua lol)

PR #1402, opened 2024-06-30
Disable setting statuscolumn for all buffers if "disable_signs" is set

### LINKED ISSUE #1400: Some neogit buffers break `statuscolumn`
### Description

I use a custom `statuscolumn` to nicely display folds which a lot of the time works perfectly with Neogit since it adds in folds for things. But I noticed some recent update has made a lot of buffers break the pretty folds and reverse my custom `statuscolumn` completely causing things to look ugly again.

### Neovim version

NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1713484068

Arch linux

### Steps to reproduce

1. `nvim -u minimal.lua`, start up the minimal configuration
2. `cd <some_git_dir>`, change directories to a git tracked directory
3. `:Neogit`, open neogit and see that the folds look nice (notice the `>` and `v` with a nice line in the fold area, ignore the signs added by neogit as I didn't disable them to show both the signs and the foldcolumn)
2024-06-29_11:29:30_screenshot
4. `:set statuscolumn?`, see that it is set to our custom statuscolumn
5. Select a commit to open the commitview
6. Scroll down and see that now we have ugly folds rather than our nice looking folds
2024-06-29_11:30:37_screenshot
7. `:set statuscolumn?`, see that it is not set to our custom statuscolumn

### Expected behavior

Neogit should not modify my statuscolumn at all. If they want to modify information that is shown in a buffer then it should change things like `set foldcolumn?`, `set number?`, etc.

That way I can get the look that I have configured

2024-06-29_11:31:18_screenshot

### Actual behavior

My custom `statuscol` is completely removed and is now just using the default columns from neovim.

### Minimal config

```Lua
local M = {}

function M.root(path)
end

	end

		vim.fn.system({
			"git",
			"clone",
			"--depth=1",
			plugin_url,
			install_destination,
		})
		if vim.v.shell_error > 0 then
			error(
				string.format("> Failed to clone plugin: '%s' in '%s'!", plugin_name, install_destination),
				vim.log.levels.ERROR
			)
		end
	end
end

function M.setup(plugins)
	vim.opt.packpath = {} -- Empty the package path so we use only the plugins specified

	-- Install required plugins
	if plugins ~= nil then
		end
	end

		callback = function()
			vim.fn.system({
				"rm",
				"-r",
				"-f",
				M.root("xdg"),
			})
		end,
	})
end

M.setup({
	statuscol = "https://github.com/luukvbaal/statuscol.nvim", -- add statuscolumn plugin
})

local builtin = require("statuscol.builtin")
require("statuscol").setup({ -- do some basic configuration of a statuscolumn to illustrate folds looking nice
	relculright = true,
	segments = {
		{ text = { builtin.foldfunc }, click = "v:lua.ScFa" },
		{ text = { "%s" }, click = "v:lua.ScSa" },
		{ text = { builtin.lnumfunc, " " }, click = "v:lua.ScLa" },
	},
})
vim.opt.fillchars = [[eob: ,fold: ,foldopen:,foldsep: ,foldclose:]] -- set fold chars
vim.opt.foldcolumn = "1" -- enable fold column to show pretty folds that we configured

```

CKolkey:
by user

Fixes #1400

### DIFF
--- a/lua/neogit/buffers/commit_select_view/init.lua
+++ b/lua/neogit/buffers/commit_select_view/init.lua
@@ -56,7 +56,7 @@ function M:open(action)
   self.buffer = Buffer.create {
     name = "NeogitCommitSelectView",
     filetype = "NeogitCommitSelectView",
-    status_column = "",
+    status_column = not config.values.disable_signs and "" or nil,
     kind = config.values.commit_select_view.kind,
     header = self.header or "Select a commit with <cr>, or <esc> to abort",
     mappings = {
--- a/lua/neogit/buffers/commit_view/init.lua
+++ b/lua/neogit/buffers/commit_view/init.lua
@@ -128,7 +128,7 @@ function M:open(kind)
     name = "NeogitCommitView",
     filetype = "NeogitCommitView",
     kind = kind,
-    status_column = "",
+    status_column = not config.values.disable_signs and "" or nil,
     context_highlight = not config.values.disable_context_highlighting,
     mappings = {
       n = {
--- a/lua/neogit/buffers/diff/init.lua
+++ b/lua/neogit/buffers/diff/init.lua
@@ -54,7 +54,7 @@ function M:open()
   self.buffer = Buffer.create {
     name = "NeogitDiffView",
     filetype = "NeogitDiffView",
-    status_column = "",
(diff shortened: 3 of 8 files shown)


CKolkey:
Fixes #1400

PR #1404, opened 2024-06-30
fix(help): correctly handle if `keymap` is `"<nop>"`

mehalter:
Currently the help pop up is broken because keymaps can be equal to `"<nop>"`. This correctly verifies that the keymap is a table before adding it so that it skips `nop`s.

Even with a completely base, minimal installation, running `?` doesn't work at all. This fixes that

### DIFF
--- a/lua/neogit/popups/help/actions.lua
+++ b/lua/neogit/popups/help/actions.lua
@@ -16,7 +16,7 @@ local function present(commands)
     end
 
     local keymap = status_mappings[cmd] or popup_mappings[cmd]
-    if keymap and #keymap > 0 then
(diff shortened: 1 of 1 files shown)


mehalter:
This issue looks to originate from: https://github.com/NeogitOrg/neogit/commit/43ac899fd0b8550241f189f63317bd20f97f3098

Miyelsh:
Nice! I was chasing down this bug as well before seeing that you already made a fix.

Miyelsh:
With your change, the help menu opens but some keybinds are not shown. With default neogit settings.

image

CKolkey:
Thanks!

PR #1405, opened 2024-06-30
Add support for `git stash --keep-index` and `git stash --staged`

ej-shafran:
- Add staged-only stash
- Add stash while keeping index
- Remove unused function


ej-shafran:
I really enjoy these features in my day-to-day git usage, and was missing them in Neogit (which is awesome, it should be mentioned!)

Two things that might be weird with this PR:

1. I'm not entirely sure I understood the purpose of the `"NeogitStash"` event/autocmd...? I added it to the two new functions because it seemed to make the most sense to me, but I might be wrong...
2. As I noted in my comment, there's a really weird git error that happens (at least on my machine) when I run `git --literal-pathspecs stash --keep-index`, which meant I just passed the results of `git ls-files` as the pathspecs for the `git stash --keep-index` command. If there's some sort of solution for this issue, I couldn't find anything about it online... If anyone knows a nicer way to do this stuff, that would be appreciated.

Thank you for your time and this wonderful plugin!

CKolkey:
Hey! Thanks for taking the time to do this. Tried it locally: works like a charm.

To answer your question about the autocmd, it's listened for here: https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/buffers/status/init.lua#L221

I've been moving towards an event-driven architecture, which is working pretty nicely so far :)

PR #1280, opened 2024-05-06
Added support for `git stash list`

AlphabetsAlphabets:
Added basic support for `git stash list`. Current functionality in the `:Neogit` screen.
1. `Z` then `l` will give a list of stashes.
2. Pressing `ENTER` on any of the stashes will bring up a diff for that stash.

PR #1406, opened 2024-06-30
Add support for `git stash list` (replaces #1280)

ej-shafran:
Took over the pull request over at #1280 - turns out all it needed was a merge and format, if I read the comments on that PR correctly :)

All of the code itself is the work of @AlphabetsAlphabets


CKolkey:
Thank you @AlphabetsAlphabets and @ej-shafran for pushing this forward :)

Issue #841, opened 2023-09-23
Stage pop up always jumps to top


mangelozzi:
### Description

Great plugin, but I often regret updating it. I can't use the stage popup anymore, because I have a lot of changes, and every time I press tab, its scrolls back to the top of the screen, super disorientating.

### Neovim version

NVIM v0.10.0-dev-fcfc87c
Build type: RelWithDebInfo
LuaJIT 2.1.1695266395
Run "nvim -V1 -v" for more info

### Operating system and version

Ubuntu 22.04

### Steps to reproduce

Open the commit pop up,
Scroll down to a diff, press tab to open, the cusor and screen will scroll to the top and you will loose your place.
Also happens is you press `s` to stage a section, scrolls back to the top of the screen

### Expected behavior

Cursor and screen position to stay where it was

### Actual behavior

Jumps back to top 

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "NeogitOrg/neogit",
        requires = {
          { "nvim-lua/plenary.nvim" },
          { "sindrets/diffview.nvim" },
        },
        config = function()
          print("loaded neogit")
          require("neogit").setup()
        end,
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing neogit and dependencies.")
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
```

CKolkey:
Yeah, theres a double-refresh bug I havent had time to sort out yet. Try turning off the file watcher, that should do the trick 😘

CKolkey:
I've been thinking about this: It's very annoying that the cursor jumps, so until I sort out the double-render thing, I'm disabling the filewatcher by default (you can opt in).

Thank you for the feedback :) I'm closing this issue, but please feel free to reopen it if you continue to experience this bug.

mangelozzi:
[...]

PR #1425, opened 2024-07-18
Fix restoring scroll when unfolding in status view

AlexanderArvidsson:
- [x] Based off `nightly` branch

The current `Ui:update` function only restores the cursor, but not the "scroll", i.e. the top line, as noted in https://github.com/NeogitOrg/neogit/issues/841#issuecomment-2237724913.

This PR addresses that issue by first restoring the proper top line, taking `scrolloff` and folds into account, then finally restoring cursor. This prevents jumping when folding in the status view.

Here's a video demonstrating before and after:
https://github.com/user-attachments/assets/cefb50a2-107c-4718-a419-417fe1b19aa0
(I'm jumping around a bit in the after one with C-D and C-U, but rest assured the folding restores the scroll like it should)

Now, an alternative to this solution would be to look into a way to update the Ui _without resetting cursor position_ completely. That could perhaps be done by updating the buffer with the new lines instead of first clearing it, and might even be more performant. But I don't know enough about the internals to make this call.

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -72,8 +72,6 @@ function M.setup(opts)
   M.config = config
   M.notification = require("neogit.lib.notification")
 
-  M.notification.info("You are using the `nightly` branch for neogit.\nThis branch will no longer be receiving updates - use `master` to stay up-to-date.")
-
   config.setup(opts)
   hl.setup()
   signs.setup()
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -72,6 +72,8 @@ function M.setup(opts)
   M.config = config
   M.notification = require("neogit.lib.notification")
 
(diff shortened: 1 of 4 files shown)


AlexanderArvidsson:
I realize that the `nightly` branch is outdated now through the warning that is sent when running neogit. I based it off `nightly` branch because the contribution guideline hadn't been updated to remove the note that says "Until neovim 0.10 is released, please base any changes on the nightly branch." so I assumed the switch hadn't been made now since neovim 0.10 is out.

I can rebase it off master, but I think it's fine as is because there's no conflicts and the master branch still has the same problem (which this fixes).

CKolkey:
Heh, it _does_ say "until nvim 0.10 is released", but no worries - I removed that. Thanks for the reminder. 

I'll have a look at what you've got cooking here.

CKolkey:
You know what, there's no conflicts or issues here, so I'll just rebase/merge this to `master` as-is. Great eye for detail!

PR #1431, opened 2024-07-23
Add cherry-pick options/switches

ej-shafran:
Not sure why these options were in a `TODO` comment; do we maybe need to keep it as a comment?

### DIFF
--- a/lua/neogit/lib/git/cherry_pick.lua
+++ b/lua/neogit/lib/git/cherry_pick.lua
@@ -1,6 +1,7 @@
 local git = require("neogit.lib.git")
 local notification = require("neogit.lib.notification")
 local util = require("neogit.lib.util")
+local client = require("neogit.client")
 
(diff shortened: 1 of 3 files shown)


CKolkey:
I think I marked those as a TODO quite a while ago and never revisited them. Have you tested that they all work as expected?

ej-shafran:
So far I've only tested `mainline`, which seems to work just fine. I'll check the others.

Could it be that these were marked as TODO because `switch_if` and `option_if` did not exist? They _were_ written as `switch` and `option` in the original comment...

CKolkey:
Yeah, could have been. I also didn't understand a lot of the internals of this app yet, especially how all the async stuff worked. I'll go over what you've got and see what needs fixing up :)

CKolkey:
Really appreciate you picking this up! It's easy to forget to uh.. do.. the todos 😅

PR #1443, opened 2024-08-10
Add ability to open splits above/below all windows

quolpr:
Before with `split_above`:
image
Now with `split_above_all`:
image


CKolkey:
Hey, nice work! I'll clean up the style on master.

PR #1444, opened 2024-08-12
Fix: actually show/focus buffers when existing window

asmodeus812:
The show method was not working correctly for existing windows, while for buffers that are not already visible it would actually make sure to show the buffer by creating the layout according to the `kind` value. Fixes: https://github.com/NeogitOrg/neogit/issues/1373

### DIFF
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -256,6 +256,7 @@ function Buffer:show()
 
   -- Already visible
   if #windows > 0 then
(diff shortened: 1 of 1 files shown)


asmodeus812:
@CKolkey this seems to fix the problem metnioned above, seems like a reasonable solution, however would like to make sure we are not going to break anything else, without realising, preliminary testing shows no adverse effects.

CKolkey:
Thanks for this - lets see if anyone can find an issue :)

PR #1463, opened 2024-08-29
Handle target function exceptions in throttle_by_id

ioreshnikov:
Hey!

I've noticed that if a throttled function fails it's considered to be running and any subsequent calls would be cancelled. This PR fixes that by running the target function in protected mode.

### DIFF
--- a/lua/neogit/lib/util.lua
+++ b/lua/neogit/lib/util.lua
@@ -574,7 +574,7 @@ function M.throttle_by_id(fn, schedule)
     while scheduled[id] do
       scheduled[id] = nil
       running[id] = true
-      fn(id, ...)
+      pcall(fn, id, ...)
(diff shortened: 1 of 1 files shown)


ioreshnikov:
And now a bit of context: it's tangentially related to the issue I have with status buffer refresh.

I am working on a giant monorepo and for some reason the very first status refresh fails at some point after the redraw. Any subsequent status refresh is cancelled since the refresh callback is still considered to be running.

I am trying to investigate the original issue, yet the locking behaviour is preventing me from rerunning the failing code.

But aside from me personal issues, I think it's a reasonable behaviour, that the throttling mechanism resets the running flag even if the function fails.

CKolkey:
Seems reasonable to me! Thanks for the patch

PR #1465, opened 2024-08-29
Improve window splitting options for `kind`

afonsocarlos:
## Changes

Add two new options for `kind` which are `split_top` and `split_bottom`. Those allow the window to be split as the top/bottom most window (That'd be the same as `:top split` and `:bottom split`  respectively). 
Also, correct the actual behavior of `split_above` in the docs/comments.

### With `kind=split` as usual:
image

### With the new `kind=split_bottom`:
image

Note: `split_above` and `split_top` do the same thing but at the top, of course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,9 @@ neogit.setup {
     -- Accepted values:
     -- "split" to show the staged diff below the commit editor
     -- "vsplit" to show it to the right
-    -- "split_above" Like :top split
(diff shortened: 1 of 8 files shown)


CKolkey:
Hey! I just merged https://github.com/NeogitOrg/neogit/pull/1443 which implements pretty much the same feature.

Great minds think alike, I guess :) Thanks all the same.

PR #1466, opened 2024-08-30
Make revert better mirror magit behaviour

ktprograms:
When reverting a range of commits, simply pass all commits to git-revert, and let git handle it.

This keeps the standard git-revert commit message, instead of Neogit's custom commit message.

---

According to https://www.youtube.com/watch?v=7D4cV0aodGw&t=188s, this is Magit's behaviour.

---

Should `git.revert.commits` be inlined into `M.changes`, as that's the only place it's used (because of this change)?


CKolkey:
Hmm. I rather like the "enhanced" functionality for revert beyond what git/magit implement. Why would the default behavior be preferred?

Issue #1479, opened 2024-09-16
DISABLING spell checker


Kyren223:
I find the spellchecker introduced recently really annoying and it's impossible to disable.
There should be a config option to disable it.
And in my opinion, it should be opt-in, but I wouldn't mind opt-out either.

rami3l:
I second this. Personally I use `cspell` for spell checking already, so I've disabled the native spellchecker. However, `neogit` seems to have ignored the global setting...

CKolkey:
Fair - I added an option to toggle it. Feel free to make a PR if you think this could hook into nvim better. I have no time right now to do much more than this, I'm afraid.

PR #1480, opened 2024-09-17
Allow spell check to be user configurable

CKolkey:


CKolkey:
Closes https://github.com/NeogitOrg/neogit/issues/1479

PR #1506, opened 2024-10-14
fix: don't use config.highlight on autocommands

### LINKED ISSUE #1505: Neovim crashes after last update
### Description

I get an error when I start neovim. You can get it using minimal.lua. 

### Neovim version

NVIM v0.10.2
Build type: Release
LuaJIT 2.1.1727870382

macOS 15.0.1

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. watch the red text

### Expected behavior

_No response_

### Actual behavior

The error:
```
Error detected while processing /Users/strash/.dotfiles/nvim/.config/nvim/init.lua:                                                                                                             
E5113: Error while calling lua chunk: /Users/strash/.config/nvim/lua/settings.lua:22: /Users/strash/.dotfiles/nvim/.config/nvim/init.lua..ColorScheme Autocommands for "*": Vim(append):Error ex
ecuting lua callback: vim/shared.lua:0: after the second argument: expected table, got nil                                                                                                      
stack traceback:                                                                                                                                                                                
        [C]: in function 'error'                                                                                                                                                                
        vim/shared.lua: in function 'validate'                                                                                                                                                  
        vim/shared.lua: in function 'make_palette'                                                                                                                                              
        ...rash/.local/share/nvim/lazy/neogit/lua/neogit/lib/hl.lua:149: in function <...rash/.local/share/nvim/lazy/neogit/lua/neogit/lib/hl.lua:148>                                          
        [C]: in function 'colo'                                                                                                                                                                 
        /Users/strash/.config/nvim/lua/settings.lua:22: in main chunk                                                                                                                           
        [C]: in function 'require'                                                                                                                                                              
        /Users/strash/.dotfiles/nvim/.config/nvim/init.lua:8: in main chunk                                                                                                                     
stack traceback:                                                                                                                                                                                
        [C]: in function 'colo'                                                                                                                                                                 
        /Users/strash/.config/nvim/lua/settings.lua:22: in main chunk                                                                                                                           
        [C]: in function 'require'                                                                                                                                                              
        /Users/strash/.dotfiles/nvim/.config/nvim/init.lua:8: in main chunk
```

### Minimal config

```Lua
local M = {}

function M.root(path)
end

	end

		vim.fn.system({
			"git",
			"clone",
			"--depth=1",
			plugin_url,
			install_destination,
		})
		if vim.v.shell_error > 0 then
				vim.log.levels.ERROR)
		end
	end
end

function M.setup(plugins)
	vim.opt.packpath = {}                   -- Empty the package path so we use only the plugins specified

	-- Install required plugins
	if plugins ~= nil then
		end
	end

		callback = function()
			vim.fn.system({
				"rm",
				"-r",
				"-f",
				M.root("xdg")
			})
		end
	})
end

M.setup({
	-- telescope = "https://github.com/nvim-telescope/telescope.nvim",
	-- diffview = "https://github.com/sindrets/diffview.nvim",
	lavish = "https://github.com/ferdinandrau/lavish.nvim",
})
require("neogit").setup({
	kind = "split",
	disable_commit_confirmation = true,
	disable_insert_on_commit = true,
}) -- For instance, setup Neogit
require("lavish").setup({
	style = {
		italic_comments = true,
	},
})

local opt, cmd = vim.opt, vim.cmd
opt.background = "dark"
-- also checked with other colorschemes, same
local colo = "lavish"
cmd.colorscheme(colo)
```

Ascenio:
Whenever colorscheme changes the `setup` function of `hl.lua` is called, which then tries to create a palette from the passed config (also considering the defaults). 

The problem was that when `ColorScheme` autocommand triggers this change, the field `config.highlight` is missing. So in this case we should ignore it.  

Fixes #1505.

### DIFF
--- a/lua/neogit/lib/hl.lua
+++ b/lua/neogit/lib/hl.lua
@@ -128,7 +128,7 @@ local function make_palette(config)
     underline  = true,
   }
 
-  return vim.tbl_extend("keep", config.highlight, default)
+  return vim.tbl_extend("keep", config.highlight or {}, default)
 end
 -- stylua: ignore end
 


Ascenio:
Not needed anymore. Reverted on eda716c44d3b5a424ea8604b10756c7fd8bed93a

CKolkey:
I'll unrevert that later and incorporate this - definitely not time wasted.

PR #1525, opened 2024-10-27
fix: windowkind types

NickyMeuleman:
While configuring this plugin, my LSP kept saying the kind I chose was an error because it was not in the `WindowKind` alias

This PR adds types to the WindowKind alias
referenced: https://github.com/NeogitOrg/neogit/blob/master/lua/neogit.lua#L231
and the [usage part of the readme](https://github.com/NeogitOrg/neogit/tree/master?tab=readme-ov-file#usage) that states which kinds are valid.

### DIFF
--- a/lua/neogit/config.lua
+++ b/lua/neogit/config.lua
@@ -79,10 +79,16 @@ function M.get_user_mappings(set)
 end
 
 ---@alias WindowKind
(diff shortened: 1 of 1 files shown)


CKolkey:
Nice, thank you :)

PR #1529, opened 2024-10-29
Merge

ergotu:


CKolkey:
I appreciate the thought, but I made this change irrelevant a few days ago - the issue was that _every_ line was being buffered in a lua table, which got real big and would crash the process.

Updated it to not do that anymore and just write straight to the text buffer.

Regardless, thanks!

PR #1531, opened 2024-10-30
Self-adaping background color for custom color palette

SheffeyG:
The self adaping bg color implemetation is great, but that doesn't work for custom color palette has only main colors, so I fetch these custom colors before the bg colors are calculated.

### DIFF
--- a/lua/neogit/lib/hl.lua
+++ b/lua/neogit/lib/hl.lua
@@ -89,13 +89,13 @@ end
 local function make_palette(config)
   local bg        = Color.from_hex(get_bg("Normal") or (vim.o.bg == "dark" and "#22252A" or "#eeeeee"))
   local fg        = Color.from_hex((vim.o.bg == "dark" and "#fcfcfc" or "#22252A"))
-  local red       = Color.from_hex(get_fg("Error") or "#E06C75")
-  local orange    = Color.from_hex(get_fg("SpecialChar") or "#ffcb6b")
-  local yellow    = Color.from_hex(get_fg("PreProc") or "#FFE082")
-  local green     = Color.from_hex(get_fg("String") or "#C3E88D")
-  local cyan      = Color.from_hex(get_fg("Operator") or "#89ddff")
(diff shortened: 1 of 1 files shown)


fnune (commented):
My 2c: I would say if this is necessary then declare new highlights and go `get_bg("NeogitPaletteRed")` instead of adding things to the config. The defaults could be moved there, including `get_fg("Error")`.

SheffeyG:
@fnune Thanks, but all it did is just make sure the priority is `custom config > get_fg > hard code`, I didnt add anything to the config, just read them.
And the reason I make this is when the custom highlights has only main colors,
```lua
        highlight = {
            red       = "#E06C75",
            orange    = "#ffcb6b",
            yellow    = "#FFE082",
            green     = "#C3E88D",
            cyan      = "#89ddff",
            blue      = "#82AAFF",
            purple    = "#C792EA",
        },
```
the rest background colors are caculated by the `get_fg()` values or the hard code ones instead of the custom highlight.
> The defaults could be moved there, including get_fg("Error").

yep, the `defaults` table would base on priority `custom > get_fg > default` now, but I don't know why `get_fg("Error")` would be changed.

One last, we do can make a whole new color table just like `default` if you all think that is better.

fnune:
I see, `highlight` is already a config option. I was looking for `highlights` and as nothing turned up I thought it wasn't there. MB!

SheffeyG:
yeah, the documents `highlight` part was just added few hours ago.

SheffeyG:
@CKolkey Hi, this commit will never change any colors or any thing whether users has no custom colors or users have their own custom color palettes as the last table merging:
https://github.com/NeogitOrg/neogit/blob/bae3412ea9ac31292e14e9090dde7cb118cd52c1/lua/neogit/lib/hl.lua#L131
 it only affect users who want to customize or use the default primary color but don't want to configure the entire color palette, as it's kind of complex(#1499).
If this commit has any unintended side effects or if something related is in progress, please let me know. Thanks.

CKolkey:
Hey, this is great! Thanks for the contribution.

PR #1543, opened 2024-11-05
Handle `incompatible`/`dependant` for options

zivarah:
Both of the `incompatible` and `dependant` concepts are only implemented for switches, not options. However, options can be incompatible or dependent on each other and switches as well.

Handle both concepts between all options and switches. Use the existing `toggle_switch`/`set_option` to do the disabling so that any recursive dependencies are resolved as well.


CKolkey:
Fantastic - no notes.

PR #1546, opened 2024-11-06
Clean up annotations on APIs that can be made public and document the ability to create custom popups

zivarah:
As noted in https://github.com/NeogitOrg/neogit/discussions/1542, it could be very useful to use Neogit as a platform for building popups and actions for custom subcommands. Clean up some of the documentation around the builder APIs and mention in the docs the possibility of creating custom commands.

### DIFF
--- a/doc/neogit.txt
+++ b/doc/neogit.txt
@@ -1848,5 +1848,55 @@ The following keys, in normal mode, will act on the commit under the cursor:
 • `b`    Insert breakpoint
 • `<cr>` Open current commit in Commit Buffer
 
+==============================================================================
+Custom Popups                                             *neogit_custom_popups*
+
+You can leverage Neogit's infrastructure to create your own popups and
+actions. For example: >lua
+    local function my_action(popup)
+      -- You can use Neogit's git abstraction for many common operations
+      -- local git = require("neogit.lib.git")
+      local input = require("neogit.lib.input")
+      local user_input = input.get_user_input("User-specified free text for the action")
+      local cli_args = popup:get_arguments()
+      vim.notify(
+        "Hello from my custom action!\n"
+        .. "CLI args: `" .. table.concat(cli_args, " ") .. "`\n"
+        .. "User input: `" .. user_input .. "`")
+    end
+
+    function create_custom_popup()
+      local popup = require("neogit.lib.popup")
+      local p = popup
+        .builder()
+        :name("NeogitMyCustomPopup")
+        :switch("s", "my-switch", "My switch")
+        :option("o", "my-option", "default_value", "My option", { key_prefix = "-" })
+        :new_action_group("My actions")
+        :action("a", "Some action", my_action)
+        :build()
+
+      p:show()
+
+      return p
+    end
+
+  require("neogit")
+<
+
+Look at the builder APIs in `lua/neogit/lib/popup/builder.lua`, the built-in
+popups/actions in `lua/neogit/popups/*`, and the git APIs in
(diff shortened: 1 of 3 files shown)


doc/neogit.txt
```diff
@@ -1848,6 +1848,13 @@ The following keys, in normal mode, will act on the commit under the cursor:
 • `b`    Insert breakpoint
 • `<cr>` Open current commit in Commit Buffer
 
+==============================================================================
+Custom Popups                                             *neogit_custom_popups*
+
+You can leverage Neogit's infrastructure to create your own popups and
+actions. Look at the built-in popups within `lua/neogit/popups/*` and the
+builder APIs in `lua/neogit/lib/popup/builder.lua` to get started.
+
```
CKolkey, line 1857: I think it would be helpful to provide a "hello world" type example, something that:
- Shows what to `require()`
- Shows how to bind the popup to a key via `vim.keymap.set()`
- Shows the switch/option/config/action/group methods (+ conditional variants)
- Shows how to bind a simple function to an action

The example can be _completely_ contrived; don't bother making it do anything actually useful. Sound reasonable?

CKolkey:
Not going to lie, this is actually _way_ simpler than I had in mind - in a good way. I had this whole idea about replacing the builder pattern with a completely declarative approach, which would be a lot of work and, as this shows... entirely not needed. Thanks for that :)

Question for you: would making the `git.*` modules public be useful? Or a subset of them? Or leave it up to the user to get data via `vim.system()` or whatever they like. Definitely not something for this PR if so. Just wondering how much of the internal infrastructure would be valuable. 

Another aspect of this that would be good to show would be how to integrate a custom popup with the status buffer:
- How the selection gets passed into the popup constructor function
- How to trigger a refresh of the status buffer after running an action

I think those are somewhat unsolved by this (which is fine) - but again, might not be a big deal. What do you think?

doc/neogit.txt
zivarah: I've pushed an update that goes into a little more detail here. I wasn't sure how valuable trying to include _every_ one of the builder methods would be (and it would lead to a little more potential maintenance to keep things in sync). If you think that would be better I can add it, but I wonder if looking at the actual built-in popups would provide a more useful/realistic view of how to use them. Let me know!

zivarah:
> Not going to lie, this is actually _way_ simpler than I had in mind - in a good way. I had this whole idea about replacing the builder pattern with a completely declarative approach, which would be a lot of work and, as this shows... entirely not needed. Thanks for that :)

Oh interesting! Yeah, I figured that the existing mechanisms were fine as they are.

> Question for you: would making the `git.*` modules public be useful? Or a subset of them? Or leave it up to the user to get data via `vim.system()` or whatever they like. Definitely not something for this PR if so. Just wondering how much of the internal infrastructure would be valuable.

It seems that would be fine -- I haven't really used them but from a glance they seem easy enough to use. The documentation is a bit sparse for some of them, but I think that could be fleshed out in the future.

> Another aspect of this that would be good to show would be how to integrate a custom popup with the status buffer:
> 
> * How the selection gets passed into the popup constructor function
> * How to trigger a refresh of the status buffer after running an action
> 
> I think those are somewhat unsolved by this (which is fine) - but again, might not be a big deal. What do you think?

I'm less familiar with these -- would you mind tweaking what I have pushed currently to include what you're envisioning for that? Happy to include it!

CKolkey:
Really appreciate your initiative here - thank you!

PR #1548, opened 2024-11-07
fix: process.lua error when we have a bare repo in the worktree list

SjB:
When generating a list of worktrees using the 'git worktree list --porcelain' command, a bare worktree only has 2 lines, whereas all other worktrees have 3 lines. Since we don't need to switch to the bare repo, we can safely ignore it. The new listing function will look for a branch listing and only add that to the list of worktrees. 

```
worktree /tmp/ws-nvim/.repos/neogit.git
bare

worktree /tmp/ws-nvim/neogit-bare-worktree-listing-error
HEAD 2b770465b79dfe60898983ddfc897cdf08cf9d34
branch refs/heads/fix-bare-worktree-listing-error

worktree /tmp/ws-nvim/neogit-master
HEAD bae3412ea9ac31292e14e9090dde7cb118cd52c1
branch refs/heads/master
```

### DIFF
--- a/lua/neogit/lib/git/worktree.lua
+++ b/lua/neogit/lib/git/worktree.lua
@@ -48,20 +48,22 @@ function M.list(opts)
   local list = git.cli.worktree.list.args("--porcelain").call({ hidden = true }).stdout
 
   local worktrees = {}
-  for i = 1, #list, 3 do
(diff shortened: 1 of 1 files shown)


CKolkey:
Thanks!

PR #1551, opened 2024-11-07
Spinner WIP

CKolkey:
@fnune Whatcha think?

### DIFF
--- a/lua/neogit/buffers/stash_list_view/init.lua
+++ b/lua/neogit/buffers/stash_list_view/init.lua
@@ -42,6 +42,7 @@ function M:open()
     mappings = {
       v = {
         [popups.mapping_for("CherryPickPopup")] = function()
+          -- TODO: implement
           -- local stash = self.buffer.ui:get_commit_under_cursor()[1]
           -- if stash then
           --   local stash_item = util.find(self.stashes, function(s)
--- a/lua/neogit/spinner.lua
+++ b/lua/neogit/spinner.lua
@@ -35,7 +35,7 @@ function Spinner:start()
   if not self.timer then
     self.timer = vim.uv.new_timer()
     self.timer:start(
-      0,
+      250,
       self.interval,
       vim.schedule_wrap(function()
         self.count = self.count + 1
--- a/lua/neogit/spinner.lua
+++ b/lua/neogit/spinner.lua
@@ -4,6 +4,8 @@
 ---@field interval number
 ---@field pattern string[]
 ---@field timer uv_timer_t
+---@field start fun(self)
+---@field stop fun(self)
 local Spinner = {}
 Spinner.__index = Spinner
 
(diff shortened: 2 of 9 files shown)


lua/neogit/buffers/stash_list_view/init.lua
```diff
@@ -42,6 +42,7 @@ function M:open()
     mappings = {
       v = {
         [popups.mapping_for("CherryPickPopup")] = function()
+          -- TODO: implement
```
fnune, line 45: Nit: implement what? Running the spinner?

lua/neogit/spinner.lua
```diff
@@ -0,0 +1,67 @@
...
+function Spinner:start()
+  if not self.timer then
+    self.timer = vim.uv.new_timer()
+    self.timer:start(
+      250,
+      self.interval,
+      vim.schedule_wrap(function()
+        self.count = self.count + 1
+        local step = self.pattern[(self.count % #self.pattern) + 1]
+        vim.cmd(string.format("redraw | echomsg '%s %s'", step, self.text))
```
fnune, line 45: So `redraw | echomsg` is what allows the same notification to be updated?

I just tested with `noice.nvim` (I don't usually use it but it surfaces bad use of e.g. `vim.notify`) and it works nicely, I hadn't managed to do this myself:

image

lua/neogit/spinner.lua
```diff
@@ -0,0 +1,67 @@
...
+      "⠙",
+      "⠹",
+      "⠸",
+      "⠼",
+      "⠴",
+      "⠦",
+      "⠧",
+      "⠇",
+      "⠏",
+    },
```
fnune, line 30: Making this customizable would be perfect :) here are some ideas if you'd like inspiration https://github.com/gtramontina/spinner-frames.cr/blob/master/src/spinner-frames/charset.cr

fnune (commented):
This is awesome! I'm sorry I never got to it. It's been a busy few weeks.

lua/neogit/spinner.lua
CKolkey: Yeah, `redraw` is commonly used to prevent hitting the "hit enter" prompt, which would otherwise happen. Very nice coincidence that it works with noice!

lua/neogit/spinner.lua
CKolkey: sure :D

pinbraerts:
How to disable this? I use `set cmdheight=0`

Btw this (as well as any other fancy resource-consuming frontend feature) should be opt-in, not opt-out. I propose a config parameter to disable any animations whatsoever.

[Screencast from 11-08-2024 11:03:54 PM.webm](https://github.com/user-attachments/assets/84dc4948-f3f7-47f2-900b-890e65f7fe8b)

CKolkey:
Set `process_spinner = false` in your config

pinbraerts:
Thanks!

PR #1552, opened 2024-11-08
feat: Add auto_show_console_on = "output" | "error" setting

fnune:
This pairs very well with the new spinner feature: https://github.com/NeogitOrg/neogit/pull/1551

Allows users to set:

```lua
{ auto_show_console_on = "error" }
```

Note that this is only considered if `auto_show_console` is `true` (this is the default).

This results in Neogit showing a spinner while, e.g. pre-commit hooks are running, and only shows the console if the pre-commit hooks fail.

The default is `auto_show_console_on = "output"`, which matches the previous behavior.

This PR also stops Neogit from calling `notification.warn` on a bad exit code, because this overlaps with showing the console, and requires the user to press enter unnecessarily in order to get to the console output. The `notification.warn` call is preserved if `auto_show_console` is `false`.

### DIFF
--- a/doc/neogit.txt
+++ b/doc/neogit.txt
@@ -120,6 +120,9 @@ TODO: Detail what these do
     console_timeout = 2000,
     -- Automatically show console if a command takes more than console_timeout milliseconds
     auto_show_console = true,
+    -- If `auto_show_console` is enabled, specify "output" (default) to show
+    -- the console always, or "error" to auto-show the console only on error
+    auto_show_console_on = "output",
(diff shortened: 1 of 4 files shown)


fnune:
@CKolkey I'm no longer sure this is necessary. I think setting `console_timeout` to something higher (10s?) produces a similar effect?

fnune:
Actually it is necessary in cases like longer-lived `git commit` sessions (which almost always exceed 10 seconds).

CKolkey:
Nice :)

Issue #1554, opened 2024-11-10
Improve naming convention of configuration parameters for better readability


D00mch:
I'd like to suggest re-evaluating the naming convention for certain configuration parameters. Currently, the parameters use "disable" as a prefix, which increases cognitive load when interpreting their values. 

“disable_*=false“ is like a negation of negation (two operation to get the meaning).

Using an “enable_*“ prefix would make it more intuitive and easy to get.

Parameters:
```
disable_context_highlighting
disable_hint
disable_insert_on_commit 
disable_line_numbers
disable_signs
```

CKolkey:
Totally agreed - these are just kinda legacy from before I took over. I'm open to changing them, but won't have time for a while. I brought twins home from the hospital juuuuust a few hours ago 😅

So, if you wanna do it, be my guest. Otherwise... no worries, it'll be a minute though.

D00mch:
I will make a PR today

ktsujister:
This seems like it's related to #227 as well.

D00mch:
Updated the PR, merged conflicts

PR #1556, opened 2024-11-10
Breaking!: rename disable_* config properties to enable_*

D00mch:
The issue: https://github.com/NeogitOrg/neogit/issues/1554


CKolkey:
Seems very reasonable - I'd like to have a system to read the deprecated values and warn people to fix it. I know it's always super annoying when plugins change stuff like this, even if they are using semver to "make that ok".

PR #1559, opened 2024-11-13
Make spinner step messages transient

davidmh:
Hi @CKolkey, I've been enjoying neogit for a while, and was pleasantly surprised when I saw the spinner messages on most git operation. Thank you!

However, I've run into a small conflict with the spinner steps. Using `echom` pushes the messages into the message history, which means that if the git operation is long enough, we'll pollute the history with redundant messages.

For some context, those messages are useful to explore the output of different processes, specially when a plugin throws an error.

You can explore the history using the built-in `:messages` command, or you can use something like `:Message` from vim-scriptease which sends the message history into the quickfix window.

In this case, I ran an amend on this commit and took a few seconds to reword the message, that sent 93 messages into the history:

Screenshot 2024-11-13 at 09 30 19

Since it would be great to keep at least one entry in the history, in this commit I'm adding one entry to the messages history, prefixed with `[neogit]` and then replace the spinner step messages with `echo`, which also prints the message but it doesn't save it in the history.

Screenshot 2024-11-13 at 09 26 46

### DIFF
--- a/lua/neogit/spinner.lua
+++ b/lua/neogit/spinner.lua
@@ -36,13 +36,15 @@ end
 function Spinner:start()
   if not self.timer then
     self.timer = vim.uv.new_timer()
+    vim.cmd(string.format("redraw | echomsg '[neogit] %s'", self.text))
+
     self.timer:start(
       250,
       self.interval,
       vim.schedule_wrap(function()
         self.count = self.count + 1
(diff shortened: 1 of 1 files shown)


CKolkey:
Oh, totally, good call.

fnune:
Oh, TIL, thanks!

PR #1587, opened 2024-12-04
fix: git-absorb getting stuck

### LINKED ISSUE #1564: stuck on commit with git-absorb
### Description

I have a problem with using [git-absorb](https://github.com/tummychow/git-absorb) integration in Neogit. The UI freezes until I either kill the process or stop with `ctrl-c`.

Initially i thought it may be related to gpg-agent, like in https://github.com/NeogitOrg/neogit/issues/1460 but eliminated this by disabling the commit signing altogether and the issue remains.

I have also tried removing my `~/.gitconfig` to confirm that it isn't one of my git settings, still no change.

### Neovim version
```
NVIM v0.11.0-nightly+7b7c95d
Build type: Release
LuaJIT 2.1.1713773202
```
```
git-absorb 0.6.13
```

NixOS 24.05

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. Open any file in a working git repository
3. Make any change and save a file
4. Stage the changed file
5. Run commit with absorb (`cx` with default keymap)
6. Select a base commit for absorb as and confirm

### Expected behavior

I would expect the rebase to complete just as it does when the command is executed directly from shell.

### Actual behavior

 After attempting to run the `cx`from status window Neogit:
- prompts for the base commit to be used with git-absorb
- the git-absorb is launched (with correct arguments as far as I can tell)
- git-absorb creates the required `fixup!` commits
- the rebase is started but it gets stuck for some reason
- the `git rebase ...` process is still running but not advancing. `git status` shows that no rebase commands have been executed

### Video
https://github.com/user-attachments/assets/e2171e12-b08b-49b7-a8ab-43415d6df05f

### Debug log files
[neogit.frozen.log](https://github.com/user-attachments/files/17828810/neogit.frozen.log)

### Minimal config

```Lua
local M = {}

function M.root(path)
end

  end

    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
    if vim.v.shell_error > 0 then
        vim.log.levels.ERROR)
    end
  end
end

function M.setup(plugins)

  -- Install required plugins
  if plugins ~= nil then
    end
  end

    callback = function()
      vim.fn.system({
        "rm",
        "-r",
        "-f",
        M.root("xdg")
      })
    end
  })
end

M.setup({
})
```

skrobul:
This fixes neogit hanging forever when using git absorb as reported in the issue #1564

Closes #1564

### DIFF
--- a/lua/neogit/popups/commit/actions.lua
+++ b/lua/neogit/popups/commit/actions.lua
@@ -182,7 +182,7 @@ function M.absorb(popup)
     return
   end
 
-  git.cli.absorb.verbose.base(commit).and_rebase.call()
+  git.cli.absorb.verbose.base(commit).env({ GIT_SEQUENCE_EDITOR = ":" }).and_rebase.call()
 end
 
 return M


CKolkey:
Awesome - why does this work? 😅

skrobul:
@CKolkey from what I gathered, the `--and-rebase` option in git-absorb does start interactive rebase, which in turn starts default git editor (confirmed by looking at the list of processes when this gets stuck). Unfortunately neogit does not display that editor for me. In the case of `git-absorb` I don't want to edit the sequence anyway, so I think it's okay to just allow it to run non-interactively by setting `GIT_SEQUENCE_EDITOR` env variable, just as [neogit does for rebase](https://github.com/NeogitOrg/neogit/blob/dc36aa4b1172189a2e80cbdbd4be2e884e60c09e/lua/neogit/lib/git/rebase.lua#L24).

full disclosure - I didn't really have a lot of time to read through the neogit's source code, so if you have any suggestions how to solve this better I'd be very grateful

CKolkey:
Ah, ok, makes perfect sense then. Didn't think about _why_ it was stuck.. waiting on an editor. 

Thanks for having a look at this :)

PR #1599, opened 2024-12-15
Add ability to revert hunk

kjughx:
This PR adds the ability to revert individual hunks from the commit_view.

In order to do this I had to expand the `Hunk` type slightly, so that it can be just by itself with the corresponding `Diff` it belongs to. Namely I added two new fields:

1. `file`, the name of the file from the where the hunk applies
2. `lines`, this was already kind of part of the type except not in the documentation.

With these changes, the `generate_patch` can work without being passed an `item`, which doesn't exist in the context of a standalone hunk. It is now also able to parse the lines directly on `Hunk`, whether it is a `SelectedHunk` or not. The `from` and `to` parameters now represents the relative lines in the hunk buffer.

With these changes; reverting a hunk is now only a matter of doing a `git apply --reverse` with a patch from `generate_patch`.


lua/neogit/buffers/commit_view/init.lua
```diff
@@ -281,7 +281,7 @@ function M:open(kind)
         end),
         [popups.mapping_for("RemotePopup")] = popups.open("remote"),
         [popups.mapping_for("RevertPopup")] = popups.open("revert", function(p)
-          p { commits = { self.commit_info.oid } }
+          p { commits = { self.commit_info.oid }, hunk = self.buffer.ui:get_hunk_or_filename_under_cursor() }
```
CKolkey, line 293:
```suggestion
          p { commits = { self.commit_info.oid }, hunk = self.buffer.ui:get_hunk_or_filename_under_cursor().hunk }
```

I think it's 2% nicer to pass in the hunk data directly, instead of having two levels of `hunk` keys. Requires updating the consumer below in revert actions.

lua/neogit/popups/revert/actions.lua
```diff
@@ -70,6 +70,14 @@ function M.changes(popup)
   end
 end
 
+function M.hunk(popup)
+  local hunk = popup.state.env.hunk
+  if hunk == nil then
+    return
+  end
+  git.revert.hunk(hunk.hunk, popup:get_arguments())
```
CKolkey, line 78:
```suggestion
  git.revert.hunk(hunk, popup:get_arguments())
```

lua/neogit/popups/revert/init.lua
```diff
@@ -23,6 +23,7 @@ function M.create(env)
     :group_heading("Revert")
     :action_if(not in_progress, "v", "Commit(s)", actions.commits)
     :action_if(not in_progress, "V", "Changes", actions.changes)
+    :action_if(not in_progress, "h", "Hunk", actions.hunk)
```
CKolkey, line 26:
```suggestion
    :action_if(((not in_progress) and env.hunk ~= nil), "h", "Hunk", actions.hunk)
```

Probably only makes sense to show the hunk action if a hunk is available for consumption.

CKolkey:
this is really great! Some minor tweaks then lets get this merged.

lua/neogit/buffers/commit_view/init.lua
kjughx: `get_hunk_or_filename_under_cursor` can return `nil` so cannot the `hunk` key until I check if `nil`. I renamed this key to item so it reads a bit better.

CKolkey: Arg, right. Ok, no reason to overhaul the API then, it's fine as is :)

kjughx:
I'm not sure about these failing tests, if they're my fault or not, because I cannot figure out how to run them locally.

CKolkey:
I don't think it's your fault, and I'll add some instructions on how to do that.

kjughx:
I tried that but seems my version (3.2.3) of ruby is not compatible.

CKolkey:
No worries - this looks good as is. Thanks for taking the time!

PR #1611, opened 2024-12-23
Revert "Add ability to revert hunk"

CKolkey:
Reverts NeogitOrg/neogit#1599

@kjughx Sadly this breaks discarding hunks in a file when there are more than one hunk. Something about the patch generated isn't quite right, but I'm not exactly sure. If you have time to figure out whats up we can try again :)


kjughx:
I'll take a look!

kjughx:
Great, I uploaded #1612 which I think fixes the issue. I'll await the tests to be sure :)

PR #1612, opened 2024-12-24
Add ability to revert hunk

kjughx:
Attempt no 2.

As far as I can tell, all the functionality is there now, including discard a hunk when a file contains many hunks.


CKolkey:
I promise I'll get this reviewed at some point. I just have zero time right now as I have my hands full with twin newborns 😅

kjughx:
No worries! Babies are more important!

PR #1617, opened 2024-12-27
feat(status): add opt-in soft wrap option

Alvaro-Kothe:
This PR introduces the ability to wrap long texts in the status buffer. The default behaviour remains `nowrap`.

With wrap off:
wrap_false

With wrap on:
wrap_true


taketwo:
Note that this will change wrapping behavior in all buffers created using [`Buffer.create()`](https://github.com/search?q=repo%3ANeogitOrg%2Fneogit%20Buffer.create&type=code). Not saying that this is undesired, just pointing out.

Personally, I found this PR while investigating wrapping behavior in `gitcommit` buffers. I prefer wrapping and I enable it explicitly in `ftplugin/gitcommit.vim`. However, it turned out that Neogit overrides my preference because it first sets buffer filetype (triggering `ftplugin`) and only then configures buffer/window.

Given this context, perhaps the best solution would be to:
* have a global configuration option, like you added
* have an ability to override it through `ftplugin` on per-buffer-type basis

To achieve the second item, we need to change buffer initialization order in `Buffer.create()` to set buffer/window options before setting filetype.

Alvaro-Kothe:
Thanks for the feedback! I see these two solutions as alternatives, and I think that the second option is the better one. I’ll update the PR accordingly.

Alvaro-Kothe:
After deleting the line `buffer:set_window_option("wrap", false)` it allows to set the wrap behaviour through the FileType autocommand and ftplugin. Here is a screenshot using `after/ftplugin`

wrap with ftplugin

CKolkey:
Perhaps the better solution would be to trigger the ftplugin /after/ setting up the buffer

taketwo:
> Perhaps the better solution would be to trigger the ftplugin /after/ setting up the buffer

Yeah, that's what I had in mind as well. It seems to me that a combination of a global option that controls behavior in all Neogit buffers and a possibility to override on per-buffer-type basis is most convenient and flexible.

Alvaro-Kothe:
> Perhaps the better solution would be to trigger the ftplugin /after/ setting up the buffer

I've updated the PR to trigger the `ftplugin` after setting the buffer configuration. The wrap option is working fine. But I think it will lead to some unexpected behaviors because of the global configuration defined with `vim.opt.*`. We may have to move some `buffer:set_window_option` commands below the `buffer:set_filetype` to enforce a sane configuration for the buffer.

Alvaro-Kothe:
As I see it, everything that comes after `buffer:set_filetype` is defined as a global configuration for all Neogit buffers, while everything before can be customized through `ftplugin`. For now, I think it’s best to make only the `wrap` option customizable, as that’s the feature I miss the most.

taketwo:
Previously, setting `wrap` was guarded by `if win then`. It looks like one may choose to create a buffer without showing it, in which case there will be no window. I'm not sure what happens if `set_window_option()` is called on a buffer that has no window. Arguably, `Buffer` should abstract this away and, perhaps, stash such set requests and apply them once a window is created.

PR #1640, opened 2025-01-15
Fix console window height

SheffeyG:
Some plugins like noice would hide the cmdline, and the neogit console window height is calculated with a hardcode `2`, this change trying to fix that.
Before:
image
After:
image


SheffeyG:
I've tried to replace the magic num `2` with all vim api, so we need something like `statuscolumn` to tell if the statusline is hiden (could be?), but unfortunally, the `statuscolumn` is not what we need, so I keep that `- 1` there.

CKolkey:
Nice :)

Issue #685, opened 2023-08-01
Add options to set the width and height of floating window


xtrats:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Add options to set the width and height of floating window (`kind = "floating"`), such as

```lua
require('neogit').setup {
  kind = 'floating',
  windows  = {
    width = 0.7,   -- percentage
    height = 0.8 
}
```

**Describe alternatives you've considered**
No for now

PR #1641, opened 2025-01-16
Allow custom floating window style

SheffeyG:
This change allow users to custom floating window like #685, so we can configure it as:
```lua
floating = {
    width = 0.5, -- percentage
    height = 30, -- row
    border = "single",
    ...
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,8 +131,8 @@ neogit.setup {
   -- Floating window style 
   floating = {
     relative = "editor",
-    width = 0.5,
-    height = 0.5,
+    width = 0.8,
+    height = 0.7,
     style = "minimal",
     border = "rounded",
   },
--- a/README.md
+++ b/README.md
@@ -128,6 +128,14 @@ neogit.setup {
   initial_branch_name = "",
   -- Change the default way of opening neogit
   kind = "tab",
+  -- Floating window style 
+  floating = {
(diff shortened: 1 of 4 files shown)


SheffeyG:
@CKolkey There're some type warnings here, and I am not familiar with Lua's type system, I can resolve the type warnings by removing the `floating` type's nil option as
```diff
- ---@field floating? NeogitConfigFloating
+ ---@field floating NeogitConfigFloating
```
or check if it's nil
```lua
if floating then
    ...
end
```
But both of these methods may be incorrect since my method of reading custom configurations might be wrong, could you tell me the correct way to do it?

CKolkey:
Nothing against this, I just am taking a break from programming due to newborn twins.

SheffeyG:
@CKolkey No worries! It's just that Noegit is one of the most powerful and extensive plugins I've used, and adding this config could involve other changes. Recently, I've been using more `auto` instead of `float` in my setup, but I will find another time to fix the type-checking issues and moving this feature on. Anyway, congratulations to you 🎉 and take your time, memories with baby twins are priceless!

P.S. This plugin has helped me a lot, and I've learned many Git commands from it. Thanks for all your hard work!

CKolkey:
Thanks for your contribution :)

PR #1654, opened 2025-02-09
feat: add snacks.picker integration

fang2hou:
Add integration support for the new picker `snacks.picker`, which may become the default picker in future LazyVim releases.

I have tested this integration with both single selection (for branch checkout) and multiple selection (for resetting multiple files).

https://github.com/user-attachments/assets/28dc75c1-a059-4523-9929-97349f85df95

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Here's an example spec for [Lazy](https://github.com/folke/lazy.nvim), but you'r
     "nvim-telescope/telescope.nvim", -- optional
     "ibhagwan/fzf-lua",              -- optional
     "echasnovski/mini.pick",         -- optional
+    "folke/snacks.nvim",             -- optional
   },
   config = true
 }
@@ -246,6 +247,11 @@ neogit.setup {
     -- is also selected then telescope is used instead
     -- Requires you to have `echasnovski/mini.pick` installed.
     mini_pick = nil,
+
+    -- If enabled, uses snacks.picker for menu selection. If the telescope integration
+    -- is also selected then telescope is used instead
+    -- Requires you to have `folke/snacks.nvim` installed.
+    snacks = nil,
   },
   sections = {
     -- Reverting/Cherry Picking
--- a/doc/neogit.txt
+++ b/doc/neogit.txt
@@ -263,6 +263,11 @@ to Neovim users.
         -- is also selected then telescope is used instead
         -- Requires you to have `echasnovski/mini.pick` installed.
         mini_pick = nil,
+
+        -- If enabled, uses snacks.picker for menu selection. If the telescope integration
+        -- is also selected then telescope is used instead
+        -- Requires you to have `folke/snacks.nvim` installed.
+        snacks = nil,
     },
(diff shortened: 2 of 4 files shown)


fang2hou:
The problem detected by linters has been fixed. (I guess 😉)

CKolkey:
Yea that looks like a git thing, not something you would have changed.

I'll try to review this soon - I have 3 month old twins at home so I'm on a bit of a forced programming break 😉

fang2hou:
@CKolkey 
No worries, please enjoy being with your twins! 😙

iniw:
Hey! I was testing this out and noticed that the stash's drop picker doesn't seem to be working properly. the ones you select to drop are actually *kept*, and the ones that weren't selected are dropped, which is the opposite of what I'd expect it to do.

fang2hou:
> Hey! I was testing this out and noticed that the stash's drop picker doesn't seem to be working properly. the ones you select to drop are actually _kept_, and the ones that weren't selected are dropped, which is the opposite of what I'd expect it to do.
> 
> EDIT: That doesn't seem to actually be the case. It seems to always be deleting the first one? It's weird.

I compared the result from snacks and vi.ui.select, there is no difference between them. However I have also noticed the weird handling when I tried to rename my stashes, sometimes after I entered new name, the stash will be dropped! And I tested both snacks & vim.ui.select, and I believe that is the issue from Neogit itself, the fuzzy finder part is not related.

iniw:
>I have also noticed the weird handling when I tried to rename my stashes, sometimes after I entered new name, the stash will be dropped! And I tested both snacks & vim.ui.select, and I believe that is the issue from Neogit itself, the fuzzy finder part is not related.

Can't reproduce this. The only weirdness I noticed is that renaming stashes doesn't focus the input box straight away, I have to `<c-w>w` to it, but I don't really mind that too much since It's not often that I rename stashes.

My initial comment was purely my fault, I expected that selecting multiple stashes with `<tab>` would drop them all, but that's not how it works :)

mehalter:
I have tested this out as well as just using `snacks.picker` by overriding `vim.ui.select` and I'm having problems using the diff range functionality (`dr`). I'm not sure if this is a bug in `snacks.picker` or the diff range implementation in neogit.

In the meantime, I have opened an issue here: https://github.com/folke/snacks.nvim/issues/1561

CKolkey:
The value of a proper integration beyond just overloading `vim.ui.select` is the ability to do stuff like this: https://github.com/NeogitOrg/neogit/blob/master/lua/neogit/lib/finder.lua#L38-L41

That allows valid git refs to be used if the input field contains certain characters, something that wouldn't be possible without overloading the "on_select" function.

Ideally a proper snacks integration would include this logic too, since otherwise it's not much more than overloading `vim.ui.select` (imo).

@fang2hou does that make sense, and is that something you can look into?

iniw:
I've been using this PR daily and can vouch that it works well. Would be nice for it to get merged so that I don't have to keep a fork :)

PR #1655, opened 2025-02-12
reafactor: use `vim.ui.input` instead of `vim.fn.input`

frankfriberg:
As I use `dressing` to show my inputs where im looking, I wanted to update neogit to use `vim.ui.input` to be able to handle this. As alot of plugins use `vim.ui.input` as a standard.

Screenshot 2025-02-12 at 19 23 34

### DIFF
--- a/lua/neogit/lib/input.lua
+++ b/lua/neogit/lib/input.lua
@@ -1,5 +1,14 @@
 local M = {}
 
+local async = require("plenary.async")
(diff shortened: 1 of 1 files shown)


iniw:
I've been using this PR daily and can vouch that it works well. Would be nice for it to get merged so that I don't have to keep a fork :)

PR #1658, opened 2025-02-15
Refs view: branch deletion

actionshrimp:
Hi there! A small QoL improvement for the refs view from magit. I suspect I may have missed an edge case or two, but it is still pretty handy for bulk-deletion when you have lots of old branches hanging round.

- Allows deleting a branch from the refs view more easily with a direct bind ('x' by default), similar to magit.
- Allows deleting multiple branches simultaneously with a visual selection which makes cleaning up lots of branches much simpler.

I wasn't entirely sure if making a new sub-section `refs_view` in the mappings was the ideal option as it didn't exist already and most of the other mappings it uses open other popups, but that's what I went for - happy to amend if it makes sense to do it a different way.

https://github.com/user-attachments/assets/4cf88f33-ed28-4507-b911-11f6484304a0

### DIFF
--- a/lua/neogit/buffers/refs_view/init.lua
+++ b/lua/neogit/buffers/refs_view/init.lua
@@ -61,6 +61,22 @@ function M.delete_branch(ref)
   end
 end
 
+function M.delete_branches(refs)
+  if #refs > 0 then
+    local input = require("neogit.lib.input")
+    local names = {}
+    for _, ref in ipairs(refs) do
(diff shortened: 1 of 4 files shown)


actionshrimp:
(Not completely sure why I'm getting that linter failure as I didn't touch that file, but it seems ok on master?).

CKolkey:
this is dope as hell, thanks for being patient :)

PR #1659, opened 2025-02-17
fix: pushRemote.branch string extraction

fsestini:
The current use of `unpack` and `vim.split` to extract the branch name of an upstream ref fails to properly handle branch names containing forward slash (`/`) symbols. In particular, it discards the branch name's portion including and following the first occurrence of a `/` symbol.

For example:
* `pushRemote.ref == "origin/branch/name"`
* `pushRemote.branch (expected) == "branch/name"`
* `pushRemote.branch (actual) == "branch"`

UI example:
Screenshot 2025-02-16 at 23 04 38
Screenshot 2025-02-16 at 23 04 28

This PR replaces this usage of `unpack` and `vim.split` with a regex match that properly handles branch names containing `/` symbols.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,7 +101,9 @@ See [the test documentation for more details](./tests/README.md).
 Additionally, linting is enforced using `selene` to catch common errors, most of which are also caught by
 `lua-language-server`.
 
-```sh make lint ```
+```sh
+make lint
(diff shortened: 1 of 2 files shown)


fsestini:
Hopefully the linter is happy now.

CKolkey:
Thanks!

PR #1662, opened 2025-02-22
Only try to detach UFO if it is loaded

gilice:
Without this check, it could be that `ufo.setup()` is not called before `ufo.detach()`, (for example if UFO is lazy-loaded with `lazy.nvim` which breaks UFO's internal assumptions, and causes an error.

Example of error:
```
Error executing Lua callback: .../share/nvim/lazy/lazy.nvim/lua/lazy/core/handler/cmd.lua:48: Vim:Error executing Lua callback: ...gilice/.local/share/nvim/lazy/nvim-ufo/lua/ufo/bufmanager.lua:111: attempt to index field 'bufDetachSet' (a nil value)
stack traceback:
	...gilice/.local/share/nvim/lazy/nvim-ufo/lua/ufo/bufmanager.lua:111: in function 'detach'
	/home/gilice/.local/share/nvim/lazy/nvim-ufo/lua/ufo/main.lua:174: in function 'detach'
	/home/gilice/.local/share/nvim/lazy/nvim-ufo/lua/ufo.lua:99: in function 'detach'
	.../.local/share/nvim/lazy/neogit/lua/neogit/lib/buffer.lua:396: in function 'show'
	.../.local/share/nvim/lazy/neogit/lua/neogit/lib/buffer.lua:699: in function 'create'
	...hare/nvim/lazy/neogit/lua/neogit/buffers/status/init.lua:98: in function 'open'
	/home/gilice/.local/share/nvim/lazy/neogit/lua/neogit.lua:114: in function 'open_status_buffer'
	/home/gilice/.local/share/nvim/lazy/neogit/lua/neogit.lua:175: in function 'open'
	/home/gilice/.local/share/nvim/lazy/neogit/plugin/neogit.lua:5: in function </home/gilice/.local/share/nvim/lazy/neogit/plugin/neogit.lua:3>
	[C]: in function 'cmd'
	.../share/nvim/lazy/lazy.nvim/lua/lazy/core/handler/cmd.lua:48: in function <.../share/nvim/lazy/lazy.nvim/lua/lazy/core/handler/cmd.lua:16>
stack traceback:
	[C]: in function 'cmd'
	.../share/nvim/lazy/lazy.nvim/lua/lazy/core/handler/cmd.lua:48: in function <.../share/nvim/lazy/lazy.nvim/lua/lazy/core/handler/cmd.lua:16>


```

PR #1667, opened 2025-02-26
Only detach UFO if it is loaded, 2nd try

algorithmiker:
Fixes cd793fb9d37ad605b8699d8482aeba8a6b17749f (pr #1662).
Fixes https://github.com/NeogitOrg/neogit/issues/1665.

Sorry for my mistake, and thank you @jedrzejboczar for pointing out I should check for `ufo` instead of `nvim-ufo`.

Fun anecdote: I had been running the correct code in my neovim config, because I ran into the same issue (checking for the wrong package name), but I forgot to copy it over to my fork of neogit, before making the PR.

I'd still like this code to be included, if possible, in Neogit, to fix the original issue I was having.

### DIFF
--- a/lua/neogit/lib/buffer.lua
+++ b/lua/neogit/lib/buffer.lua
@@ -388,10 +388,12 @@ function Buffer:show()
   end
 
   -- Workaround UFO getting folds wrong.
-  local ok, ufo = pcall(require, "ufo")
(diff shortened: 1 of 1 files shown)


CKolkey:
Thanks for fixing this, and thanks for being patient 😅

Issue #1621, opened 2025-01-07
Can't stage after resolving conflicts during a rebase


dbatten5:
### Description

If a rebase results in a conflict, I fix the conflict and then try and stage the fixed file using `s`, I see a warning message that says "Conflicts must be resolved before staging" and I'm not able to stage it. However the file does stage if I do `<c-s>`. I'm sure this used to work fine. Interestingly when I had the `diffview` integration on, pressing `s` opened up a `diffview` tab, and then when I closed the tab I saw that the file was staged. Since turning off the integration I'm now seeing the warning message above.

Also, when I am able to eventually stage and commit, when I press `r` I used to have a popup that had `continue` as an option. Now when I press `r` I just get the main rebase menu pop up, no option to `continue`. It seems like `neogit` doesn't know that there's a rebase in progress.

### Neovim version

NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1716656478

### Operating system and version

macOS 14.6.1

### Steps to reproduce

1. Add some commits
2. Rebase with another branch that causes a conflict
3. Fix the conflict in the editor 
4. Try and stage the file in the neogit console using `s`, see that it doesn't stage

### Expected behavior

The file stages

### Actual behavior

The file is not able to be staged.

### Minimal config

```Lua
-- NOTE: See the end of this file if you are reporting an issue, etc. Ignore all the "scary" functions up top, those are
-- used for setup and other operations.
local M = {}

local base_root_path = vim.fn.fnamemodify(debug.getinfo(1, "S").source:sub(2), ":p:h") .. "/.min"
function M.root(path)
  return base_root_path .. "/" .. (path or "")
end

function M.load_plugin(plugin_name, plugin_url)
  local package_root = M.root("plugins/")
  local install_destination = package_root .. plugin_name
  vim.opt.runtimepath:append(install_destination)

  if not vim.loop.fs_stat(package_root) then
    vim.fn.mkdir(package_root, "p")
  end

  if not vim.loop.fs_stat(install_destination) then
    print(string.format("> Downloading plugin '%s' to '%s'", plugin_name, install_destination))
    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
    if vim.v.shell_error > 0 then
      error(string.format("> Failed to clone plugin: '%s' in '%s'!", plugin_name, install_destination),
        vim.log.levels.ERROR)
    end
  end
end

[...]

PR #1675, opened 2025-03-11
Fix: Allow `s` to stage resolved files

gollth:
Thought I might try to fix https://github.com/NeogitOrg/neogit/issues/1621 (=


mofrim:
I made an addition to this in https://github.com/NeogitOrg/neogit/pull/1685

PR #1680, opened 2025-03-13
Error notifications when pulling

sohamshanbhag:
Presently, there are no notifications when there are errors when pulling. This commit adds them.

I'm not sure how you want those to look like, so I've added what feel like sensible defaults. Please feel free to change those.


CKolkey:
Sensible :)

PR #1688, opened 2025-03-15
add option for persistent recent section

### LINKED ISSUE #1687: add an option for persistent recent commits section
Currently the recent section does not display once there are unmerged commits. This behavior is replicated from Magit, however the recent commit section still offers some value even when there are unmerged commits.

Ideally it would be great to have a config that allows from a persistent section. This can be disabled by default, so it doesn't change the default behavior.

Add the option `always = <boolean?>` to the `config.sections.recent` table.
- If this option is not set or set to `false` the recent section should only show when there is no unmerged commits (current behavior)
- if this option is set to `true` the recent sections is always shown

I have a work branch of this change, PR incoming shortly.
N/A

**Additional context**
This options was briefly discussed in #1378

laughingman-hass:
fixes #1687 

This PR adds a new option `config.sections.recent.always`, which defaults to `false`. This keeps the current behavior which only shows the recent section if there is no unmerged commits. which looks like this

image

Setting this option to `true` will always show the recent section. which looks like this

image

### DIFF
--- a/tests/specs/neogit/config_spec.lua
+++ b/tests/specs/neogit/config_spec.lua
@@ -404,6 +404,16 @@ describe("Neogit config", function()
         assert.True(vim.tbl_count(require("neogit.config").validate_config()) ~= 0)
       end)
 
+      it("should return invalid when sections.recent.hidden isn't a boolean", function()
+        config.values.sections.recent.hidden = "not a boolean"
+        assert.True(vim.tbl_count(require("neogit.config").validate_config()) ~= 0)
+      end)
+
+      it("should return invalid when sections.recent.always isn't a boolean", function()
+        config.values.sections.recent.always = "not a boolean"
+        assert.False(vim.tbl_count(require("neogit.config").validate_config()) ~= 0)
+      end)
+
       it("should return invalid when sections.recent.folded isn't a boolean", function()
         config.values.sections.recent.folded = "not a boolean"
         assert.True(vim.tbl_count(require("neogit.config").validate_config()) ~= 0)
--- a/README.md
+++ b/README.md
@@ -290,6 +290,7 @@ neogit.setup {
     recent = {
       folded = true,
       hidden = false,
+      always = false,
     },
     rebase = {
       folded = true,
--- a/doc/neogit.txt
+++ b/doc/neogit.txt
@@ -307,6 +307,7 @@ to Neovim users.
         recent = {
         folded = true,
         hidden = false,
+        always = false,
         },
         rebase = {
         folded = true,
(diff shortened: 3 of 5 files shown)


tests/specs/neogit/config_spec.lua
```diff
@@ -404,6 +404,16 @@ describe("Neogit config", function()
         assert.True(vim.tbl_count(require("neogit.config").validate_config()) ~= 0)
       end)
 
+      it("should return invalid when sections.recent.hidden isn't a boolean", function()
+        config.values.sections.recent.hidden = "not a boolean"
+        assert.True(vim.tbl_count(require("neogit.config").validate_config()) ~= 0)
```
laughingman-hass, line 409: Added the additional test for the `hidden`. I'm unsure why this results to `true`, even the the default config is `false`

laughingman-hass:
I didn't find a documentation section, where I could provide additional context for the option or the current behavior.

tests/specs/neogit/config_spec.lua
SheffeyG: This test simulate `hidden` with a string value, and the result means this test passed I guess.

SheffeyG:
I also prefer the 'Recent Commits' section for rebasing and other actions, but displaying both the Recent and Unmerged sections simultaneously causes a rendering bug:

image

In the screenshot, both Unpush and Recent sections has commit `test`, and I choosed the first one in the 'Unpushed' section, but the last one in the 'Recent' section get rendered.

This bug is kind hard to fix, as neogit use `oid` (commit hash) to identify each commit components, after the components `oid` indexed, if buffer components `oid` is not unique, it will always point to the last component. I've tried to fix this, but there's no cursor status stored, so I give up.

A workaround for this is hide the Unmerged section as `unmerged_upstream = { hidden = true }`, so the Recent Commits section is always shown, and it will correctly rendered.

laughingman-hass:
@SheffeyG thats interesting. I'm currently away and will have a look if there is a nice way to fix this when i'm back. Personally i didn't like the same commit showing up in recent and unmerged, maybe there is a way to just show new commits in unmerged and recent to show previous commits.

CKolkey:
So... I think I made a pretty core assumption that an object's OID would only be in a given buffer once: https://github.com/NeogitOrg/neogit/blob/df08cb97f126c546a5ec265bafab2d78cc7399fa/lua/neogit/lib/ui/renderer.lua#L54-L56

That index is used to apply the "active item" highlight.
https://github.com/NeogitOrg/neogit/blob/df08cb97f126c546a5ec265bafab2d78cc7399fa/lua/neogit/lib/buffer.lua#L896

And given that it's a hash-like lua table, last entry for a given OID wins. This is what @SheffeyG was pointing out, and... I don't think there's an easy fix. 

@laughingman-hass You could in theory XOR the "unmerged" and "recent" commits, but both are fetched concurrently when we re-sync internal state to repo state, so you'd need to have some kind of callback that operates on both tables after they are updated. Kind of a PITA if you ask me.

PR #1699, opened 2025-04-03
Add support for disabling commit signing on rebase

skrobul:
Current version of Neogit allows users to enable commit signing if they have it globally disabled. However, there is no way to explicitly disable it if it has been enabled globally.

My personal use case for this is that I generally sign all of my commits, but occasionally when I'm doing local rebase of big branches, I want to switch the GPG signing off.


CKolkey:
With https://github.com/NeogitOrg/neogit/pull/1705 there is now an API for adding whatever switches/options/actions to popups that you would desire - will that work?

PR #1705, opened 2025-04-16
feat: #1684 Ability to add custom popup switches

### LINKED ISSUE #1684: Ability to add custom predefined argument switches to the push menu
There's a few arguments you can pass while pushing:
Image

It would be great if it was possible to add additional custom options here that can be appended to the push command. One example is `git push -o merge_request.create` where I would like to be able to add an argument switch on keys `-m` which would add the argument `-o merge_request.create` to the push command.

The use-case is once I am done with my feature branch and it's time to make an MR, it's easy to do so automatically via the `merge_request.create` option.
I believe `merge_request.create` is a GitLab specific option, hence why my suggested solution is the ability to append *custom* arguments to the push command, so you can tailor it to whatever backend you use. There's probably other useful options that can be passed via `-o <option>`.

Right now I have not found any alternatives, I push via the CLI every time I want to make an MR at the same time as the push.

I'd be happy to take a stab at this myself and make an MR, if this is something you guys would be OK with.

AlexanderArvidsson:
Adds the ability to add custom argument switches to any popups by allowing users to create custom builder hooks to modify popups while they're building.
image

The above is implemented in config like this to provide the `-o merge_request.create` GitLab specific option:
```lua
require('neogit').setup({
  builders = {
    NeogitPushPopup = function(builder)
      --
      builder:switch('m', 'merge_request.create', 'Create merge request', { cli_prefix = '-o ', persisted = false })
    end,
  },
})
```

This PR also introduces the `persisted = false` option for switches, and it also migrates all existing non-persisted switches to this flag instead of the `ignored_settings` option. It doesn't remove `ignored_settings`, as people might have overridden it.
It also adds `persisted = false` to some extra popups that wasn't covered before (mostly `force` switches).

Fixes #1684.

## Future changes
* Modify existing switches' options, to change their keybinds, text, persisted option, etc.
* Set an ordering for switches, or the ability to _inject_ new options anywhere in the list.


CKolkey:
Good stuff! A quick look at the code doesn't show any glaring issues. But would you mind adding some documentation to the help file? Under API or something that makes sense :)

AlexanderArvidsson:
I added some docs under `neogit_custom_popups` and also updated the references to the default config, which has now changed (migrated to `persisted` flags, it _shouldn't_ introduce any breaking changes afaict.

Let me know if you want to make some changes or move it somewhere else!

CKolkey:
This is great, thanks a ton for implementing it!

PR #1706, opened 2025-04-16
fix: set border=none for header text explicitly

folliehiyuki:
This is only noticeable when a user sets `vim.o.winborder` to a custom value. Without the border option explicitly provided, the header will inherit the border style from the global winborder option, overlapping some first lines in the log view.

Screenshots with `vim.o.winborder = "rounded"`:
* current settings:

Screenshot_20250416_225754

* after this PR (will be back to the expected behavior):

Screenshot_20250416_225708


folliehiyuki:
For context, `winborder` global option was introduced in Neovim 0.11 recently.

CKolkey:
Thanks!

PR #653, opened 2023-07-21
Add flag to include author information in recent commits

treatybreaker:
This PR does the following:
- Adds capability to show author name and email in recent commits
  - By default this is disabled, it can be enabled by setting `status.recent_commits_include_author_info` to `true`
- Adds syntax matching for recent commit messages to color this new information
- Adds tests to cover this new functionality and prior functionality that wasn't being tested
- Updates the Readme to reflect the new flag

See below for how the recent commits look with the author information included:
image

Closes #652 😌

PR #1715, opened 2025-04-27
Add log margin popup

lucasadelino:
This is another attempt at adding author (and date) information to the recent commits section in the status buffer, following #653.

**What currently works:**
- Toggling the visibility on/off (hides both author and date)
- Cycling through three different date styles (relative date, abbreviated relative date, or an "ISO-like" format converted to the local timezone)
- Toggling the author name on/off

I tried to follow Magit's default behavior as closely as I could. For the ISO-like date, I added a field to `lib/git/log.lua` to retrieve the commited date as a UNIX timestamp, which was then rendered in the local timezone. I think that should be fine, but if that introduces peformance concerns, or if there is a cleverer way to do this, I'm happy to change this.

**What doesn't work yet**
Pretty much all of the other features. Most notably:
- Toggling shortstat
- All of the popup arguments
- The popup only works in the status buffer; in Magit, it also works in the log buffer
- Magit also has a config setting to control what date (author or commit) is shown in the margins. Currently, only the commit date is shown.

I wanted to put this out here for preliminary feedback, but I still need to write some tests before this is ready to merge. After that, I'm happy to work on the other features, be that before or after merging this initial implementation (maybe after, judging by the discussion in #653?).

**A question**
I used popup actions to implement the three features above. This was the closest Neogit correlate I found to mimick Magit, but I'm wondering if I have this right. 

I think what's tripping me up is that these features seem more like switches than actions proper (for instance, these "actions" don't spawn any new buffers or popups), so mayble internal switches would be better here. Actions also seem to always close the popup, whereas the margin popup stays open in Magit (though we may also opt not to follow Magit's behavior). I may be overthinking this, but it was something I thought I should check anyway.

Happy to hear any thoughts on this; this the first time I've worked on a Neovim plugin, so any feedback is much appeciated!


CKolkey:
> A question
> I used popup actions to implement the three features above. This was the closest Neogit correlate I found to mimick Magit, but I'm wondering if I have this right.
> 
> I think what's tripping me up is that these features seem more like switches than actions proper (for instance, these "actions" don't spawn any new buffers or popups), so mayble internal switches would be better here. Actions also seem to always close the popup, whereas the margin popup stays open in Magit (though we may also opt not to follow Magit's behavior). I may be overthinking this, but it was something I thought I should check anyway.

Yeah, for the time being, actions seem like the correct way to build this. The auto-close thing is not always ideal - for a while it wasn't an issue, but this isn't the first time it's been a limitation on desirable behavior. Maybe I can do something about that.


Overall this looks very reasonable to me!

lucasadelino:
Okay, I think I figured out why the help popup test was failing. I also added a test spec for the margin popup. Hopefully we should be good now!

lucasadelino:
OK, _now_ we should be good. I think whatever was causing that branch test to fail was solved in one of the recent commits to master. Just did one last force push to get rid of an unintended merge commit. Thanks for bearing with me!

CKolkey:
Ah, no worries, I'm always happy when people contribute :)  

The branch popup test is a bit flakey, no worries there. I'll look at it some day.. maybe. 🤣

lucasadelino:
I've been taking another look at this this week, and implemented two of the arguments: order (topo, author, or commit), and showing/hiding refnames. 

In the process of doing so, and while following [your suggestion](https://github.com/NeogitOrg/neogit/pull/654#issuecomment-1646562987) of implementing these arguments as configs, I noticed that the status buffer automatically refreshes when changing configs in the margin popup. I am now wondering if the actions we have so far (date style, margin visibility, etc) would be best implemented as configs. This would be a weird in-between, as far as following Magit's behavior: it would be closer to Magit in that we'd be able to cycle through the configs without closing the popup, but farther in that the popup would look very different. If you want to see what that looks like, check out [this branch](https://github.com/lucasadelino/neogit/tree/margin_config) on my fork.

What do you think, @CKolkey? IMO, having everything as configs feels nicer to use, but I'll defer to what you think is best!

I've also taken a shot at the shorststat and graph toggles, but I think it would be better for these features to be on different PRs. The shortstat is pretty much done and seems to work okay, but I haven't stress-tested it on a large repo. I pretty much just copied Magit's implementation, which they [describe](https://github.com/magit/magit/blob/fd1882b8c981c5e859522bde1dd9c88af9485709/lisp/magit-log.el#L1605) as "experimental and rather slow". The graph will be a doozy, since it would require adding support for printing empty lines (for graph connectors, as in the Log view) in recent commits. Both these features seem a bit more complex than what we have so far, which already serves users who just want the basic functionality of showing author names and dates.

CKolkey:
I'm going to push a small change that will allow cycling options much more nicely. No reason to make them pseudo-configs, when they are more appropriately represented as actions.

Definitely prefer this over the config approach, aesthetically.

I didn't realize this popup should control the margin in the log popup too! Makes sense, in hindsight, though does present some challenges.. 

I really like your use of the `state` system to track these values. Never would have thought of that use case here.

Finally, I did a little formatting change to the virtual text logic to hide the work behind a conditional, so it's not evaluated and discarded when not visible.

CKolkey:
To begin consuming some of the args, you'd want to put them here: https://github.com/lucasadelino/neogit/blob/f38a8dd9eb6f7ceaefe5be2d579ed446bb541cdb/lua/neogit/lib/git/log.lua#L422-L430

And iirc options can have a function attached (https://github.com/lucasadelino/neogit/blob/f38a8dd9eb6f7ceaefe5be2d579ed446bb541cdb/lua/neogit/popups/log/init.lua#L33) so that might be a way to set the state when changing options.

Donno if switches have that, but it could be implemented.

CKolkey:
Had to change the original way I persisted popups, but this way should work now. Also took the liberty of implementing the refresh buffer function - pretty simple, just some plumbing.

CKolkey:
So, this is way too cool to just languish in a PR. Finally took a second to polish it up for releasing - you can see the changes I made were quite minor. 

If you're up for it, it would be really cool to get the rest of the options working, but even without that this seems like a reasonable checkpoint to make a release of the feature.

Thanks for being patient, and thanks for the PR!

lucasadelino:
Thank you so much! Neogit is awesome, I'm happy to be able to contribute. I also finally had some time to get back to this. I already have a draft for two of the options; I'll open up another PR for them (and a separate PR for the shortstat action, as I had mentioned).

Issue #1695, opened 2025-03-25
Directory tree layout for files in status view


ceuk:
Hey,

Is there any appetite for an option to structure the file lists in the status buffer in a tree view?

e.g. this is how LazyGit does it:

![Image](https://github.com/user-attachments/assets/18af1ebd-8610-4917-9328-162a013c064a)

We could either keep the current foldable hunk diffs and/or leverage the peek diff view from the log where the diff updates as you scroll the file list

I was thinking about putting together a PR to this effect. Is that something the maintainers would be open to?

Thanks

CKolkey:
I'm not personally keen on it, but I'm not opposed to a PR either. I suspect there's some appetite for it from users.

PR #1720, opened 2025-05-05
Add status.tree_view option

ceuk:
Hello :)

Bit of a POC really to show how #1695 might work in order to get some feedback/direction.

image

I think it'd also be good to be able to stage/unstage entire sections of files. Will look at that shortly, trying to squeeze this in around my work/family commitments

Let me know if this is going in the right direction anyway


DanRioDev:
I found this draft searching for a solution to how to stage everything inside a folder, will this be an expected behavior?

PR #1726, opened 2025-05-09
Add command to copy finder selection

benjiwolff:
I was looking for a convenient way to copy branch names and found #1520. However, the proposed solution does not allow copying other branch names than the current one or commit hashes.

I would love a shortcut that allows copying the selection from any finder.

# Todo
- [x] implement/test multi-select finders

### DIFF
--- a/lua/neogit/lib/finder.lua
+++ b/lua/neogit/lib/finder.lua
@@ -12,7 +12,7 @@ end
 local copy_selection = function()
   local selection = require("telescope.actions.state").get_selected_entry()
   if selection ~= nil then
-    vim.cmd.let(("@+='%s'"):format(selection[1]))
+    vim.cmd.let(("@+=%q"):format(selection[1]))
   end
 end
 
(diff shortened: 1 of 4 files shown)


CKolkey:
Seems perfectly reasonable to me :)

lua/neogit/lib/finder.lua
```diff
@@ -9,6 +9,13 @@ local function refocus_status_buffer()
   end
 end
 
+local copy_selection = function()
+  local selection = require("telescope.actions.state").get_selected_entry()
+  if selection ~= nil then
+    vim.cmd.let(("@+='%s'"):format(selection[1]))
```
CKolkey, line 15: I think you could use `%q` instead of `'%s'` to let :format() handle quoting and escaping here.

benjiwolff:
Multi-select finders are not affected. Copy "selection" only refers to the current position of the cursor.

PR #1731, opened 2025-05-12
Fix snacks.picker confirm action and on_close callback

garrett-hopper:
Snacks.picker does not call a `close` action by default. This change uses the `on_close` callback instead. ([snacks.nvim/docs/picker.md](https://github.com/folke/snacks.nvim/blob/main/docs/picker.md))

Note: `confirm` is a shortcut to `actions = { confirm = confirm }`


CKolkey:
Thanks :) I don't use snacks, so I appreciate you helping out here

PR #1737, opened 2025-05-21
Allow access to logger config via lua api at runtime

actionshrimp:
Expose the logger config to allow updating at runtime.

I've been interested in investigating some slowdown in a large repo with lots of submodules for a while, but the slowness only shows up when neovim has been open for a long time, and I usually haven't started neovim with neogit's logging configured to investigate easily.

This allows you to dynamically turn on logging and set the log level to debug, and also updates the contributing doc to explain how.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,9 +50,23 @@ Simply clone *Neogit* to your project directory of choice to be able to use your
 Logging is a useful tool for inspecting what happens in the code and in what order. Neogit uses
 [`Plenary`](https://github.com/nvim-lua/plenary.nvim) for logging.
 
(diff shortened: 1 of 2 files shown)


CKolkey:
Dope - lemme know what you find!

actionshrimp:
Oh interesting, does updating the env vars via direnv enable the logger while nvim is already running then? AFAICT it looked like it would only pay attention to the env when the logger module was first loaded but perhaps I misread.

(Shoutout to direnv though, I'm a big fan - check out my pure lua direnv plugin 😇  https://github.com/actionshrimp/direnv.nvim)

CKolkey:
Ah, right on, that's cool! You're right though, it only checks the variables when nvim starts, changing them during runtime has no effect. I just have debugging turned on all the time - you know, stuff randomly comes up while you're at work and you can just see what the heck happened.

PR #1748, opened 2025-05-30
Improved diff popup with fzf-lua & advanced range selections

Aitai:
This PR brings a bunch of new commands to the diff popup that I've been missing personally.

*   **New diffing options:**
    *   Diff between branch ranges (`b`), commit/ref ranges (`c`), and tag ranges (`t`).
    *   Diff HEAD to a specific commit/ref (`h`).
    *   Diff specific files (`f`) or paths/globs (`p`) against HEAD.
    *   The Stash diff (`S`) is also updated to use the new picker.
*   **Integrating fzf-lua:**
    *   Uses `fzf-lua` (if available) for selecting branches, commits, tags, and stashes.
*   **Internal changes:**
    *   `diffview.open` can now accept an array of arguments, simplifying how diffs with multiple file arguments are invoked.
    *   Removed `--literal-pathspecs` from default git arguments to allow globs in paths.

**Testing**:
I've tested these changes on Neovim v0.12.0-dev-461+gb28bbee539. The fzf-lua integration works as expected, and the fallback to another fuzzy finder (which I've tested with Telescope) works too.


CKolkey:
I edited my above comment before seeing you had replied already, so have a look.

> The existing fzf-lua integration works well for general list fuzzy finding, but I've tried to enhance the diff menu by using fzf-lua's specific providers (like fzf_lua.git_branches(), fzf_lua.git_commits(), etc.) when available for actions like branch/commit/stash/tag ranges. This offers richer previews and native fzf-lua handling for those specific object types. If these specific fzf-lua methods aren't available or for other generic lists, it falls back to Neogit's FuzzyFinderBuffer (which then uses the existing generic fzf-lua integration if enabled).

Ah, ok, that's going to be a no-go then. I'm not interested in having fzf specific stuff at this layer, because that would mean snacks and telescope specific stuff would also belong here... which quickly gets out of hand.

CKolkey:
Really great write up - lets dig into it.

1. One of the ideas with this plugin is that the entire UI is very interactive. For example, the reason that the fuzzy finder doesn't list commits is because there's already a view for that - the log and reflog views. When you have a commit under the cursor in the log view, it's (almost always) passed in some way to the action invoked. Now, you've correctly spotted that this _isn't_ wired up for the diff#this and diff#range actions - but it should be.

As for why it wouldn't work with manual entry... I don't know. That should 😅

Regardless, it's quite a simple fix, so I'll merge that now. I know discoverability of this "everything is interactive" isn't great - in fact I think it's mostly hidden unless you happen to notice it working by accident. Open to ideas on improving that.

2. I think there are two main points here: showing more specialized information in the picker, and having individual actions for different "types" of refs.

To the first point, maybe I'm weird, but I don't _personally_ see the value-add of that information. I can understand that others would, but IMO having a single list that supports commits/branches/tags/refs/stashes is ideal. 

That said, I know not everyone will agree with me 😄 

The second point, having separate actions per type, I think boils down to either providing a more "user friendly" and "rich" UI with specific actions for specific "types" of objects, or a UI that recognizes that all those "types" are all just revisions under the surface, and providing one interface that doesn't discriminate (but is therefore less "rich").

I think I see an issue, though, with more bespoke pickers like you describe. Items like `HEAD` and such no longer have a place, but are quite valuable. If you wanted to diff a tag against `CHERRY_PICK_HEAD`, for example, how would that be accomplished?

(the above issue aside, I think your "item_type" solution is reasonable)

There's probably some points I missed, but that's the meat of it :)

PR #1756, opened 2025-06-04
Fix stash diff error by adding nil check for stash_id

### LINKED ISSUE #1713: Error message on stash d-d
### Description

get error message when trying to show diff on stash item
steps to reproduce:
- open Neogit tab
- unfold stash list
- select any stash
- type d d (diff this)
- error message appears

it seems issue in file lua/neogit/integrations/diffview.lua:129
it is trying to get stash name, but instead of stash description line parameter item_name has value of commit hash and match function returns empty line (and concatenation fails).

proposed fix: update line https://github.com/NeogitOrg/neogit/blob/9bb1e73c534f767607e0a888f3de4c942825c501/lua/neogit/integrations/diffview.lua#L130 to
```
    view = dv_lib.diffview_open(dv_utils.tbl_pack(stash_id and stash_id .. "^!" or item_name))
```

### Neovim version

NVIM v0.11.0
build type: release
LuaJIT 2.1.1241730670

git 2.49.0

Linux Ubuntu 20.04

### Steps to reproduce

steps to reproduce:
- open Neogit tab
- unfold stash list
- select any stash
- type d d (diff this)
- error message appears

### Expected behavior

show diff

### Actual behavior

concatenation error appears

### Minimal config

```Lua
local M = {}

function M.root(path)
end

  end

    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
    if vim.v.shell_error > 0 then
        vim.log.levels.ERROR)
    end
  end
end

function M.setup(plugins)

  -- Install required plugins
  if plugins ~= nil then
    end
  end

    callback = function()
      vim.fn.system({
        "rm",
        "-r",
        "-f",
        M.root("xdg")
      })
    end
  })
end

M.setup({
})
```

Aitai:
Closes #1713

I implemented the proposed fix from the issue to resolve the error that occurs when trying to show a diff on a stash item. The problem was in the `diffview.lua` file where it was trying to parse the stash name via the `item_name` variable, but the match function could return nil, causing concatenation to fail.

The fix adds a nil check for `stash_id` and provides a fallback to use `item_name` directly if no match is found, preventing the concatenation error when viewing stash diffs.

### DIFF
--- a/lua/neogit/buffers/stash_list_view/init.lua
+++ b/lua/neogit/buffers/stash_list_view/init.lua
@@ -97,7 +97,7 @@ function M:open()
         [popups.mapping_for("DiffPopup")] = popups.open("diff", function(p)
           local items = self.buffer.ui:get_commits_in_selection()
           p {
-            section = { name = "log" },
+            section = { name = "stashes" },
             item = { name = items },
           }
         end),
@@ -166,7 +166,7 @@ function M:open()
         [popups.mapping_for("DiffPopup")] = popups.open("diff", function(p)
           local item = self.buffer.ui:get_commit_under_cursor()
           p {
-            section = { name = "log" },
+            section = { name = "stashes" },
             item = { name = item },
           }
         end),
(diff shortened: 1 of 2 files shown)


hoopoepg (approved):
thank you!!!

CKolkey:
I'm going to go about solving this a bit more simply: just consolidate the commit and stash cases, since they can actually be handled the same way, and add "and item_name" clauses where they are required.

https://github.com/NeogitOrg/neogit/pull/1763

That said, I appreciate your PR none the less :)

PR #1757, opened 2025-06-05
Several option flags

antonrogov:
* simple_headers (default false) - use a simple line instead of windows for headers. Header window didn't close for me occasionally and stick until vim is closed.
* force_if_includes (default true) - for some reason pushing to another remote branch (different name, not an upstream etc) with --force-with-lease didn't work for me if --force-if-includes was set. Also an option to opt out probably shouldn't hurt.
* log_view_esc_close (default true) - I'd like to keep things consistent and use the same "Close" binding for all buffers and also keep spamming esc in log view without it closing
* hard_reset_backup (default true) - ability to turn those commits off

### DIFF
--- a/lua/neogit/buffers/log_view/init.lua
+++ b/lua/neogit/buffers/log_view/init.lua
@@ -9,6 +9,7 @@ local git = require("neogit.lib.git")
 local a = require("plenary.async")
 local notification = require("neogit.lib.notification")
 local git = require("neogit.lib.git")
+local logger = require("neogit.logger") -- TODO: Add logging
 
 ---@class LogViewBuffer
 ---@field commits CommitLogEntry[]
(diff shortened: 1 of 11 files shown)


CKolkey:
I think some extra changes are sneaking in on accident because you are opening a PR from your master to mine - you could make a branch on your fork to be more selective about what changes are included here :)

antonrogov:
Sorry, yes :) Added all the changes in one batch first and then split into this one, https://github.com/NeogitOrg/neogit/pull/1759 and https://github.com/NeogitOrg/neogit/pull/1758 (both from separate branches), but forgot one commit - removed and re-pushed. Also just realised it's not needed since there's `neogit.action(..)` already

CKolkey:
No worries. I'll try to make some time to review this all soon ;)

CKolkey:
> * simple_headers (default false) - use a simple line instead of windows for headers. Header window didn't close for me occasionally and stick until vim is closed.

👍🏼 

I suspect you will want to disable this, as well: https://github.com/NeogitOrg/neogit/blob/63f95f204fafcab0252d789902fc360b17e15ab0/lua/neogit/lib/buffer.lua#L953-L956


> * force_if_includes (default true) - for some reason pushing to another remote branch (different name, not an upstream etc) with --force-with-lease didn't work for me if --force-if-includes was set. Also an option to opt out probably shouldn't hurt.

👍🏼 


> * hard_reset_backup (default true) - ability to turn those commits off

👍🏼 

> * log_view_esc_close (default true)

I think this should be handled the same way all the other "mappings" stuff is, with a new table in the `config.mappings` table called `log_view`, that lets you set `["<esc>"] = false` to disable.

PR #1760, opened 2025-06-07
Keep current file selected after staging

### LINKED ISSUE #1448: Diffviewer changes focus to first file after save
### Description

In diffview, after staging chunk of changes, the focus goes back to the first file in the navigator. This becomes a problem when diffing multiple files and auto-save plugin is enabled.

What happens is that I constantly need to find the last staged file, and move to the next one. It is quite an issue when trying to stage many different files.

https://github.com/user-attachments/assets/12657e65-7534-4dd6-8bc3-5e0aae6830d1

### Neovim version

NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1720049189

Fedora Bluefin

### Steps to reproduce

1. Prepare repo:
    ```bash
    mkdir /tmp/repro
    cd /tmp/repro
    echo "README\n" > README.md
    echo "UPDATING\n" > UPDATING.md
    echo "TEST\n" > TEST.md
    git add .
    git commit -m "Add files"
    ```
2. Make some changes:
    ```bash
    echo "Hello There" >> README.md
    echo "Hello There" >> UPDATING.md
    echo "Hello There" >> TEST.md
    ```
3. `nvim -u minimal.lua`
4. `:Neogit<CR>`
5. `d u` (open diff for unstaged files)
6. Stage changes for the first file
7. Stage changes for the second file

### Expected behavior

Focus should remain on the just-staged file.

### Actual behavior

Focus moves to the first file in tree.

### Minimal config

```Lua
vim.env.LAZY_STDPATH = "/tmp/.repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
    {
      "NeogitOrg/neogit",
      dependencies = {
        "nvim-lua/plenary.nvim",         -- required
        "sindrets/diffview.nvim",        -- optional - Diff integration

        -- Only one of these is needed, not both.
        "nvim-telescope/telescope.nvim", -- optional
        "ibhagwan/fzf-lua",              -- optional
      },
      config = true
    },
  },
})
```

Aitai:
This PR fixes #1448 where Diffview would jump back to the first file after saving.

To fix this in Neogit, I've updated the Diffview integration:
1.  When files are staged, I now capture the path of the file currently open in the active diff buffer.
2.  This is then passed to `update_files`, which then uses it to ensure the correct file remains selected.

**Note on a related Diffview issue:**
I also noticed that Diffview itself doesn't always handle writing edits to *staged* files correctly (changes made in the diff view to a staged file might not reflect in the worktree, appearing lost on refresh). I've fixed this in my fork of Diffview.

### DIFF
--- a/lua/neogit/integrations/diffview.lua
+++ b/lua/neogit/integrations/diffview.lua
@@ -18,7 +18,7 @@ local function get_local_diff_view(section_name, item_name, opts)
     section_name = "working"
   end
 
-  local function update_files()
+  local function update_files(current_file_path)
     local files = {}
 
     local sections = {
@@ -44,7 +44,9 @@ local function get_local_diff_view(section_name, item_name, opts)
           } or nil,
           left_null = vim.tbl_contains({ "A", "?" }, item.mode),
           right_null = false,
-          selected = (item_name and item.name == item_name) or (not item_name and idx == 1),
+          selected = (current_file_path and item.name == current_file_path)
+            or (item_name and item.name == item_name)
+            or (not item_name and not current_file_path and idx == 1),
         }
 
         if opts.only then
@@ -83,9 +85,15 @@ local function get_local_diff_view(section_name, item_name, opts)
     end,
   }
 
-  view:on_files_staged(a.void(function(_)
+  view:on_files_staged(a.void(function()
+    local current_file_path
+    local current_buf_name = vim.api.nvim_buf_get_name(0)
+    if current_buf_name and current_buf_name ~= "" then
+      current_file_path = current_buf_name:gsub(view.git_root .. "/", "")
+    end
+
     Watcher.instance():dispatch_refresh()
-    view:update_files()
+    view:update_files(current_file_path)
   end))
 
   dv_lib.add_view(view)


CKolkey:
I'm going to do my best to review all your PR's soon, just bear in mind I have three kids, two of which are six months old 😅

Aitai:
No worries! Take your time, I know how demanding twins can be.

lua/neogit/integrations/diffview.lua
```diff
@@ -18,23 +17,33 @@ local function get_local_diff_view(section_name, item_name, opts)
     section_name = "working"
   end
 
-  local function update_files()
+  local function update_files(current_file_path)
```
CKolkey, line 20: it looks like this argument will always be nil? I don't see any callers passing it.

lua/neogit/integrations/diffview.lua
```diff
@@ -18,23 +17,33 @@ local function get_local_diff_view(section_name, item_name, opts)
...
 
+    git.repo:dispatch_refresh {
+      source = "diffview_update",
+      callback = function() end,
+    }
+
+    local repo_state = git.repo.state
+    if not repo_state then
+      return files
+    end
```
CKolkey, line 31: Not sure about this - the git repo should _always_ have state at this point. How would you get here with an uninitialized repository?

lua/neogit/integrations/diffview.lua
```diff
@@ -84,11 +95,11 @@ local function get_local_diff_view(section_name, item_name, opts)
...
-  view:on_files_staged(a.void(function(_)
-    Watcher.instance():dispatch_refresh()
-    view:update_files()
-  end))
-
+  view:on_files_staged(function()
+    vim.schedule(function()
+      Watcher.instance():dispatch_refresh()
+    end)
+  end)
```
CKolkey, line 102: Can you explain why this needs to be scheduled, instead of non-blocking async? Not sure I follow.

lua/neogit/integrations/diffview.lua
Aitai: You're right, that argument is never a file path as its name implied. My initial implementation incorrectly assumed I needed to manage the selected state during refreshes. I later realized diffview actually passes its view object to the update callback, and the old code was only working due to a subtle side-effect: the truthy view object prevented my fallback logic from running, which implicitly preserved the selection.

I've made a new commit which aligns correctly with the diffview API. The selected flag is now set only on the initial file list passed to the constructor. The update_files function acts as a pure data provider for refreshes, which properly leaves the selection management to diffview itself.

PR #1761, opened 2025-06-08
Add auto-fetch feature

### LINKED ISSUE #1670: Automatically fetch from remotes at an interval

When checking out e.g. the branch of a pull request, I need to first `fa` (fetch all) and then `bl` to find the remote branch among the selection options. If Neogit fetched remotes automatically at an interval, this would not be necessary.

I'd like Neogit to automatically fetch from all (or configured) remotes when Neovim starts, or when Neogit starts, at an interval.

See how e.g. [Tower does this](https://www.git-tower.com/help/guides/remote-repositories/fetch/windows).

Configuration suggestion:

```lua
auto_fetch_remotes_interval = int | nil -- Nil is default, nil disables
```

Use a ZSH plug-in such as [this one](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/git-auto-fetch/git-auto-fetch.plugin.zsh) but it seems to requires Oh-my-ZSH which I don't want to use.

Aitai:
This PR introduces an automatic fetching feature that periodically fetches from all remotes.

### Configuration

The auto-fetch feature is controlled by the following settings:

- **`auto_fetch_enabled`**: A boolean to enable or disable the feature. It is `false` by default.
- **`auto_fetch_interval`**: The time in milliseconds between fetches. Defaults to `300000` (5 minutes).
- **`auto_fetch_on_startup`**: A boolean to perform a fetch immediately when Neogit starts. Defaults to `false`.

When enabled, Neogit will execute `git fetch --all` at the specified interval.

Closes #1670

### DIFF
--- a/lua/neogit.lua
+++ b/lua/neogit.lua
@@ -73,34 +73,6 @@ function M.setup(opts)
   signs.setup(config.values)
   state.setup(config.values)
   autocmds.setup()
-
-  if config.values.auto_fetch_enabled then
-    local git = require("neogit.lib.git")
-
-    if M.auto_fetch_timer then
-      M.auto_fetch_timer:close()
-    end
-
-    M.auto_fetch_timer = vim.uv.new_timer()
-    M.auto_fetch_timer:start(
-      config.values.auto_fetch_interval,
-      config.values.auto_fetch_interval,
-      vim.schedule_wrap(function()
-        if git.cli.is_inside_worktree(vim.uv.cwd()) then
-          git.fetch.fetch("--all")
-        end
-      end)
-    )
-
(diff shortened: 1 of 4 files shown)


CKolkey (changes requested):

Aitai:
Hey, thanks for the detailed feedback! I've implemented the requested changes.

The auto-fetch logic is now managed by each NeogitRepo instance, and the git.fetch function has been refactored to separate its core logic from user-facing side effects like notifications.

Let me know what you think!

Wikiwix:
@CKolkey @Aitai Did this get lost or is there a specific reason for the hold up?

Issue #1765, opened 2025-06-12
NeogitStatus folds cannot be manually opened


cassava:
### Description

Folds in the `NeogitStatus` window, such as "Unstaged changes" cannot be navigated with the standard fold-keybindings, such as `zo`, `zc` UNTIL the toggle action has been applied to it (i.e. `<tab>`)

### Neovim version

NVIM v0.11.2
Build type: Release
LuaJIT 2.1.1741730670

### Operating system and version

Ubuntu 22.04

### Steps to reproduce

1. Navigate to a Git repo with unstaged changes.
2. `nvim -nu minimal.lua`
3. On a file with unstaged changes, type: `zo`

### Expected behavior

Fold should open with diff, just as when hitting `<Tab>`

### Actual behavior

Fold does open, but without any content.
I have to use `<Tab>` first and then commands such as `zo`, `zc`, `zr` etc. work.

### Minimal config

```Lua
-- NOTE: See the end of this file if you are reporting an issue, etc. Ignore all the "scary" functions up top, those are
-- used for setup and other operations.
local M = {}

local base_root_path = vim.fn.fnamemodify(debug.getinfo(1, "S").source:sub(2), ":p:h") .. "/.min"
function M.root(path)
  return base_root_path .. "/" .. (path or "")
end

function M.load_plugin(plugin_name, plugin_url)
  local package_root = M.root("plugins/")
  local install_destination = package_root .. plugin_name
  vim.opt.runtimepath:append(install_destination)

  if not vim.loop.fs_stat(package_root) then
    vim.fn.mkdir(package_root, "p")
  end

  if not vim.loop.fs_stat(install_destination) then
    print(string.format("> Downloading plugin '%s' to '%s'", plugin_name, install_destination))
    vim.fn.system({
      "git",
      "clone",
      "--depth=1",
      plugin_url,
      install_destination,
    })
    if vim.v.shell_error > 0 then
      error(string.format("> Failed to clone plugin: '%s' in '%s'!", plugin_name, install_destination),
        vim.log.levels.ERROR)
    end
  end
end

---@alias PluginName string The plugin name, will be used as part of the git clone destination
---@alias PluginUrl string The git url at which a plugin is located, can be a path. See https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols for details
---@alias MinPlugins table<PluginName, PluginUrl>

---Do the initial setup. Downloads plugins, ensures the minimal init does not pollute the filesystem by keeping
[...]

PR #1766, opened 2025-06-12
Allow keys that would act on folds (za, zo) to work with lazy loaded folds.

CKolkey:
closes: https://github.com/NeogitOrg/neogit/issues/1765

### DIFF
--- a/lua/neogit/buffers/status/actions.lua
+++ b/lua/neogit/buffers/status/actions.lua
@@ -508,6 +508,25 @@ M.n_toggle = function(self)
 end
 
 ---@param self StatusBuffer
+M.n_open_fold = function(self)
+  return function()
+    local fold = self.buffer.ui:get_fold_under_cursor()
+    if fold then
+      if fold.options.on_open then
+        fold.options.on_open(fold, self.buffer.ui)
+      else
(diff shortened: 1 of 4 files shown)


lua/neogit/buffers/status/actions.lua
```diff
@@ -507,6 +507,25 @@ M.n_toggle = function(self)
   end
 end
 
+---@param self StatusBuffer
+M.n_open_fold = function(self)
+  return function()
+    local fold = self.buffer.ui:get_fold_under_cursor()
```
cassava, line 513: Is this where the lazy loading of the diff is happening?

CKolkey: Yes, it's the `fold.options.on_open(fold, self.buffer.ui)` function that does this.

defined here:
- https://github.com/NeogitOrg/neogit/blob/b3d69c2f5b729cc3574691864c503da6b66e3d79/lua/neogit/buffers/status/ui.lua#L303
- https://github.com/NeogitOrg/neogit/blob/b3d69c2f5b729cc3574691864c503da6b66e3d79/lua/neogit/buffers/status/ui.lua#L212-L243

PR #1770, opened 2025-06-23
Show a warning notification when commit fails

jbradaric:
If a commit isn't successful due to `pre-commit` hook failure, nothing gets reported and the user has to guess what happened.

Ideally, process output should be shown so the user can see what happened. But a notification will at least show that something went wrong.


CKolkey:
You don't see process output? That should work - I see it every time I commit something. Maybe you have it disabled?

jbradaric:
> You don't see process output? That should work - I see it every time I commit something. Maybe you have it disabled?

No, I don't. I just tried using the minimal config and don't get any output or anything (except this notification) when the commit fails.

PR #1792, opened 2025-08-13
fix push other workflow

actionshrimp:
Not sure how widely used this workflow is - it might be that this was working nicely for some people and this PR breaks it, but I suspect that it's just not really used that much and this fixes it.

For me the workflow is (sorry for lack of video rn!):
- You have a branch you've been working on, and the push remote is set
- You've committed a bunch of stuff but you don't really wanna push all of it, just one or two of the earliest commits
- You go for `push`, `other`, then move the selection down to some older commit
- Neogit asks you to confirm the remote branch name just to be sure
- you end up pushing the specific commit of your local branch to the remote.

Happy to clarify anything (or make a video if it would help), and lmk if there is some other workflow that uses this that I'm overlooking!


actionshrimp:
Forgot to mention - first issue I was having was the push didn't actually work because of the `origin/` prefix, and the second issue was much more minor - the inclusion of the local sha which i'd just picked in the selector for the remote branch seemed unecessary.

CKolkey:
I'm still playing around with this, but I think you're probably right.

PR #1795, opened 2025-08-20
Add shortstat to margin popup

lucasadelino:
This is implemented pretty much as in Magit: by calling `git show --format= --shortstat` for each commit. In the Magit source, this is [described](https://github.com/magit/magit/blob/fd1882b8c981c5e859522bde1dd9c88af9485709/lisp/magit-log.el#L1605) as "experimental and rather slow", but it seems to work just fine for medium-sized repos. Largeish repos (e.g. [rust's](https://github.com/rust-lang/rust)) cause some slowdown in Neogit, but I didn't see a significant difference when testing with vs. without this patch.

In order to (mostly) follow Magit's presentation, I had to add an optional argument to a couple of util functions, so that I could pad strings with spaces to the left (rather than to the right). This should be OK, but I'll flag these changes for review just in case.

One thing to note is that the UI assumes the addition and the deletion columns will take up to 5 characters (4 digits and a +/- sign), and the changed files column up to 3 characters. Magit will still try to display the columns if any of them go over those limits, but the alignment breaks. This is also the case for this implementation, although the alignment breaks in a slightly different way (see the screenshot below). At first this bugged me, but I think this is reasonable. Trying to display the data at the possible cost of breaking alignment seems better than the alternatives (clamping numbers; pre-allocating more space than needed; etc), and certainly better than preemptively optimizing around single commits that add/delete 10,000+ lines over 1,000+ files.

shortstat_difference

### DIFF
--- a/lua/neogit/buffers/status/ui.lua
+++ b/lua/neogit/buffers/status/ui.lua
@@ -5,6 +5,7 @@ local common = require("neogit.buffers.common")
 local config = require("neogit.config")
 local a = require("plenary.async")
 local state = require("neogit.lib.state")
+local git = require("neogit.lib.git")
 
 local col = Ui.col
 local row = Ui.row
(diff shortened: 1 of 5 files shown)


lua/neogit/buffers/status/ui.lua
```diff
@@ -5,6 +5,7 @@ local common = require("neogit.buffers.common")
 local config = require("neogit.config")
 local a = require("plenary.async")
 local state = require("neogit.lib.state")
+local git = require("neogit.lib.git")
```
CKolkey, line 8: I'd like to keep all data-fetching out of the view layer. I know some views break this rule, currently, but the eventual goal is to have the views be purely presenters, with all state (repo data to present) fetched up-front and passed into the constructor. 

The compromise for this, if the data is _very_ expensive to fetch, is to set it up to lazy-load via metatable `__index` method. There are a few examples of this scattered around. That keeps knowledge of fetching data out of the view layer, even if it's triggered there.

CKolkey:
Thanks for your patience :)

PR #1807, opened 2025-08-26
Use `--force-with-lease` in push popup

### LINKED ISSUE #1649: When asking about force push it should use --force-with-lease
Recently a feature was added to neogit where if you do not have configured --force flags, it would ask you if you wish to force push if the history is overriden, however i think --force may be too agressive, force-with-lease will be less destructive, if the history in origin has changed, you do not want to push over somebody else's changes.

Use --force-with-lease instead of --force

Alvaro-Kothe:
This PR updates the force push logic in the push popup to use `--force-with-lease` instead of `--force` to avoid overwriting someone else's changes by mistake.

Closes #1649


CKolkey:
Yea, reasonable.

PR #1811, opened 2025-09-04
add 'commit_order' option for margin popup

hammerlink:
PR #1794  introduced a default --topo-order for the margin popup. This makes Neogit’s status view nearly unusable due to the cost of topological sorting on repos with a large commit history (like the kernel).

This PR adds a commit_order config option (default "topo") so users can override or disable the behavior (commit_order = ""). That restores fast refreshes while still allowing topological or date-based ordering when desired.

### DIFF
--- a/lua/neogit/lib/git/log.lua
+++ b/lua/neogit/lib/git/log.lua
@@ -429,16 +429,13 @@ function M.register(meta)
 
     if count > 0 then
       local args = { "--max-count=" .. tostring(count) }
-      local graph = nil;
+      local graph = nil
       if order and order ~= "" then
         table.insert(args, "--" .. order .. "-order")
         graph = {}
       end
 
-      repo_state.recent.items = util.filter_map(
-        M.list(args, graph, {}, false),
-        M.present_commit
-      )
+      repo_state.recent.items = util.filter_map(M.list(args, graph, {}, false), M.present_commit)
     end
   end
 end


CKolkey:
I'm not opposed to this, but when I try opening the kernel in neogit it opens quite quickly. How long does it take you to open?

hammerlink:
This is my workflow:
```sh
git clone git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git neogit_test
cd neogit_test
echo hello > hello & git add hello
neovim
:Neogit
```

It takes me approximately 21 seconds before I can see the staged changes. Any subsequent change like `git rm -f hello` takes 21 seconds again before it is visible on the status page.

CKolkey:
Well, that is objectively terrible 😮‍💨

If this helps you bring that time down, then I'll call that a win.

hammerlink:
@CKolkey thanks!

CKolkey:
> @CKolkey thanks!

No worries! Thank you for the PR :)

lucasadelino:
> This is my workflow:
> 
> ```shell
> git clone git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git neogit_test
> cd neogit_test
> echo hello > hello & git add hello
> neovim
> :Neogit
> ```
> 
> It takes me approximately 21 seconds before I can see the staged changes. Any subsequent change like `git rm -f hello` takes 21 seconds again before it is visible on the status page.

Oof, sorry about the performance hit! Should have thought about allowing a config override. Thanks for adding it!

PR #1813, opened 2025-09-11
Ignore users' configuration `diff.noprefix`

SheffeyG:
Recently when I trying to discard a hunk with neogit, an error raised: `hunk has no filepath`, It turns out somehow I set the `diff.norefix` in my global git configurations, while neogit rely on the prefix `a/` and `b/` for location. so we should alway set `diff.nopreix` to false when building command.


CKolkey:
Thanks! Didn't know that one

PR #1823, opened 2025-09-20
Pass through the search prompt if there were no matches when using Snacks picker.

jordan-turbofish:
This brings the behaviour in line with the Telescope picker: https://github.com/NeogitOrg/neogit/blob/d6112de97160a09c767e2b9dfba45e8942cbd5a4/lua/neogit/lib/finder.lua#L56

I noticed this trying to do a soft reset with HEAD^1 as the argument, as someone else was asking about here: https://github.com/NeogitOrg/neogit/discussions/1669


CKolkey:
Much appreciated!

PR #1826, opened 2025-09-21
Use parent commit for rebase.subset.

jordan-turbofish:
Similar to the interactive rebase case, the parent of the selected commit needs to be used so that the selected commit is in the range of commits being rebased.

Fixes: https://github.com/NeogitOrg/neogit/issues/1812


CKolkey:
Thanks - definitely appreciate you polishing these little details.

PR #1817, opened 2025-09-16
Add 'Open commit' function

CKolkey:
breaking! Add "o" mapping to commit view, which opens the commit's view in web browser. This changes the `config.git_services.x` table format
from:

```lua
{
  ["github.com"] = "<pull request url template>",
  ...
}
```

to:
```lua
{
  ["github.com"] = {
    pull_request = "<pull request url template>",
    commit = "<commit url template>",
    tree = "<tree url template>",
  },
  ...
}
```

Github, gitlab, and bitbucket are supported out-of-the-box.

PR #1829, opened 2025-09-24
docs: update get_services documentation for changes from #1817

Patrick-Beuks:
This files was missed when adding the new open commit feature


CKolkey:
Thank you!

PR #1836, opened 2025-10-04
feat: Codeberg API

foxyseta:
Codeberg uses ForgeJo's API, which is only slightly different from M$ GitH*b's:

https://codeberg.org/api/swagger#/repository

Observe that ForgeJo can be freely selfhosted, so the same config could also scale to arbitrary ForgeJo instances in the future.

I ran `make test`, `make lint`, and `stylua .`.

I tested the "open pull request" feature, but I could not find the commands to open the commit/branch tree on the web, despite reading the README. Can somebody help me test those two out, too, before the merge, please?


CKolkey:
To open the tree, its 'o' from the status buffer (main neogit buffer) and to open the commit, its 'o' from the commit view. I suppose that could be better documented.

foxyseta:
Thanks! I now tested all three.

Btw, typos detected two mispellings in code other than mine:

```
typos
error: `indx` should be `index`
   ╭▸ ./lua/neogit/lib/git/stash.lua:96:20
   │
96 │       assert(idx, "indx cannot be nil")
   ╰╴                   ━━━━
error: `connetors` should be `connectors`
     ╭▸ ./lua/neogit/lib/graph/kitty.lua:1001:55
     │
1001 │     -- now lets get the intervals between the stopped connetors
     ╰╴                                                      ━━━━━━━━━
```

foxyseta:
And one thing I noticed in the doc which is not necessarily a typo, but surely a grammar error, is `it's` instead of `its` for the possessive adjective.

CKolkey:
Thanks!

foxyseta:
No problem!

> Btw, typos detected two mispellings in code other than mine:

(This was fixed by 0f48491ae0046796841aaa97d439267982fe72db.)

> And one thing I noticed in the doc which is not necessarily a typo, but surely a grammar error, is it's instead of its for the possessive adjective.

(This is a separate issue, so I opened #1848.)

CKolkey:
Much appreciated 😉

PR #1841, opened 2025-10-11
fix: fixed auto_close_console not working for pre-commit hooks

benjasper:
Somehow the closing of the "Neogit console" float recently stopped working in the context of running pre-commit hooks, preventing me from seeing the commit buffer (I am using a split for the commit view).

The problem can be easily reproduced by putting something long running with output into the commit-hook that returns fine with exit code 0. And then try to commit.
For Example
```bash
for i in {1..10}; do
    echo "Waiting $i seconds..."
    sleep 1
done
```

I am not really sure as to the root cause of this issue, because i am very sure that this worked at some point, but couldn't find a commit where this worked.

If you have any input or would like me to make changes let me know!

### DIFF
--- a/lua/neogit/client.lua
+++ b/lua/neogit/client.lua
@@ -77,6 +77,10 @@ function M.editor(target, client, show_diff)
   logger.debug(("[CLIENT] Invoked editor with target: %s, from: %s"):format(target, client))
   require("neogit.process").hide_preview_buffers()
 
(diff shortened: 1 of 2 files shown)


benjasper:
Seems like my fix is kind of flaky, I'm closing this as it's not a great fix. I'll use the split_above option for now to workaround the issue that the console is overlaying the commit editor.

PR #1857, opened 2025-11-09
Feature: fast commits for large repositories

benjiwolff:
I am currently working exclusively with a large codebase. The main issue is that git status on the entire repo takes around 2 seconds which makes staging and committing super tedious and frustrating.

This implementation removes 2 `git status` calls before the commit message editor is shown.

The solution is not quite as pretty, so I added a configuration option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,3 @@
-<div align="center" markdown="1">
-   <br>
-   <br>
-   <a href="https://www.warp.dev/neogit">
-      <img alt="Warp sponsorship" width="600" src="https://github.com/user-attachments/assets/c58acc85-7438-46a7-a89a-0f404c269256">
-   </a>
-
-### [Warp, the intelligent terminal for developers](https://www.warp.dev/neogit)
-#### [Try running neogit in Warp](https://www.warp.dev/neogit)<br>
-
-</div>
-
-<hr>
-
 <div align="center">
     <div>
         <div><img src="https://github.com/NeogitOrg/neogit/assets/7228095/7684545f-47b5-40e2-aedd-ccf56e0553f4" width="400px"/></div>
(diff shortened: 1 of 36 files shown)


lua/neogit/popups/commit/actions.lua
```diff
@@ -105,9 +110,11 @@ local function commit_special(popup, method, opts)
 end
 
 function M.commit(popup)
-  if not git.status.anything_staged() and not allow_empty(popup) then
-    notification.warn("No changes to commit.")
-    return
+  if not config.values.commit_editor.fast then
```
CKolkey: We could also modify the `git.status.anything_staged()` implementation to check `git.repo.state.staged.items[1] ~= nil` if "fast" is set to true. Then we don't need to skip this check, but instead just rely on the repo state data already in memory.

CKolkey: We could also put `not allow_empty(popup)` _before_ `git.status.anything_changed()`, so you could short-circuit the conditional by enabling "--allow-empty".

lua/neogit/popups/commit/actions.lua
```diff
@@ -33,13 +33,18 @@ local function confirm_modifications()
 end
 
 local function do_commit(popup, cmd)
-  client.wrap(cmd.arg_list(popup:get_arguments()), {
+  cmd.arg_list(popup:get_arguments())
```
CKolkey: Have you tried just adding an option for `-m` to the commit popup?
You could do `popup:option("m", "message", "", "Commit message", { key_prefix = "-" })` and that also just bypasses the editor.	

https://github.com/NeogitOrg/neogit/blob/d2a2ae4415872fbddb2441920ab109ee52fd4916/doc/neogit.txt#L2267-L2283

CKolkey (commented):
I think there might be ways to accomplish what you are trying to do without needing to modify much. Do you mind trying my suggestions and letting me know if they work for your use case?

PR #1861, opened 2025-11-17
Use correct destination branch in pushing another branch

swnakamura:
Fixes a problem that happens when pushing a branch that is not the current branch to the remote by `P` `o` command.
Previously, for example, when pushing a branch named `master` to `origin/master`, a branch called `origin/origin/master` is created on the remote. This PR fixes this.


CKolkey:
thanks :)

PR #1866, opened 2025-11-26
improve: Add config option to change <cr> over hunks on commit_view

txemaotero:
- New `commit_view.cr_on_hunk_jumps_to_current` option to interchange the behaviour of `<cr>` with the mapping for the `OpenFileInWorktree` command.
- This allows to set `<cr>` as the mapping to jump to the current version of the file.
- Defaulted to false, i.e., preserve existing behavior (same as magit).

### DIFF
--- a/doc/neogit.txt
+++ b/doc/neogit.txt
@@ -392,6 +392,7 @@ The following mappings can all be customized via the setup function.
       }
 
       commit_view = {
+        ["<cr>"]       = "OpenFileInCommit",
         ["a"]          = "OpenFileInWorktree",
       }
 
@@ -2187,10 +2188,13 @@ Mappings (normal mode):
                  you to a wrong location in the file if the lines have been
                  moved around.
 
-  • `<cr>`       On a diff hunk, open the file contents from the commit in a
-                 temporary read-only tab, or the parent revision for deleted
-                 lines; on a filepath line, jump to that file's diff section
-                 within the buffer.
+  • `<cr>`       (|neogit_setup_mappings|: "OpenFileInCommit" in commit_view)
+                 When the cursor is on a diff hunk, opens the version of the
+                 file as it exists in the commit in a temporary read-only
+                 buffer (or the parent revision if the lines were deleted).
+
+  • `<cr>`       On a filepath line, jump to that file's diff section within
+                 the buffer.
 
   • `o`          Open the commit in the configured git service (requires
                  Neovim >= 0.10 for |vim.ui.open|).
--- a/lua/neogit/buffers/commit_view/init.lua
+++ b/lua/neogit/buffers/commit_view/init.lua
@@ -219,67 +219,45 @@ local function is_jumpable_hunk_line_component(c)
     or c.options.line_hl == "NeogitDiffDelete"
 end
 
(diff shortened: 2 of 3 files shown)


txemaotero:
I see tests failed but due to unrelated changes. I've checked locally that they also fail with the code in master.

txemaotero:
Failure in unit tests introduced [here](https://github.com/txemaotero/neogit/commit/911b10617c1f7260ec9aebfeb6d1c6965a1d2830). Proposing a fix in this [PR](https://github.com/NeogitOrg/neogit/pull/1867).

txemaotero:
I don't think the test failure is related. It seems an error due to a tight timing:
```
       Expected [" Stashes (1)                                                                    ", "stash@{0} On master: test                                           1 second ago"]
          to eq [" Stashes (1)                                                                    ", "stash@{0} On master: test                                          0 seconds ago"]
```

CKolkey:
Yea, the e2e tests are a bit weird. They run way more reliably locally than on CI, and I don't know why... and debugging that feels more like "work" than "fun". 

Anyways, instead of boolean configs like this, I'd rather solve it with mappings more like how the status buffer is set up: `{ "key" = "ActionNane" }` 

I think thats more flexible - having a load of boolean configs seems like a road to madness - as well as consistent.

txemaotero:
> I think thats more flexible - having a load of boolean configs seems like a road to madness - as well as consistent.

I completely agree with that. The thing is that I discarded initially that approach because I was reusing a mapping for the "open file in commit" functionality (`<cr>`, if hit on a file path line, jumps to that section in the commit view). This is similar to what is done with the `GoToFile` mapping in the status view (depending where you hit `<cr>` you can open the commit view or jump to the file). However, for this use case, I didn't want to define a new map that overrides both the "open the file" and the "jump to the hunks section" behaviors. As I didn't see anything in the code doing this kind of decoupling, I decided to go with the boolean approach to minimize code changes.

That said, after your feedback, I decided to give it a try and implement a logic to allow clashing mappings that target different components in the commit view. With that infrastructure, we can now define a new mapping for "OpenFileInCommit" that can be set to `<cr>`. Let me explain what I did to facilitate the review:
- Declare `mappings` before assigning to the buffer creation options
- Remove from that declaration the definition of mappings for `<cr>` and `commit_view_maps["OpenFileInWorktree"]`.
- Attach those removed mappings via the `attach_jump_mappings` function. This function is the one that handles the combination of multiple mappings with the same key but with different actions, each targeting different components in the view. In other words, we can attach one behavior to `<cr>` if the cursor is on a file path and a different one if we are on a hunk. 

If you think this is too complicated just let me know and we can discuss a better way :smile:

txemaotero:
Hi @CKolkey , any thoughts on last version? (Asking JIC it slipped through your inbox 😅)

CKolkey:
Hey, apologies, i dont have a ton of free time right now - three kids, two are 1 year old twins 😅

Ill try to make time as soon as i can :)

txemaotero:
No hurries! I have an idea of what you're going through 😉 it was a reminder JIC, take your time an thanks! 😊

PR #1867, opened 2025-11-26
Fix: fallback to OS check in newline character determination

txemaotero:
- This prevents errors when the path passed to eol_character doesn't exists
- Fixes unit tests


CKolkey:
Thanks :)

Issue #1822, opened 2025-09-18
Better support for git submodules


rafaelsmgomes:
**Is your feature request related to a problem? Please describe.**
Not exactly. When working with submodules, Neogit tells me I have commits under the submodule (points to a new hash). That is good, means I have to commit and push the changes to the submodule.

The problem is then: how do I do that?

Well, I would think that pressing enter on that submodule would drop me into the submodule so that I can review and commit those changes (like it happens on lazygit). But that doens't work. So what I have to do is to quit neovim (or open a new pane), navigate to that submodule directory, open neogit, and now I can commit those changes. 

That is not a scalable solution. Specially when I have a lot of submodules. I like how lazygit does it. I can just press enter and it will take me to the submodule's git. I can then see and review all those changes. 

Personally, I prefer Neogit's UI and developer experience. But I just have to use lazygit when working with submodules because of this. 

It would be even greater if neogit would automatically detect the submodules structure and show them on the top, so I could easily change to and from that git directory. 

**Describe alternatives you've considered**

I haven't thought much about other ways of doing it. 

**Additional context**
Sure. Here is what it shows me when there is a change in the submodule:

<img width="439" height="162" alt="Image" src="https://github.com/user-attachments/assets/cf7f17d8-91ba-4122-9c39-46c8ec839f54" />

and then I press enter on it, it navigates me to the oil directory (that's nice):

<img width="403" height="295" alt="Image" src="https://github.com/user-attachments/assets/d81c059f-1d96-4eea-9b44-db5e123a678c" />

But I can't really do anything with this, since if I call neogit again, I'll just get the same screenshot above (the initial neogit status).

txemaotero:
Hi! I also found this feature a very "nice to have" so I give it a try and propose a basic implementation in a PR. This initial solution aligns with what you described:
- Hitting enter on an entry in the status view referencing to a submodule will open a new status view with the submodule as root.
- You can then do your stuffs (commit, revert...)
- Once you're done, hit "gp" to jump back to the parent directory and commit (or whatever) the new state of the submodule

In my head, once we have this, we could extend the functionality to:
- Have a dedicated section for submodules in the status view
- Define some new mappings to trigger submodule updates, init... That selectively apply if the cursor is over an entry in the submodule section.

But, first, let's see if this initial proposal is a good enough starting point 😉

PR #1875, opened 2025-12-09
feat: Basic submodule support

txemaotero:
This PR aims to provide some support to work with submodules. It's motivated by the issue #1822 

New functionality:
- When hitting `<cr>` on a submodule entry with changes in the status view, a new neogit instance is opened with the submodule as root.
- When inside the status view of a submodule, hitting `gp` (default for the new `GoToParentRepo` mapping) opens a new neogit instance with the parent repo as root.

Implementation details:
- On new neogit instance creation, information about submodules and parent repo is cached.
- Helper functions are defined to get the submodule under the cursor or the parent repo if present.
- New git commands were added to:
    - Get a list of the submodules in the repo (in the new `git/submodule.lua` file)
    - Get the path to the parent repo (as a new flag for the `rev_parse` command)
- The feature was validated manually but a new integration test was also introduced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,6 +482,7 @@ neogit.setup {
       ["y"] = "ShowRefs",
       ["$"] = "CommandHistory",
       ["Y"] = "YankSelected",
+      ["gp"] = "GoToParentRepo",
       ["<c-r>"] = "RefreshBuffer",
       ["<cr>"] = "GoToFile",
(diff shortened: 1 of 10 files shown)


CKolkey:
Great stuff! Small thing re: syntax preference but otherwise fine 👍🏼

Issue #1620, opened 2025-01-04
Autosquash (`r f`) destroys commits instead of squashing


loqusion:
### Description

[Autosquash] functionality (default keybind: `r f`) in Neogit does not seem to work as intended. Instead of squashing commits starting with `fixup! `/`squash! `/`amend! ` into the commits they target, it resets the current branch to the selected commit.

[Autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

### Neovim version

NVIM v0.10.3
Build type: RelWithDebInfo
LuaJIT 2.1.1731601260

### Operating system and version

Arch Linux (`6.12.7.zen1-1`)

### Steps to reproduce

1. Clone <https://github.com/loqusion/neogit-autosquash-bug-repro>
2. `cd` to the local repo, open Neovim (`nvim -u minimal.lua`), and type `:Neogit<CR>`
3. Press `<Tab>` on **Recent Commits** and hover over the last commit with the message `initial commit` (alternatively, hover over the commit in the log view)
4. Press `r f`

### Expected behavior

Magit for Emacs runs the command `git … rebase -i --root --autosquash --keep-empty --autostash`, triggering a merge conflict as expected in this particular case.

The general case where there are no merge conflicts also works as expected: the fixup commit is squashed into the commit it targets.

![magit-behavior](https://github.com/user-attachments/assets/60993469-b177-4118-91d4-79cb2ad34cc0)

### Actual behavior

Neogit runs the command `git rebase --onto 7313e5cd4f7b9c2723e011f233bf88e44107e8d7 HEAD --autostash --autosquash --keep-empty`. The current branch now points to `7313e5c` (`initial commit`).

This behavior does not depend on the target commit being the root commit: performing the same operation on `second commit` and renaming the fixup commit to `fixup! second commit` similarly results in the current branch pointing to `second commit` without squashing the fixup commit.

> [!NOTE]
> I've also reproduced the bug with `minimal.lua` off camera.

![neogit-autosquash-bug](https://github.com/user-attachments/assets/b51b8396-3aaf-46d4-8d4e-78dac29f02fe)

### Minimal config

```Lua
-- NOTE: See the end of this file if you are reporting an issue, etc. Ignore all the "scary" functions up top, those are
-- used for setup and other operations.
local M = {}

local base_root_path = vim.fn.fnamemodify(debug.getinfo(1, "S").source:sub(2), ":p:h") .. "/.min"
function M.root(path)
[...]

PR #1880, opened 2025-12-17
Fix: autosquash command (`r f`) bug is fixed.

swnakamura:
When `r f` is pressed, obtain the parent of currently specified commit as new base and run a interactive (autosquashing) rebase onto the commit.
Currently, there's a bug that deletes the rebased commits (#1620). This PR fixes it.

### DIFF
--- a/lua/neogit/popups/rebase/actions.lua
+++ b/lua/neogit/popups/rebase/actions.lua
@@ -164,19 +164,22 @@ function M.edit()
 end
 
 function M.autosquash(popup)
+  local args = util.deduplicate(util.merge(popup:get_arguments(), { "--autosquash", "--keep-empty" }))
   local base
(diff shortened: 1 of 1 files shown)


lua/neogit/popups/rebase/actions.lua
```diff
@@ -164,19 +164,22 @@ function M.edit()
...
     base = git.rebase.merge_base_HEAD()
   end
 
   if base then
-    git.rebase.onto(
-      "HEAD",
-      base,
-      util.deduplicate(util.merge(popup:get_arguments(), { "--autosquash", "--keep-empty" }))
-    )
+    git.rebase.rebase_interactive(base, args)
```
disrupted: is it even necessary to bring up the interactive editor view for this?

I think it would be even more convenient if this used the `git.rebase.instantly` path (similar to instant fixup).

PR #1883, opened 2025-12-26
feat: add codediff.nvim as alternative diff viewer

sotte:
This is a draft implementation against the **next branch of vscode-diff.nvim**.

Note that I'm not using all the features of diffview and therefore vscode-diff.nvim, so the implementation is not going to be 100% correct :)

I think we should wait till **next** is merged before proceeding with this. But it gets the discussion started.

----

Add support for vscode-diff.nvim alongside diffview.nvim for viewing diffs. Users can explicitly choose via the new `diff_viewer` config option, or let Neogit auto-detect (diffview first, then vscode-diff).

Why:

- Provides choice for users who prefer vscode-diff's interface
- Maintains backwards compatibility (diffview remains default)
- Auto-detection means zero config for most users

Implementation:

- New `diff_viewer` config option: "diffview", "vscode_diff", or nil
- New integration module at integrations/vscode-diff.lua with same interface as diffview (M.open with section_name, item_name, opts)
- get_diff_viewer() helper handles explicit choice vs auto-detection
- Diff popup and status buffer conflict resolution dispatch to the configured viewer
- on_close callback for auto-staging after conflict resolution works with both viewers using the same BufEnter autocmd pattern

### DIFF
--- a/lua/neogit/integrations/codediff.lua
+++ b/lua/neogit/integrations/codediff.lua
@@ -72,7 +72,7 @@ local function make_explorer_data(status_result, focus_file)
 end
 
 local function open_explorer(view, git_root, status_result, original_revision, modified_revision, focus_file)
-  ---@type SessionConfig
+  ---@type table
   local session_config = {
     mode = "explorer",
     git_root = git_root,
@@ -341,7 +341,7 @@ function M.open(section_name, item_name, opts)
       local filetype = vim.filetype.match { filename = file_path } or ""
       local original_revision, modified_revision = get_conflict_revisions()
 
-      ---@type SessionConfig
+      ---@type table
       local session_config = {
         mode = "standalone",
         git_root = git_root,
--- a/lua/neogit/popups/diff/init.lua
+++ b/lua/neogit/popups/diff/init.lua
@@ -6,23 +6,29 @@ local actions = require("neogit.popups.diff.actions")
 
 function M.create(env)
   local diff_viewer = config.get_diff_viewer()
-  local commit_selected = (env.section and env.section.name == "log") and type(env.item.name) == "string"
+  local has_diff_viewer = diff_viewer ~= nil
+  local has_item = env.item ~= nil
+  local commit_selected = env.section
+      and env.section.name == "log"
+      and env.item
+      and type(env.item.name) == "string"
+    or false
 
   local p = popup
     .builder()
     :name("NeogitDiffPopup")
     :group_heading("Diff")
-    :action_if(diff_viewer and env.item, "d", "this", actions.this)
-    :action_if(diff_viewer and commit_selected, "h", "this..HEAD", actions.this_to_HEAD)
-    :action_if(diff_viewer, "r", "range", actions.range)
+    :action_if(has_diff_viewer and has_item, "d", "this", actions.this)
+    :action_if(has_diff_viewer and commit_selected, "h", "this..HEAD", actions.this_to_HEAD)
+    :action_if(has_diff_viewer, "r", "range", actions.range)
     :action("p", "paths")
     :new_action_group()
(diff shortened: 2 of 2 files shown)


DanRioDev:
YES! YES! YES!

georgeharker:
It would be great to see this merged.

IanLiuTW:
This is really good. I saw the author of the vscode-diff released a newer version of the same project. It would be a good idea to migrate the support to V2. Thanks again for the work!

sotte:
@CKolkey already switched to the new version/fork.

Everybody, please test this. Any feedback is welcome.

Knaifhogg:
Some commits when I diff (double tap d in NeogitStatus view) result in the left file not loading text and sometimes also getting an error saying "Invalid codediff URL". See it happen on a single-file commit and 5-file commit. Seems to be consistent for specific commits for some reason, both in stashes and the short commit list. Others work perfectly. I can't spot a pattern for this.

sotte:
Thanks for the feedback! My pr was based on the next branch. I assume some things have changed since the plug-in was moved into a new repo and next was merged. 

I have a few minutes to investigate today.

chek1337:
When is merge expected?

DanRioDev:
Bring it on @sotte

stevenxxiu:
@DanRioDev Please just upvote instead of notifying everyone subscribed to the issue.

sotte:
I just integrated the newest version of `codediff.nvim`. I also rebased on the current `master`.  It works for me :tm: 

Please try it and give feedback as I don't use all the features of neogit/diffview/codediff, just the `dd` and `dw` feature really.

Wikiwix:
Thanks for the effort!

Seems to work for me too.
Though I ran into (non reproducible) issues the first time I tried to diff in the log view (e.g. `ll d`) (the windows opened, but were blank. Once I put my cursor in the relevant windows, the diff appeared. After that it just worked. (I tried `dd` before that worked from the beginning)

---

Not a demand by any means (and probably a later PR anyways):
I would really like to see CodeDiff to be used for the diff rendering on the "Status page". [This part of their docs](<https://github.com/esmuellert/codediff.nvim#:~:text=%2D%2D%20Example%201%3A%20Compute%20diff%20between%20two%20sets%20of%20lines%0Alocal%20lines_a%20%3D%20%7B%22line%201%22%2C%20%22line%202%22%7D%0Alocal%20lines_b%20%3D%20%7B%22line%201%22%2C%20%22modified%20line%202%22%7D%0Alocal%20lines_diff%20%3D%20diff.compute_diff(lines_a%2C%20lines_b)>) makes it look like their API might perhaps allow that.

#### Edit
The "quoted" API apparently moved a little, though I was successfully able to apply a diff to only a part of a buffer with this code below (and two buffers with the matching ids and containing amongst other things the noted texts):

```lua
local core = require("codediff.ui.core")
local diff =  require("codediff.core.diff")
local lines_a = {"line 1", ""}
local lines_b = {"line 1", "modified lin 2"}
local lines_diff = diff.compute_diff(lines_a, lines_b)
core.render_diff(276, 253, lines_a, lines_b, lines_diff)
```

Not sure how practical using this would actually be though.

---

Another location where it would make sense are the diffs shown on the commit details (e.g. `ll<enter>`)

sotte:
Thanks for testing, @Wikiwix .

I'll address the lld issue. Maybe just a focus issue. 

The other topic is IMO out of scope of this pr, but should be addressed at some point. And it seems you're half way there already.

CKolkey:
Tested, and this works well enough to merge. If anyone finds issues... open an issue. :)

Thanks @sotte !

sotte:
Thanks for merging @CKolkey .

PR #1891, opened 2026-01-27
feat: add prompt_amend_commit option

alixinne:
This adds an option to disable the confirmation prompt when amending a commit that has already been pushed. This can be useful for users that regularly amend commits while working on a branch by themselves.


seowalex:
Would love to see this as well.

PR #1893, opened 2026-02-05
fix: prevent `--literal-pathspecs` leaking into git hooks

disrupted:
Neogit currently adds `--literal-pathspecs` to all git commands, which causes git to export `GIT_LITERAL_PATHSPECS=1` to hooks. Hook managers such as [hk](https://hk.jdx.dev/) can fail when running `git stash`.

e.g. git error:

```
pathspec ':/' did not match any file(s) known to git
```

in this case commits abort with `Commit failed` before the message editor opens.

This change skips `--literal-pathspecs` for hook-triggering subcommands (commit/merge/rebase/checkout/push), preventing the env leak while keeping the flag for other commands.

### DIFF
--- a/lua/neogit/lib/git/cli.lua
+++ b/lua/neogit/lib/git/cli.lua
@@ -1205,6 +1205,7 @@ local function new_builder(subcommand)
       table.insert(cmd, "-")
     end
 
+    -- stylua: ignore
     local base = {
       get_git_executable(),
       "--no-pager",


disrupted:
@CKolkey what do you think of this? I am happy to tweak it or add tests if you give me feedback.
This is currently blocking our transition from pre-commit to https://hk.jdx.dev 🥲

lua/neogit/lib/git/cli.lua
```diff
@@ -1189,20 +1205,21 @@ local function new_builder(subcommand)
...
-        "--literal-pathspecs",
-        "--no-optional-locks",
-        "-c", "core.preloadindex=true",
-        "-c", "color.ui=always",
-        "-c", "diff.noprefix=false",
-        subcommand
-      },
-      cmd
-    )
+    local base = {
```
CKolkey: gotta add back the "-- stylua: ignore" comment here for the formatting to be nicer.

CKolkey:
"checkout" might be the only one to cause issue, since you can checkout a specific file, but the others I can't imagine why you would add a path to.

disrupted:
> "checkout" might be the only one to cause issue, since you can checkout a specific file, but the others I can't imagine why you would add a path to.

do you mean manually running `git checkout` using keybind `Q`? tbh I had to look that up because I've never used it within Neogit

CKolkey:
> > "checkout" might be the only one to cause issue, since you can checkout a specific file, but the others I can't imagine why you would add a path to.
> 
> do you mean manually running `git checkout` using keybind `Q`? tbh I had to look that up because I've never used it within Neogit

No, not manually, the normal workflow. But my guess is it'll only be an issue if someone has weird glob-like characters in the filepath.

PR #1930, opened 2026-03-26
fix : ambiguous description for branch options (2 options with 'new b…

Soufi-ane:


CKolkey:
I get what you're saying, but the "Checkout" header applies to the first two columns. The ui philosophy is that the header applies until a new one is declared, so I'm going to leave it as-is. Thanks though :)

PR #1950, opened 2026-04-23
Set default red color based on ErrorMsg

p-ouellette:
The previously used Error highlight group does not actually have a red foreground in the default Neovim color scheme. It has a red background and white foreground. The ErrorMsg group is probably more likely to have a red foreground in most color schemes.


CKolkey:
Sensible, thanks

PR #1951, opened 2026-04-23
Fix crash when NeogitConsole buffer is reused across git operations 

### LINKED ISSUE #1896: Crash: "Terminal already connected to buffer" when opening process console
### Problem Description
I encountered a crash when opening the Neogit process console (e.g., via `$`). The error message is:
`Error ... msg_show.lua_error ... Terminal already connected to buffer ...`

### Environment
- OS: MacOS (darwin)
- Neogit version: (latest from lazy)

### Stack Trace
```
Error ... msg_show.lua_error vim.schedule callback: .../neogit/lib/buffer.lua:557: Terminal already connected to buffer 252
stack traceback:
[C]: in function 'nvim_open_term'
.../neogit/lib/buffer.lua:557: in function 'open_terminal_channel'
.../neogit/buffers/process/init.lua:111: in function 'after'
.../neogit/lib/buffer.lua:829: in function 'f'
...
.../neogit/buffers/process/init.lua:103: in function 'open'
```

### Analysis
The issue appears to happen when the `NeogitConsole` buffer exists but is in a "stuck" state where `nvim_open_term` cannot be called on it again (likely because it's already a terminal buffer from a previous session or zombie process), but `Buffer.create` attempts to reuse it.

### Proposed Fix
I was able to resolve this locally by ensuring any stale `NeogitConsole` buffer is cleaned up before creating a new `Buffer`.

Patch for `lua/neogit/buffers/process/init.lua` in `M:open()`:

```lua
  function M:open()
    if self.buffer then
      return self
    end

    local status_maps = config.get_reversed_status_maps()

    -- FIX: Check for stale terminal buffer and delete it
    local existing = vim.fn.bufnr("NeogitConsole")
    if existing ~= -1 then
      local buftype = vim.api.nvim_get_option_value("buftype", { buf = existing })
      if buftype == "terminal" then
        vim.api.nvim_buf_delete(existing, { force = true })
      end
    end

    self.buffer = Buffer.create {
      name = "NeogitConsole",
      -- ...
```

ktsujister:
Fixes NeogitOrg/neogit#1896

## Changes
- Track the active terminal channel in a module-level variable and close it before calling `nvim_open_term` on reuse
- Skip setting `buftype` when reusing an existing terminal buffer, since Neovim does not allow changing `buftype` on terminal buffers                                                                                                                                                                                                                                                                                                       

## Testing 
When Testing this, I have a repository with git pre-push hook that takes a while to run.
- Running `git push` opens `Git Console`
- Close floating buffer when that's done, or press `q` to just close.
- Re-push the same commit to try to re-use `NeogitConsole`.

### DIFF
--- a/lua/neogit/buffers/process/init.lua
+++ b/lua/neogit/buffers/process/init.lua
@@ -9,6 +9,10 @@ local config = require("neogit.config")
 local M = {}
 M.__index = M
 
+-- Tracks the terminal channel of the current NeogitConsole buffer so it can be
+-- closed before the buffer is reused for a new process.
+local current_chan = nil
+
 ---@param process Process
 ---@param mask_fn fun(string):string
 ---@return ProcessBuffer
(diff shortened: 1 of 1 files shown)


CKolkey:
ugh, I apparently have some CI issues to clean up. 

Thanks for the PR :D

PR #1962, opened 2026-05-27
Add opt-in `stream_hook_output` to show git hook output live (e.g. post-checkout migrations)

jcarlos7121:
## What's this?

A small, **opt-in** config flag — `stream_hook_output` (default `false`) — that lets you actually *watch* a git hook's output stream into the Neogit console while it runs.

My setup runs a `post-checkout` hook (the `hookup` Ruby gem) that runs DB migrations whenever I switch branches. Today that hook fires on every checkout but basically gets lost currently. The output is captured, but it gets swallowed silently. 

## Why it doesn't already work

Turns out checkout is the **only** hook-bearing op that runs *blocking* (`branch.checkout` → `.call { await = true }`). A blocking `jobwait` freezes redraws, and the `vim.schedule`'d "show console" timer can't fire mid-run — by the time it could, the result is already set and it bails, then auto-close finishes it off on success. So the console never gets a chance to stream. Meanwhile `merge`/`push` run async and *already* stream their hook output via the existing 800ms timer — checkout was just the odd one out.

## What the flag does

When `stream_hook_output = true`:

- `branch.checkout`/`track` run **async instead of blocking** — but only when a `post-checkout` hook actually exists — so the existing console machinery can stream the output live as it's produced.
- A tiny tweak in `process.lua`'s timer force-shows the console for hook processes, so it works even if you've turned `auto_show_console` off.
- **Close-on-success / stay-open-on-failure are unchanged** — those already do exactly the right thing (`auto_close_console` + the existing hook-failure branch). So you watch it run, it tidies up after itself on success, and it sticks around so you can read the error if the hook fails.

Default is `false`, so this is **zero behavior change** unless you opt in.

## Scope / notes

- Applies to the non-suppressed hook ops (checkout, merge, push). `rebase`/`bisect` run with `long = true` (console suppressed), so they're intentionally out of scope.
- Switching checkout to async is safe: popup actions already run inside an `a.void` coroutine, so the async path yields and resumes with the `ProcessResult` exactly like before, and the runner falls back to blocking if it's ever called outside a coroutine.

## Tests

- New `branch_checkout_await_spec.lua` asserting checkout takes the async path only when the flag + a `post-checkout` hook are present (and blocks otherwise).
- New config-validation case in `config_spec.lua`.
- Full suite green, `stylua` clean. Documented in README + `doc/neogit.txt`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,9 @@ neogit.setup {
   auto_show_console = true,
   -- Automatically close the console if the process exits with a 0 (success) status
   auto_close_console = true,
+  -- Stream git-hook output (e.g. post-checkout) to the console live as it runs.
+  -- Closes on success, stays open on failure. Off by default. (Excludes rebase/bisect.)
+  stream_hook_output = false,
   notification_icon = "󰊢",
   status = {
     show_head_commit_hash = true,
--- a/doc/neogit.txt
+++ b/doc/neogit.txt
@@ -206,6 +206,9 @@ to Neovim users.
     auto_show_console = true,
     -- Automatically close the console if the process exits with a 0 (success) status
     auto_close_console = true,
+    -- Stream git-hook output (e.g. post-checkout) to the console live as it runs.
+    -- Closes on success, stays open on failure. Off by default. (Excludes rebase/bisect.)
+    stream_hook_output = false,
     notification_icon = "󰊢",
     status = {
         show_head_commit_hash = true,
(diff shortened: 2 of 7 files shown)


3dyuval:
Exactly what I've been wanting. Commitlint errors gets swallowed is pain.

CKolkey:
I can't see why checkout/track should be blocking. If you remove the `await = true` entirely, does that provide the behavior you want?

PR #1971, opened 2026-06-16
Fix extraneous empty final line diff in diffview integration

micampe:
When using "Diff this" or "Diff unstaged", the diffview integration always shows an extra final line diff, even when there is no change.

This is a limitation in diffview and vim diff: the final newline is lost in the actual text after it's been loaded in a buffer.

This change strips the final empty line, which is also what diffview does when it reads git commits on its own, resulting in the diff looking correct.

### DIFF
--- a/lua/neogit/integrations/diffview.lua
+++ b/lua/neogit/integrations/diffview.lua
@@ -10,6 +10,17 @@ local Watcher = require("neogit.watcher")
 local git = require("neogit.lib.git")
 local a = require("neogit.lib.async")
 
+-- `git show` represents the file's final newline as an empty string "".
+-- diffview compares line arrays, and its other side is a buffer where the
+-- trailing newline is stripped with no final empty line.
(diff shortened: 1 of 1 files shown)


micampe:
I'm not sure I explained the issue clearly so I'm showing an example here.

This is a file diff:

```diff
diff --git i/Makefile w/Makefile
index 24e70b13..d3be66e5 100644
--- i/Makefile
+++ w/Makefile
@@ -1,9 +1,6 @@
 test:
 	TEMP_DIR=$$TEMP_DIR TEST_FILES=$$TEST_FILES GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null NVIM_APPNAME=neogit-test nvim --headless -S "./tests/init.lua"
 
-specs:
-	bundle install && CI=1 bundle exec rspec --format Fuubar
-
 lint:
 	selene --config selene/config.toml lua
 	typos
```

This is what is being shown when using "Diff this" from Neogit:

image

PR #1987, opened 2026-07-11
CommitView: '}' mapping goes to first hunk if not in one already

washtubs:
I was frustrated that in the commit view, the "Goto next" `}` mapping is inert if your cursor is at the top.

The current behavior is that it checks if you are hovering a hunk or diff component, otherwise it does nothing. Now I made it so it searches down to the first hunk or diff component and places the cursor there.


lua/neogit/lib/ui/init.lua
```diff
@@ -66,7 +66,7 @@ end
 ---@param f fun(c: UiComponent): boolean
 ---@param options FindOptions|nil
 function Ui:find_component(f, options)
-  return Ui._find_component(self.layout, f, options or {})
+  return Ui._find_component(self.layout.children, f, options or {})
```
washtubs, line 69: `find_component` seemed to be unused. It was not working because the layout doesn't have ipairs as `_find_component` expects. `layout.children` does.

PR #1988, opened 2026-07-13
Allow commit hunk jumping from the metadata section

micampe:
This change allows pressing } when opening a commit to jump directly to the first hunk from the metadata section at the top. Previously the {/} keymaps only worked when the cursor was already within the diff portion of the commit.

Also, { jumps back to the metadata section from the first hunk.

This behavior matches how Previous/NextSection work in the status view.

Refactored the } implementation to mirror how { is defined for clarity.

### DIFF
--- a/lua/neogit/buffers/commit_view/init.lua
+++ b/lua/neogit/buffers/commit_view/init.lua
@@ -343,6 +343,8 @@ function M:open(kind)
               end
 
               return first
+            else
+              return 1
             end
           end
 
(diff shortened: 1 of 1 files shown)


micampe:
oops, I didn't notice #1987 had already been submitted... I guess you'll just have to pick which implementation you prefer :D

if that one is preferred I think `{` also needs to be fixed (trivial, can use that portion of my patch).
I would update both keymaps to use `find_component` but that's just stylistic.

PR #1993, opened 2026-07-22
feat: cleanup buffers after loading a session

igorlfs:
Context: https://github.com/neovim/neovim/issues/40336

TLDR: Neogit leaves leftover buffers when restoring vim sessions. We could have a "full restore" functionality, but that's somewhat complex. Instead, just wipe the leftover buffers, which is a low-cost improvement.

### DIFF
--- a/lua/neogit/autocmds.lua
+++ b/lua/neogit/autocmds.lua
@@ -18,6 +18,17 @@ function M.setup()
     group = group,
   })
 
+  api.nvim_create_autocmd("SessionLoadPost", {
+    callback = function()
+      for _, buf in ipairs(api.nvim_list_bufs()) do
+        local buf_name = api.nvim_buf_get_name(buf)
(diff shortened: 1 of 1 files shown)


lua/neogit/autocmds.lua
```diff
@@ -18,6 +18,17 @@ function M.setup()
     group = group,
   })
 
+  api.nvim_create_autocmd("SessionLoadPost", {
+    callback = function()
+      for _, buf in ipairs(api.nvim_list_bufs()) do
+        local buf_name = api.nvim_buf_get_name(buf)
+        if buf_name:match("Neogit%w+") then
```
igorlfs, line 25: Note: we can't use **`^Neogit%w+`** (i.e., `^`) because when restoring, the buffer is no longer owned by neogit and neovim ends up prepending the full path to the cwd.

CKolkey:
Thanks :)

PR #1995, opened 2026-07-25
fix(sessions): prevent error on session cleanup

igorlfs:
An oversight. When multiple tabs with neogit buffers are shown, it could raise an error:

```
Error in UIEnter Autocommands for "*":
E5108: Lua: [string "vim/_core/editor"]:490: UIEnter Autocommands for "*"..script nvim_exec2() called at UIEnter Autocommands for "*":0[1]../tmp/nvim.dev/restart_session_cHeRn8[82]..SessionLoadPost Autocommands for "*": Vim(append):E814: Cannot close window, only autocmd window would remain
```

Solution: `vim.schedule` ftw


CKolkey:
All good, thanks for following up:)

PR #1996, opened 2026-07-27
fix(status): avoid loading status mappings at the top-level

jungomi:
The status mappings defined in the `setup()` call have been ignored completely.

Some modules loaded the status mappings at the top-level, which ignored the user mappings. This is due to the fact that mappings are cached in `neogit/config.lua`. When the `get_versed_status_map()` is called at the top-level of the files, it is executed before the setup can be called and therefore does not incorporate the changes made by the user.

All calls to the status mappings have be moved inside the functions that use them.


CKolkey:
Good catch, thanks

PR #2001, opened 2026-08-05
fix(finder): handle abort in fzf-lua integration

aliaksandr-trush:
`Finder:find` documents that `on_select` is invoked with the selected item or `nil` if aborted. Every integration honors this except `fzf_lua`: the `esc`/`ctrl-c`/`ctrl-q` actions only fire when fzf prints one of those keys back, and when the fzf process is instead terminated (window closed, job killed) fzf-lua dispatches no action at all.

Through `find_async = a.wrap(Finder.find, 2)`, callers using `open_async` (e.g. `bb` -> `checkout_branch_revision`) `coroutine.yield` waiting for `on_select` to resume them. When `on_select` never runs, the coroutine is suspended forever holding the popup lock (`permits == 0`) and the picker appears dead until Neovim is restarted.

Wire fzf-lua's `winopts.on_close` to an abort handler so a picker close still completes the finder via `on_select(nil)`. `on_close` runs before the selected action is dispatched, so it is deferred to the next event loop tick; the `completed` guard (mirroring `mini_pick_choose` and `snacks_confirm`) ensures `on_select` runs exactly once.


ifakhrutdinov:
I've tried the patch and I think this also fixes my issue, which is:
* Try to push changes with neogit
* Get a pop-up to choose a remote (I have a few)
* Hit ESC to close the pop-up
* Try pushing again, but no pop-up appears and nothing happens

ifakhrutdinov:
@aliaksandr-trush , this patch seems to have broken the range diff functionality. If you use `Neogit diff` with a range, after picking the second branch and the range type (e.g., `(1) Range (a..b)`), nothing happens.

aliaksandr-trush:
@ifakhrutdinov thank you for the finding. Ive updated the PR and it works for me now.

ifakhrutdinov:
Thanks, that has fixed the diff issue.

PR #2004, opened 2026-08-20
fix: async input in rebase editor

EvgeniyPeshkov:
Input of a command to exec is broken in rebase editor after switching to vim.ui.input.
Fixed by wrapping `get_user_input` into coroutine.


EvgeniyPeshkov:
Init repo command had the same bug, added a fix for it as well

CKolkey:
Thanks 👌

